PR #11, opened 2020-03-07
Updated naming to show SafePaths on home screen instead of tripleblind

nic-rodriguez:
done on iOS, app still ran after fix


penrods:
The changes look good to me, were you able to run this on a iPhone?  If so, let's merge it

PR #13, opened 2020-03-07
now store locations in an array

nic-rodriguez:


penrods:
@vitorpamplona Are your changes to use SQL ready, or shall I just take this one?

PR #31, opened 2020-03-13
Refactoring buttons

vitorpamplona:
Removed the extra mentions to NegButton as per PR review.


penrods:
Looks like there really isn't anything here, I'm gonna just close it unmerged.

PR #34, opened 2020-03-13
Google Auto Import (Prototype)

vitorpamplona:
Screen requests users to download the Google Takeout file for the Location History. 
After the download, users have to go back and come back to import. 

Watch the logs to see what the app is doing. We don't have the UI to let the user know the current progress yet.


penrods:
I had a little issue with getting the import to work, but it is a good start and didn't crash.  I'm gonna merge it and then we can refine further.

PR #35, opened 2020-03-14
Curate and collapse GPS location data

### LINKED ISSUE #2: Combine and curate location data
Have the location data store under single object, not as independent pieces.

Basic operation logic:
```
locs = []
if ("loc_data" exists)
   temp_locs = load("loc_data")

   # Curate
   for (i=0; i < len(temp_locs); i++)
      if (temp_locs[i] < 28 days old)
         locs += temp_locs[i]

...

# After every snapshot
save(locs)
```

penrods:
This changes several things about the data being stored:
* The time being stored in the structure is now UTC.  It was previous local
  time
* The data being stored is now simplified to:
   [ (utc_unix_time, GPS lat, GPS lon) ... ]
* The stored data is now limited to 28 days

 ### Fixed Issues ###
Curation resolves #2

 ### Tech Notes ###
All of the previous trails are basically trash.  This shouldn't crash, but
the time is not in the correct timezone.  Potentially we could attempt to
convert them, but there is a chance that the data is in the wrong timezone.
I guess we could convert based on the GPS coords, but not worth it now.

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -17,18 +17,14 @@ import {
 } from 'react-native-webview';
 import Button from "../components/Button";
 import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
-
-
-import {
-    GetStoreData,
-    SetStoreData
-} from '../helpers/General';
+import { GetStoreData, SetStoreData } from '../helpers/General';
 
 class LocationTracking extends Component {
     constructor(props) {
         super(props);
     }
     componentDidMount() {
+
         BackgroundGeolocation.configure({
             desiredAccuracy: BackgroundGeolocation.HIGH_ACCURACY,
             stationaryRadius: 50,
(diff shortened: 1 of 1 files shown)


app/views/LocationTracking.js
```diff
@@ -71,8 +67,22 @@ class LocationTracking extends Component {
...
-                    locationData.push(location);
-                    SetStoreData('LOCATION_DATA', locationData);
+                    // Curate the list of points
+                    var nowUTC = new Date().toISOString();
+                    var unixtimeUTC = Date.parse(nowUTC);
+                    var unixtimeUTC_28daysAgo = unixtimeUTC - (60 * 60 * 24 * 1000 * 28);
+
+                    var curated = [];
+                    for (var i = 0; i < locationData.length; i++) {
+                        if (locationData[i]["time"] > unixtimeUTC_28daysAgo) {
```
tremblerz: Shouldn't we store the timestamp as well during the location logging

PR #37, opened 2020-03-14
LocationServices Refactor

davidbiga:
Creates Statically Typed LocationServices class to handle all background instantiation once.

BackgroundGeolocation is smart in itself and will handle any form of termination events.  In addition, the onListen event `stop` was added where future release can include a local push notification to tell the user that location services were terminated.


penrods:
Woot! Excellent start.

I pushed a change to add a log message when saving a point.  It revealed a couple things:

* In the emulator, I shutdown and launched from the icon.  It looks like it was running two copies -- double logs.  Possibly there are two apps or maybe two services within a single app?

davidbiga:
Definitely should not be.

PR #38, opened 2020-03-14
Add file based sharing

tremblerz:
- Allows the file to be shared instead of raw string


tremblerz:
Will remove the dead code

penrods:
NOTES:  This requires accessing the filesystem when you do an Export.  Potentially we'll need to warn people this is going to cause a popup.

Issue #50, opened 2020-03-15
Put notification if location is not getting logged


tremblerz:
The location might not be getting logged due to the location turned off through settings or some other reason. We need a rigorous way of testing it.

tushar-dadlani:
How aggressively do you want to notify the user?

tremblerz:
We have push notifications now for iOS. On Android they still don't appear

penrods:
Still an issue?

summetj:
Yes lack of notification is still an issue on android. When running and logging in background, if locations permission is removed, there is no notification. (However, if the app is manually re-started, it detects that it needs to ask for location access again and goes through the "getting started" flow.)

maher1719:
Maybe we can find the location with the network tower id who provide mobile signals and check the id with remote tower database

Shabanisimo:
The bug is fixed.

Issue #17, opened 2020-03-11
No indicator during Location service failures


penrods:
There are several reasons that location information might not be working properly:
  * GPS has been turned off
  * User disabled the permission under settings
  * Under iOS, app was shutdown in the background
  * Probably others

We need to validate access to GPS and notify/prompt the user when this happens.

StartupSam:
Was this issue completed with #64? If so, should close for housekeeping purposes

harshvitra:
@tremblerz @penrods 

Please check if we can close this 

If not what is the todo for this issue ?

PR #64, opened 2020-03-16
Add missing notifications for Location Services

### LINKED ISSUE #23: Check and prompt if the location is disabled
This should be helpful -
[https://stackoverflow.com/questions/41795968/react-native-check-and-prompt-user-to-enable-network-gps](https://stackoverflow.com/questions/41795968/react-native-check-and-prompt-user-to-enable-network-gps)

penrods:
Several conditions cause Private Kit to stop failing when it cannot access
GPS info.  We should notify the user when this has happened.  That includes:
* The app doesn't have proper permissions
* The permissions have been revoked
* The location service has been shut down

Still to do is a notification to show the user that the background service
is not doing anything when the user turns off location services.

 ### Testing Notes ###
Toggle permissions and location services, go in and out of the app.  There
should be notification within the app whenever it cannot access things.

 ### Fixed Issues ###
Resolves #23, #17

Leaving #50 open until we add a Notification that pops up

### DIFF
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -73,13 +73,13 @@ export default class LocationServices {
             locationProvider: BackgroundGeolocation.ACTIVITY_PROVIDER,
 
             // DEBUG: Use these to get a faster output
-            interval: 2000,
-            fastestInterval: 2000, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
-            activitiesInterval: 2000,
+                /*interval: 2000,
+                fastestInterval: 2000, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
+                activitiesInterval: 2000,*/
 
-            //interval: 20000,
-            //fastestInterval: 60000 * 5, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
-            //activitiesInterval: 20000,
+            interval: 20000,
+            fastestInterval: 60000 * 5, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
+            activitiesInterval: 20000,
 
             stopOnStillActivity: false,
             postTemplate: {
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -3,6 +3,7 @@ import {
     SetStoreData
 } from '../helpers/General';
 import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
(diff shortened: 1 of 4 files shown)


tremblerz:
Testing it now. If I press No(on the application), leave the App and come back to the app, it doesn't ask to turn on locations now. Would say it's more of a feature creep for this one

app/services/LocationService.js
```diff
@@ -64,21 +65,21 @@ export default class LocationServices {
             desiredAccuracy: BackgroundGeolocation.HIGH_ACCURACY,
             stationaryRadius: 50,
             distanceFilter: 50,
-            notificationTitle: 'PrivateKit Enabled',
-            notificationText: 'PrivateKit is recording path information on this device.',
+            notificationTitle: 'Private Kit Enabled',
+            notificationText: 'Private Kit is securely storing your GPS coordinates once every five minutes on this device.',
```
tremblerz: Might not say securely storing as we are not encrypting the coordinates, they are stored in raw format. However, no other app can access it so it is good as well.

penrods:
Yeah, I debated a little.  If you intentionally say "No", should we keep bugging you?  I think the Notifier 
will take care of this for us, too.

app/services/LocationService.js
```diff
@@ -64,21 +65,21 @@ export default class LocationServices {
...
             debug: false,    // when true, it beeps every time a loc is read
             startOnBoot: false,
             stopOnTerminate: false,
             locationProvider: BackgroundGeolocation.ACTIVITY_PROVIDER,
 
             // DEBUG: Use these to get a faster output
-            //            interval: 2000,
-            //            fastestInterval: 2000, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
-            //            activitiesInterval: 2000,
+            interval: 2000,
```
tremblerz: will comment it

penrods:
Also, I got lazy around 5am.  ;-)

app/services/LocationService.js
```diff
@@ -195,10 +203,39 @@ export default class LocationServices {
...
+                        onPress: () => BackgroundGeolocation.showLocationSettings()
+                    },
+                    {
+                        text: 'No',
+                        onPress: () => console.log('No Pressed'),
+                        style: 'cancel'
+                    }
+                    ]), 1000);
+            }
+            else if (!status.authorization) {
```
tremblerz: If I don't authorize and press Yes for settings, then I end up at App settings which doesn't have location information anywhere.

Issue #65, opened 2020-03-16
iOS not building on develop (45fc937419069a94dab3d4dd11deab4f33b080a0)


tushar-dadlani:
I am getting this error on the iOS build for develop https://github.com/tripleblindmarket/private-kit/commit/45fc937419069a94dab3d4dd11deab4f33b080a0

It gets stuck on a white screen and the app fails to load. I tested release v0.5 and it worked.

**Device iPhone 8 - iOS 13.3.1**
```
2020-03-16 01:01:35.631 [info][tid:main][RNGestureHandlerManager.m:136] [GESTURE HANDLER] Initialize gesture handler for root view <RCTRootContentView: 0x1013049b0; reactTag: 1; frame = (0 0; 375 667); gestureRecognizers = <NSArray: 0x2800f1170>; layer = <CALayer: 0x280eb8ec0>>
2020-03-16 01:01:35.660749-0700 PrivateKit[438:28526] +backgroundSessionConfiguration: is deprecated. Please use +backgroundSessionConfigurationWithIdentifier: instead
2020-03-16 01:01:35:660 PrivateKit[438:28526] Backing up original handler.
2020-03-16 01:01:35:660 PrivateKit[438:28526] Setting new handler.
2020-03-16 01:01:35.693860-0700 PrivateKit[438:28526] Current db version: 3
2020-03-16 01:01:35:694 PrivateKit[438:28519] BgGeo #configure: Config: distanceFilter=50 stationaryRadius=50 desiredAccuracy=0 activityType=OtherNavigation activitiesInterval=10000 isDebugging=0 stopOnTerminate=0 url= syncThreshold=100 maxLocations=10000 httpHeaders={
} pauseLocationUpdates=0 saveBatteryOnBackground=0 locationProvider=1 postTemplate={
    foo = bar;
    lat = "@latitude";
    lon = "@longitude";
}
2020-03-16 01:01:35.695266-0700 PrivateKit[438:28523] BgGeo #configure: Config: distanceFilter=50 stationaryRadius=50 desiredAccuracy=0 activityType=OtherNavigation activitiesInterval=10000 isDebugging=0 stopOnTerminate=0 url= syncThreshold=100 maxLocations=10000 httpHeaders={
} pauseLocationUpdates=0 saveBatteryOnBackground=0 locationProvider=1 postTemplate={
    foo = bar;
    lat = "@latitude";
    lon = "@longitude";
}
[...]

PR #57, opened 2020-03-16
Andynuzzo/fix require cycles

andreanuzzo:
Changes to fix warnings in https://github.com/tripleblindmarket/private-kit/issues/54 

It uses a small post install patch, because the issue is from the native rn-fetch-blob package.

Issue #54, opened 2020-03-15
Debugger warnings when loading app (require cycles)


andreanuzzo:
I am getting a lot of these warnings when opening the app. 
Xcode 10.2 and iOS 13.3.2 simulator (iPhone 8).

![image](https://user-images.githubusercontent.com/36167533/76712058-99172500-66eb-11ea-9468-452848d20560.png)

I'm not an iOS developer, but I can try to fix it if needed.

penrods:
I see this under Android too.  If you can investigate, please do.  I don't think this is dangerous, so I've been ignoring it.  But it would be nice to eliminate if possible.

andreanuzzo:
It is related to a known bug of rn-fetch-blob, left unfixed for a couple of years. 

https://github.com/joltup/rn-fetch-blob/issues/183#issuecomment-450826541

I have fixed it manually with MIT's `patch-package`.

PR #67, opened 2020-03-16
Andynuzzo/fix require cycles + Clearing all Warnings and Bug Fixes for iOS

vitorpamplona:
This PR includes Andy's original PR #57 + a few bug fixes for iOS related to the latest use of BackgroundLocation Android Headless Task (that only exists in Android) on Issue #65 and #54


vitorpamplona:
It works for me on iOS and Android

PR #86, opened 2020-03-17
Fixed stucked notification

### LINKED ISSUE #66: Stop recording location does not remove the notification bar
When the location logging is not on, even then the notification bar keeps showing the location is getting recorded.

HrushikeshAgrawal:
Fixes #66

### DIFF
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -1,295 +1,271 @@
-import {GetStoreData, SetStoreData} from '../helpers/General';
+import {
+  GetStoreData,
+  SetStoreData
+} from '../helpers/General';
 import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
-import {Alert} from 'react-native';
+import { Alert } from 'react-native';
 
 var instanceCount = 0;
 var lastPointCount = 0;
 
+
 function saveLocation(location) {
   // Persist this location data in our local storage of time/lat/lon values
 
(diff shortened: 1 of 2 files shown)


app/services/LocationService.js
```diff
@@ -1,270 +1,295 @@
...
+  static stop() {
+    // unregister all event listeners
+    BackgroundGeolocation.removeAllListeners();
+    BackgroundGeolocation.stop();
+    instanceCount -= 1;
+  }
+
+  static optOut(nav) {
+    BackgroundGeolocation.removeAllListeners();
+    BackgroundGeolocation.stop();
```
tremblerz: This is the only code change right? just want to make sure that I am not missing any extra piece of code which has been added/removed.
Also, if possible please apply changes in such a way that code difference is minimized, right now it is hard to navigate across.

HrushikeshAgrawal: Yes, that (BackgroundGeolocation.stop();) is the only code change I had made. The rest of the changes were due to the code formatter I used while making the change. Will be careful from next time and not use the code formatter. So, would you like me to close this pull request and create a new one or should we continue with this one?

rparet: Hi there. Let's continue with this one. Please keep this PR open but push up a new commit that is only your change, and no code formatter changes. Thanks!

rparet:
Fixed in #115 - thank you very much!

PR #106, opened 2020-03-19
Update README.md

amclough:


rparet:
Per @amclough we're waiting for some updated language here. Also, this change is now in conflict with the current README.md. @amclough can you rebase from master and send in a new language draft?

penrods:
See #126.   I think this on can be closed.

PR #109, opened 2020-03-19
Location Issues fix

### LINKED ISSUE #74: VERIFY:  Backfill Stationary point captures if needed
The BackgroundGeolocation library has the ability to only report changes in location when the user moves.  This saves power in some cases, e.g. if the phone is sitting on a charger and not moving, the accelerometer shows it is stationary.  The library saves power by returning a Stationary point.

Verify the behavior around this.  I suspect it only generates a callback once, then stops calling until the phone moves again.  If this happens, upon the move we should backfill so that we have the same point reported at 5 minute intervals until that next point report.

Example:

@3:01 - I set my phone down on the desk
@3:02 - The BackgroundGeolocation callback triggers and we save a coordinate
@3:40 - I move my phone
@3:40 - The BackgroundGeolocation callback triggers and we save a coordinate

If the above _does_ occur (which we need to verify), we should save the same reported coordinate from 3:02 at 3:07, 3:12, 3:17, 3:22, 3:27, 3:32, 3:37 and then save the new 3:40 coordinate.

### LINKED ISSUE #89: Apple Rejection: Do we need Motion Tracking?
We noticed that your app requests the user’s consent to access their motion but does not clarify the use of the motion in the applicable purpose string.

Next Steps

Please revise the relevant purpose string in your app’s Info.plist file to specify why the app is requesting access to the user's motion and location. You can modify your app's Info.plist file using the property list editor in Xcode.

To help users understand why your app is requesting access to their personal data, all permission request alerts in your app should specify how your app will use the requested feature.

### LINKED ISSUE #94: iOS doesn't record location even if the application is running in foreground.
The first two points were logged, nothing afterwards.
Devices 
- iPhone11 - 13.3

EDIT - 
Seems like it is only happening when it is stationary. If I use bike riding simulator, it stops.

tremblerz:
remove motiontracker requirement; Fixes #89 
adjust location params;
backfill stationary points; Fixes #94, Fixes #74 

NOTE - 
- Speed at which location callback is happening is slightly inconsistent.

### DIFF
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -7,7 +7,9 @@ import { Alert } from 'react-native';
 
 var instanceCount = 0;
 var lastPointCount = 0;
-
+var locationInterval = 60000 * 5;  // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
+// DEBUG: Reduce Time intervall for faster debugging
+// var locationInterval = 5000;
 
 function saveLocation(location) {
     // Persist this location data in our local storage of time/lat/lon values
@@ -32,6 +34,15 @@ function saveLocation(location) {
             for (var i = 0; i < locationArray.length; i++) {
                 if (locationArray[i]["time"] > unixtimeUTC_28daysAgo) {
                     curated.push(locationArray[i]);
+                }
+            }
+
+            // Backfill the stationary points, if available
+            if (curated.length >= 1) {
+                var lastLocationArray = curated[curated.length - 1];
+                var lastTS = lastLocationArray["time"];
+                for (; lastTS < unixtimeUTC - locationInterval; lastTS += locationInterval) {
+                    curated.push(JSON.parse(JSON.stringify(lastLocationArray)));
                 }
             }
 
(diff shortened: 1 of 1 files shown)


tremblerz:
tested on iOS 13.3 and Pixel 29

app/services/LocationService.js
```diff
@@ -63,30 +74,23 @@ export default class LocationServices {
...
+            locationProvider: BackgroundGeolocation.DISTANCE_FILTER_PROVIDER,
 
-            interval: 20000,
-            fastestInterval: 60000 * 5, // Time (in milliseconds) between location information polls.  E.g. 60000*5 = 5 minutes
-            activitiesInterval: 20000,
+            interval: locationInterval,
+            fastestInterval: locationInterval,
+            activitiesInterval: locationInterval,
 
+            activityType: "AutomotiveNavigation",
```
davidbiga: We might be knocked on this from for app audit because we are not a navigation app.

rparet:
@davidbiga @tremblerz I think this can be closed now that we've merged the change in a subsequent PR. Yes?

tremblerz:
closing

Issue #129, opened 2020-03-20
Even after denying location permission, the button changes to Stop Logging


nehalsep7:
When permission is denied, it should stay at Start Logging

PR #141, opened 2020-03-21
Check for background geo authorization before updating state

rparet:
Per #129 if you denied permission in the system dialog, we'd still update the button as if you had started logging. This fix creates the desired behavior, setting the state back to the way it was before you requested permission.

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -16,6 +16,7 @@ import {
 } from 'react-native';
 import colors from "../constants/colors";
 import LocationServices from '../services/LocationService';
+import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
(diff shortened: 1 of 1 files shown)


app/views/LocationTracking.js
```diff
@@ -77,9 +78,22 @@ class LocationTracking extends Component {
         SetStoreData('PARTICIPATE', 'true').then(() =>
             LocationServices.start()
         );
-        this.setState({
-            isLogging:true
-        })
+
+        // Check and see if they actually authorized in the system dialog.
+        // If not, stop services and set the state to !isLogging
+        // Fixes tripleblindmarket/private-kit#129
```
penrods: I personally don't include references to issues, I just explain the "why" like you did in the comment.  But it doesn't bother me too much.

Issue #136, opened 2020-03-21
Non-breaking error in dev_setup.sh script - unexpected token in package.json?


StartupSam:
When running dev_setup.sh, got an error about an unexpected token that seems to originate in package.json. The rest of the script completed without errors, and after a yarn install the app ran correctly in emulator, so this isn't breaking anything.

This should be a very low priority issue - just creating here for the sake of reference.

System: Ubuntu 19.10
Node: 13.8.0
NPM: 6.13.6

Verbose log:
```
0 info it worked if it ends with ok
1 verbose cli [
1 verbose cli   '/home/samstowers/.nvm/versions/node/v13.8.0/bin/node',
1 verbose cli   '/home/samstowers/.nvm/versions/node/v13.8.0/bin/npm',
1 verbose cli   'install'
1 verbose cli ]
2 info using npm@6.13.6
3 info using node@v13.8.0
4 verbose npm-session 94eb47014aa4fd5b
5 silly install runPreinstallTopLevelLifecycles
6 silly preinstall privatekit@0.5.3
7 info lifecycle privatekit@0.5.3~preinstall: privatekit@0.5.3
8 silly install loadCurrentTree
9 silly install readLocalPackageData
10 timing stage:loadCurrentTree Completed in 27ms
11 silly install loadIdealTree
12 silly install cloneCurrentTreeToIdealTree
13 timing stage:loadIdealTree:cloneCurrentTree Completed in 0ms
14 silly install loadShrinkwrap
15 timing stage:rollbackFailedOptional Completed in 1ms
16 timing stage:runTopLevelLifecycles Completed in 75ms
17 silly saveTree privatekit@0.5.3
18 verbose stack SyntaxError: Unexpected token
18 verbose stack  in JSON at position 207831 while parsing '{
18 verbose stack   "name": "privatekit",
18 verbose stack   "version": "'
18 verbose stack     at JSON.parse (<anonymous>)
18 verbose stack     at parseJson (/home/samstowers/.nvm/versions/node/v13.8.0/lib/node_modules/npm/node_modules/json-parse-better-errors/index.js:7:17)
18 verbose stack     at module.exports (/home/samstowers/.nvm/versions/node/v13.8.0/lib/node_modules/npm/lib/utils/parse-json.js:4:10)
18 verbose stack     at parsePkgLock (/home/samstowers/.nvm/versions/node/v13.8.0/lib/node_modules/npm/lib/install/read-shrinkwrap.js:67:12)
[...]

PR #148, opened 2020-03-22
Fix outdated package.lock

rparet:


rparet:
Fixes #148 #136

rparet:
No longer needed.

Issue #150, opened 2020-03-22
Feature: Distributed health authorities


rparet:
In the app there will be a config dialog where you can enter your "Trusted Health Authority" (can be multiple of them)

Each authority is basically just a URL (later we will pull a global registry -- it can just be a file within github that is served by rawgit).  But there must always be the ability for a user to type in a unique authority to ensure global usability.

The app GETs that URL and downloads a JSON file:

```json
{ "authority_name":  "Platte County Health Departartment",
  "publish_date": "sometime",
  "info_website": "http://local.authority/website/with/news.html",
  "concern_points": [ { "time": 123, "latitude": 12.34", "longitude": 12.34}, ...]
}
```

With the "info_website" can add a localized tab on the News page.  This allows local authorities to message users.  This is the local news feed, we don't control it.

We use the Authority Name to populate the list box of selected authorities.

We pull the points once a day an run comparisons.

==

Currently News tab opens https://privatekit.mit.edu/views

* Redsigning the news tab to allow customers to choose from different news sources
* Using the above tab to give customers a good overview of what features exist / what is coming.

StartupSam:
Would be great to get information in here from mainstay sources, e.g. CDC, WHO. Adding some summarized symptom and testing info somewhere would also be excellent - offline availability for that info could be quite useful

DenisPitcher:
Has a git file been identified that contains the health authority list?

PR #160, opened 2020-03-22
Add Settings Screen - #150

davidbiga:
This is the basic setup for view implementation of `Authorities` under `Settings`.  The idea is that users will be able to submit authority api urls that will be used to download a `json` file that contain important information that is relevant to that user and their location/country. 

TODO: add list with ability to add - delete - edit urls. 
TODO: configure settings on authorities to be stored inside local storage `AUTHORITY_URLS`


penrods:
See #160

PR #173, opened 2020-03-23
moving pod install command to npm run install:pod

Utzel-Butzel:
Simplifying the usage by moving all the commands into npm/npx


tremblerz:
Please fix the json conflict

PR #174, opened 2020-03-23
Improve data exporter to share data with better names

### LINKED ISSUE #68: Refine Share
The basic Share functionality works, but it could use some refinement.  Specfically two things:

* Improve the filename.  Currently it is "<number>.null".  Change it to be "PrivateKit_<number>.json"
* Add a message body.  Try:
  `Here is my location log from Private Kit.`
* If titles are supported, make it: "Shared PrivateKit_<number>.json"

tushar-dadlani:
- Updated react-native-share to 3.1.0 because of [issue](react-native-community/react-native-share#702)
- Separated out iOS specific and android share logic as the behaviors for sharing are significanlty different even for a simple case like filename

Fixes #68

### DIFF
--- a/app/views/Export.js
+++ b/app/views/Export.js
@@ -1,15 +1,14 @@
 import React, { Component } from 'react';
 import {
-  SafeAreaView,
-  StyleSheet,
-  ScrollView,
-  Linking,
-  View,
-  Text,
-  Image,
-  Dimensions,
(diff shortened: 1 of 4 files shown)


tushar-dadlani:
rebased and fixed merge conflict in ios/Podfile.lock

If we remove the 'message' property on iOS it should only create a single

Issue #188, opened 2020-03-24
Add feedback on Check Overlap function if there is no exposure


StartupSam:
At the moment, pressing the "check overlap" button loads, but provides no visual feedback in case of no exposure. To a regular user, this feels broken - will probably result in bad reviews in respective app stores. Should add some kind of visual feedback here ASAP.

StartupSam:
I'll work on this after I get my fork's commit history fixed

PR #190, opened 2020-03-24
Added feedback for trace overlap if it comes back with no results

StartupSam:
See issue #188 - I'd advise this is merged ASAP since the current build feels broken from a user perspective.

### DIFF
--- a/app/views/Overlap.js
+++ b/app/views/Overlap.js
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
 import React, {
     Component
 } from 'react';
@@ -38,6 +39,8 @@ const base64 = RNFetchBlob.base64
 
 const public_data = "https://docs.google.com/spreadsheets/d/1itaohdPiAeniCXNlntNztZ_oRvjh0HsGuJXUJWET008/export?format=csv"
 const show_button_text = "Show Me Trace Overlap";
+const overlap_true_button_text = "Trace Overlap Found";
+const no_overlap_button_text = "No Trace Overlap Found";
 const INITIAL_REGION = {
     latitude: 36.56,
     longitude: 20.39,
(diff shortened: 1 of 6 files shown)


app/views/Overlap.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
penrods: What's this for?

package.json
```diff
@@ -52,6 +52,7 @@
     "eslint": "^6.8.0",
     "jest": "^25.1.0",
     "metro-react-native-babel-preset": "^0.58.0",
+    "prettier": "1.17.0",
```
penrods: Is this and the package-json intentional in this PR?  If not, let's remove these files.

app/views/Overlap.js
```diff
@@ -38,6 +39,8 @@ const base64 = RNFetchBlob.base64
 
 const public_data = "https://docs.google.com/spreadsheets/d/1itaohdPiAeniCXNlntNztZ_oRvjh0HsGuJXUJWET008/export?format=csv"
 const show_button_text = "Show Me Trace Overlap";
+const overlap_true_button_text = "Trace Overlap Found";
```
penrods: We need to make these translatable strings.

penrods: See the ```languages.t('label.overlap')``` below for an example.  The strings come from ../lcoales/us/overlap.json

app/views/Overlap.js
StartupSam: Ah, sorry, this was for personal sanity editing the file - with Prettier enabled, it's just a mess of red lines. Didn't want to do a Prettier fix in the same PR so temporarily disabled it so I could edit. Will remove the line

package.json
StartupSam: This was intentional - realized I should've included it in the primary Prettier PR, since different versions of Prettier have slightly different rules. I can make it a separate PR if you need

app/views/Overlap.js
StartupSam: Translation wasn't set up for the button text at all, but I can go ahead and add it as part of this PR

StartupSam:
Removed the Prettier disabling comment and added translation for the button texts and title.

Issue #185, opened 2020-03-24
Layout broken


vitorpamplona:
We need to be more conscious of the different screen sizes. Should we move Import/Export buttons to Menu/Settings? 

![image (1)](https://user-images.githubusercontent.com/532031/77455363-21728580-6dd0-11ea-9c38-74388b181115.png)

StartupSam:
Getting an even worse break on my older Android phone (OnePlus 3):

![screenshotAndroid](https://user-images.githubusercontent.com/6233577/77473521-9bf8d080-6de3-11ea-8b0e-2bfd54798e2e.jpg)

StartupSam:
Can someone mark this as urgent? A number of core functions are unreadable as it is

E3V3A:
@rparet  Please add to the back log.

penrods:
The immediate need has been addressed with the nightly buiild.  A larger UI overhaul will be happening in the near future as more things get crammed on the main screen.

PR #196, opened 2020-03-25
Fix broken layout on low-res devices using flex layout on LocationTracking.js and Overlap.js

StartupSam:
Per issue #185 - key UI elements were getting blocked out due to hard-set values. Reorganized and moved elements in the main two affected screens (Location Tracking and Overlap Checking) over to flex styling - it looks good on low-res devices now, but I need someone to check high-res for me.

Also fixed a typo (lowercase) on a button label.

Many lines were changed in the file since I also re-formatted them with Prettier. (I'll probably set up some kind of hook to do this for everyone automatically, so this isn't necessary)


StartupSam:
Also need to test on an iPhone notch

vitorpamplona:
Looks good in big and small screens, including the iPhone with the notch. 
Screen Shot 2020-03-24 at 11 00 20 PM

PR #202, opened 2020-03-25
Enhancing the License screen

penrods:
The kebab > License screen now is a combination of an About screen
and a license.  The top of the screen shows information about the current
app version, OS version on which it is running and screen resolution info.

Below that is license test.  It pulls from the new LICENSE.json in the
assets directory.  It will render an array of licenses, so later we can
add the licenses of components we use in addition to the main Private
Kit MIT license.

 ### Testing Notes ###
Go to kebab > Licenses.  You should see live info about your system.

 ### Environment Notes ###
This adds a asset/LICENSE.md, and it also imports the package.json.
I think this should build naturally, but need to test.

### DIFF
--- a/app/views/Licenses.js
+++ b/app/views/Licenses.js
@@ -4,15 +4,19 @@ import {
   StyleSheet,
   View,
   Text,
+  Platform,
   Image,
   Dimensions,
   TouchableOpacity,
   BackHandler,
 } from 'react-native';
+import { WebView } from 'react-native-webview';
+import packageJson from '../../package.json';
 
 import colors from '../constants/colors';
 import backArrow from './../assets/images/backArrow.png';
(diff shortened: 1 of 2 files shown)


app/views/Licenses.js
```diff
@@ -38,6 +42,25 @@ class LicensesScreen extends Component {
...
+
+    for (var i = 0; i < licenses.licenses.length; i++) {
+      var element = licenses.licenses[i];
+
+      result += '<B>' + element.name + '</B><P>';
+      result += element.text.replace(/\n/g, '<br/>');
+      result += '<hr/>';
+    }
+    result += '</body></html>';
+    console.log('Here 2!:', result);
```
rparet: redact?

PR #216, opened 2020-03-26
Fixup PR#97

rparet:
PR #97's commit history has been hacked up, so extracted the original fix, put it on develop~HEAD and created this PR.


tremblerz:
let's revisit this one later if needed.

PR #231, opened 2020-03-27
Improve export point string (#71)

### LINKED ISSUE #71: Improve the saved point description
Currently the Export page shows some basic info about the saved data.  It displays:
   ```Points to be shared:  6134```

We should expand this into something more friendly.  Describe number of points, oldest date, and how recently it was updated.

Example:  
```
Log has 22 days, 345 points, last updated 4 minutes ago.
```

Once that is possible, we can see it if will fit on the main screen, updated each time a new point is saved.

NOTE: Be aware that the stored datetime info is in UTC.  Should be able to just subtract the first point's time from the current unixtimeUTC, then do division to figure out the number of days saved.

Rikku-1:
I hope this works.  This is my first PR on this project, so please let me know if there's anything I'm missing!

Closes #71


tremblerz:
Great contribution @Rikku-1 , thanks for submitting this. While this PR is complete in itself, right now we have another standing PR #244 which solves the same issue but also re-formats the structure and also adds the translation files, so to avoid merge complications we will go ahead with that one.
Thanks

Issue #114, opened 2020-03-19
Bug on Import (Android)


vitorpamplona:
"Great idea. Unfortunately, the import from Google Location isn't working. I can get everything to download but then I get Error: -1"

This error was reported by Matthew Heflin on the PlayStore.

manalive:
It's me, the Matthew Heflin. 

I'm on a Pixel 3XL running Android 9.

conradm:
Same error here

Pixel 3xl
Android 10
Build: QQ2A.200305.002

![Screenshot_20200321-081146](https://user-images.githubusercontent.com/3872958/77221646-b9137200-6b4b-11ea-9fc6-744f95b19eec.png)

Creating a download works fine
Downloading a file works fine. 
After downloading a file I get the err - 1

greggles:
I have this issue as well on a pixel3 google fi connected on wifi and Android 10. My location history goes back a few years and I selected the option to split by month and store the data in zip files in Google Drive.

I tried downloading the most recent month of data inside the iframe and I think that might have actually worked as I see a Download complete for a `.bin` file. However inside of PrivateKit i see the message in the screenshot from @conradm above: 

```
Error loading page 
Domain: undefined 
Error Code: -1
Description: net::ERR_FAILED
```

Looking through reviews in the google play store this is a pretty common error.

ssirag:
I got the same error (Android 9, LG V30) after trying "add to drive" import.  I wasn't able to retry import until I rebooted the phone.

PR #234, opened 2020-03-27
Android device fix for Google Takeout errors #114

popofibo:
Google takeout download for the 1st time fails for fairly large sized files though it downloads the file in the background. This seems to be a chromium regression impacting RN WebView details [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1023678). 

Fix is to reload once and override renderError w/o an error message if reloading. Treat subsequent renderErrors as normal.

### DIFF
--- a/app/views/Import.js
+++ b/app/views/Import.js
@@ -44,6 +44,7 @@ class ImportScreen extends Component {
   }
 
   render() {
+    var counter = 0;
     return (
       <SafeAreaView style={styles.container}>
         <View style={styles.headerContainer}>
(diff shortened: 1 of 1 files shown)


app/views/Import.js
```diff
@@ -44,6 +44,7 @@ class ImportScreen extends Component {
   }
 
   render() {
+    var counter = 0;
```
rparet: I believe if you're running the pre-commit hook eslint is going to want you to change this to 'let'.

greggles:
This indeed seems like it would fix #114.

I notice there's a conflict in Import.js, though.

popofibo:
Opened another pull request with clean merges considering comments above, closing this

PR #240, opened 2020-03-27
Enforce Yarn, remove package-lock.json and add Jest testing to pre-commit hooks

### LINKED ISSUE #227: Have a package-lock.json *and* a yarn.lock - shouldn't we just have one?
Getting a warning in console about different lock versions. Is there any way to enforce using just one (preferably Yarn) for this project?

pathakparth:
Closes #227 

**Remove -** 
package-lock.json

**Add -**
- .npmrc file
- .yarnrc file
- preinstall script for Yarn enforcement
- Jest testing on pre-commit hooks


thetimbanks:
Just out of curiosity, why are the lock files being removed? I understand the point of not having both, but we should at least have one of them, right?

pathakparth:
@thetimbanks 
Agreed !! we should have at least 1 (preferably yarn.lock) but not mandatory to have it on projects, we can prevent using NPM on adding/updating packages. It will also add a prerequisite of everyone knowing/using yarn towards contributing to the project.

**It can be achieved by preinstall script-** 
```
"scripts": {
    "preinstall": "node -e \"if(process.env.npm_execpath.indexOf('yarn') === -1) throw new Error('You must use Yarn to install, not NPM')\""
} 
```
Screen Shot 2020-03-27 at 1 19 00 PM (2)

@rparet @vitorpamplona  - Let me know your thoughts after discussing it with the core dev team and I can update PR accordingly.

thetimbanks:
I think we should continue to use just `yarn.lock` so we can keep everyone on the same versions of the packages.

pathakparth:
@thetimbanks updated PR

ios/Podfile.lock
```diff
@@ -205,6 +205,8 @@ PODS:
     - React-cxxreact (= 0.61.5)
     - React-jsi (= 0.61.5)
   - React-jsinspector (0.61.5)
+  - react-native-background-timer (2.2.0):
```
thetimbanks: Just to make sure, did you mean to commit the `Podfile.lock` in this PR?

pathakparth: Yes, it was modifying for me since the project setup on step "pod install" under ios folder, but I wanted to confirm before I commit the lock file changes.

harshvitra:
@pathakparth can you resolve the conflicts

pathakparth:
@harshvitra, @tremblerz  it would be great if we can get this merged else it will keep running into conflicts.

.yarnrc
```diff
@@ -0,0 +1 @@
+--install.no-lockfile false
```
ten-sis, line 1: looks like [`no-lockfile`](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-no-lockfile) is disabled by default and having `--install.no-lockfile false` is redundant. What was your intention of having `--install.no-lockfile false`? Can `.yarnrc` be removed if we don't need any options here?

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
ten-sis, line 1: nit-picky: looks like you are already taking care of enforcing `yarn` usage in `preinstall` script of package.json. In such case `.npmrc` is not necessary as nobody will be able to install using `npm`.

package.json
```diff
@@ -17,7 +18,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged && yarn test"
```
ten-sis, line 21: ```yarn test``` :+1:

.yarnrc
pathakparth: good catch - removed file

.npmrc
pathakparth: good catch - removed file

pitops:
Even if you keep `yarn.lock` when someone new goes and does a `yarn install` yarn will automatically update some of the dependencies underneath.

package.json
tstirrat: yarn test took 16.951s on my machine. Having this on a pre-commit hook, and requiring that tests pass is going to be a productivity loss and will not scale well as more tests are added.

Github PR actions provide the same safety check.

sergesemashko:
@pitops, addressing your comment:

> Even if you keep yarn.lock when someone new does a yarn install yarn will automatically update some of the dependencies underneath.

from [yarn install docs](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install):
> If yarn.lock is absent, or is not enough to satisfy all the dependencies listed in package.json (for example, if you manually add a dependency to package.json), Yarn looks for the newest versions available that satisfy the constraints in package.json. The results are written to yarn.lock.

So, updated `yarn.lock` on `yarn install` should not be a concern as long as everybody installs packages using `yarn add`

package.json
tstirrat: Please remove this change. It would need larger discussion. Github actions should cover test green-ness

pathakparth: It just took 3-5 seconds on my machine to complete, however, I agree that it might take longer if the unit test cases increases. I believe it's more productivity loss to wait until Github Actions does the same job. Sometimes Github Actions or any other CICD tool might be down or running slow due to unavailable active agents or surge in processing request, it might delay outputs. Having it on pre-commit hooks on local setup is more productive during such cases. 

It's just an additional check and can be skipped using "--no-verify" if no functional changes made on PRs.

pathakparth: @tstirrat - removed from pre-commit hooks for now.

PR #247, opened 2020-03-27
android import fix #114

### LINKED ISSUE #114: Bug on Import (Android)

popofibo:
Fixes #114 

Google takeout download for the 1st time fails for fairly large sized files though it downloads the file in the background. This seems to be a chromium regression impacting RN WebView details [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1023678).

Fix is to reload once and override renderError w/o an error message if reloading. Treat subsequent renderErrors as normal.

### DIFF
--- a/app/views/Import.js
+++ b/app/views/Import.js
@@ -48,6 +48,7 @@ class ImportScreen extends Component {
   }
 
   render() {
+    let counter = 0;
     return (
       <SafeAreaView style={styles.container}>
         <View style={styles.headerContainer}>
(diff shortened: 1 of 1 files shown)


tremblerz:
Hi @popofibo 
Thanks for the PR, have you tested this one?
Seems to be calling style `errorLabel` which is not defined.

popofibo:
> Hi @popofibo
> Thanks for the PR, have you tested this one?
> Seems to be calling style `errorLabel` which is not defined.

I just checked, it was an erroneous reference that I missed to replace, it is in place now.

tremblerz:
Hi,
I tried playing around by making the module throw error, If I change the url to some invalid address, it still throws standard error instead of going inside the `renderError` function

popofibo:
> Hi,
> I tried playing around by making the module throw error, If I change the url to some invalid address, it still throws standard error instead of going inside the `renderError` function

Is your test on iOS? There is an issue with [react-native-webview](https://github.com/react-native-community/react-native-webview/issues/533) that is unresolved for invalid urls. Works in Android for the original issue.

penrods:
Where does this stand?  Conversation seems to have fallen off without a resolution.

popofibo:
My take, this does fix the original Android issue:

- Takeout file that has a significant size get donwloaded in the background however callback throws an exception
- Navigating back to import screen does consider takeout file contents
- This fix ensures while the file is downloading the screen is reloading the 1st time that does not offer an ugly exception screen to the user
- If a subsequent error then it is handled inside `renderError ` with a simplistic error message

Tested on Android and iOS

Issue #144, opened 2020-03-22
V3 sha256 based set intersection implementation, MVP


tremblerz:
Simple interface (which will be improved by UX folks separately) which allows performing intersection between two sets where one set is Sender (think of carrier) and the other set is Receiver (curious healthy person), Receiver gets SHA256 of the elements of the set and tries to match with his own set using our App's interface.
More details will be put soon

solso:
From what I understand that this would allow access to 2nd degree neighbours (contacts of contacts). With an external whiteboard (server) it would be possible do *n*-th degree contacts (i.e. the whole population) without compromising privacy; i did a small write up with the details here (point 2, point1 of the paper is already addressed from what I see).

https://gist.github.com/solso/d3d298b75eaa917b855773118f0872d6#2-contacts-should-be-transitive-contacts-are-not-lists-but-networks

PR #248, opened 2020-03-27
Hashing based private set intersection

tremblerz:
Performs sha256 -> geohashing
The front end is not implemented yet as we do not have "authorized download" or "self-reporting features"
For efficiency of the scheme, I would lean towards storing the hash at the same time when data is logged as it will save us from batched hashing of a file.


tremblerz:
We gotta update the safe-paths branch with develop so that file changes are small

rparet:
Goes with #144

kenpugsley:
@tremblerz - I think this is likely a good candidate for something that we should first implement on the new private-kit repo.  Do you agree?

tremblerz:
@kenpugsley agree. Anyway it looks like a mess right now with respect to the file changes. Closing

PR #250, opened 2020-03-27
Change location marker and description, get rid of redundant markers as well

tremblerz:


rparet:
@tremblerz conflicts with Overlap.js now. Also you can ditch 'react-native-map-clustering' from package.json.

PR #259, opened 2020-03-28
Update Overlap screen to update when it comes into focus

### LINKED ISSUE #253: Refresh map visualization when user has pressed back and goes to the screen again
Right now if I press back and go to the check overlap screen again, it shows same location as no `state` has changed from React's perspective. However, my location might have changed meanwhile and rendering old data becomes sort of useless.
Solution:
In overlap.js, Check component mount and refresh the state by placing markers everytime the component is mounted.
NOTE: This is just one way of implementing, there could be more elegant ways as well.

thetimbanks:
Converted the Overlap class component to a functional component in
order to take advantage of hooks. The hooks allow us to hook into
the focus event so that we can update the map markers when the view
comes into focus. Previously the map would not update after the initial
mounting of the component.

Closes #253


rparet:
looks good @thetimbanks but @tremblerz landed a large-ish overlap feature last night, so some conflicts here.

thetimbanks:
I figured I would need to update this after he got his PR merged. I’ll get it updated today.

thetimbanks:
@rparet Conflicts have been taken care of and made sure the tests were passing.

rparet:
Thanks! @tremblerz can you look at this one? You've been doing the most work on Overlap, so would appreciate your check here.

tremblerz:
looking at it

PR #271, opened 2020-03-29
Remove duplicate function willParticipate

filkaris:
This function appears twice in the `app/views/LocationTracking.js` file, seems to have happened when merging by mistake

Removed the one without any logic. The function left in the file is identical to this one + has some extra logic

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -113,16 +113,6 @@ class LocationTracking extends Component {
     this.props.navigation.navigate('NotificationScreen', {});
   }
 
-  willParticipate = () => {
-    SetStoreData('PARTICIPATE', 'true').then(() => {
-      LocationServices.start();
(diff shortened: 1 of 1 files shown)


rparet:
Good catch, but I think we want to keep the one with setState, and delete the other one.

filkaris:
Hey Rich, thanks for checking it out.

Since I'm new to the project, could you explain why?
I'd be happy to remove the other one, just looking to understand this more.

From what I can gather, someone has intentionally added logic to 
>Check and see if they actually authorized in the system dialog.

according to #129 

Is there a reason we should remove this piece of logic?

filkaris:
> From what I can gather, **someone** has intentionally added logic to

that someone being you :smile:

rparet:
How embarrassing. :) You're totally right. Thanks for this!

PR #275, opened 2020-03-29
Disable share button on Export screen if no log data

thetimbanks:
The Share button on the Export screen should not be enabled if
there is no data in the log. This will disable the button by
default and then enable it once the data is retrieved and determined
that there is at least 1 data point in the log.

### Screenshots
Android (disabled)

image

iOS (enabled)

image


rparet:
Great enhancement! Thanks!

PR #280, opened 2020-03-29
Fix disable location sharing ble

davidbiga:
Adds instantiation checking for `react-native-ble-advertiser` under `isBTActive` inside the `BroadcastingService.js`.  Without this, iOS would crash on an undefined native module.


app/services/BroadcastingService.js
```diff
@@ -167,50 +239,63 @@ function generateNewUUIDAndBroadcast() {
...
-              },
-              {
-                text: 'No',
-                onPress: () => console.log('No Pressed'),
-                style: 'cancel',
-              },
-            ],
-          ),
-        1000);
+    if (AndroidBLEAdvertiserModule == undefined) {
```
davidbiga: This is the code that was added

PR #281, opened 2020-03-29
[Dev setup] Mac fixes, and clearer instructions

tstirrat:
* Enforce homebrew on Mac
* Install Node on mac, if not there
* Only try one method of installing node on linux
  * try apt-get install, if unsuccessful, use `curl | sh` method
* Remove `~/mobileapp` refs
* Don't bail out on android studio detection on mac, it's still important to create the environment vars
* Always create env vars file, use Linux and MacOS well known SDK paths
* Use `ANDROID_SDK_ROOT` over `ANDROID_HOME`, the latter is deprecated
* Add instructions on sourcing the env config file
* Install watchman on Mac, without it, it fails when booting up the Android app with `fsevents is not a constructor`: https://github.com/facebook/create-react-app/issues/7366

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -218,10 +218,10 @@ echo "Try this to get started:"
 echo
 echo "  Run Android Studio then:"
 echo "    * Select 'Configure > AVD Manager' at the bottom"
-echo "    * Create and run a virtula device, such as a Pixel 2"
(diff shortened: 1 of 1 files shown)


carlosdp:
Looks good to me. I setup my mac using this branch and it's all in working order!

Issue #77, opened 2020-03-16
Import Locations from Google on iOS: Browser opens the Zip file in the Webview. 


vitorpamplona:
On Android, the app saves a file in the Downloads folder. 

On iOS, the WebView shows the zipped binary file in the app itself. 

We need to figure out a way to either download it to a file and then open with the SearchAndImport function or keep it in memory (500MB in memory?) and pass it to the SearchAndImport function.

1geek0:
How large can the location history file get?

vitorpamplona:
My 7 years are 500MB :(

1geek0:
Do we need that much? Can we add a date range to it?

remintz:
Just to illustrate what happens on IOS after downloading...
![IMG_903E13968C76-1](https://user-images.githubusercontent.com/4155238/77321194-95d5f780-6cf0-11ea-87a1-b57e95cb8df8.jpeg)

sergesemashko:
#69 is related to this issue, also there is WIP PR [#282 Change Import from webview to file picker](https://github.com/tripleblindmarket/covid-safe-paths/pull/282) that removes embedded webview for improved experience/privacy. 

If it was up to me, I'd prefer to keep an approach from PR [#282](https://github.com/tripleblindmarket/covid-safe-paths/pull/282) that user has to login into google account rather from browser instead of webview and being able to select file from downloads folder. 
 
So, this issue might no longer be relevant if the core and product team decides to favor the approach from #69 issue.

popofibo:
I kind of agree to replace webview with a file picker but unsure what is the timeline for that merge to happen.  It's a cleaner approach no doubt but I personally did not get a chance to test it; this maybe a stopgap to alleviate the current broken iOS experience.

tremblerz:
I was working on #282 to fix, it still doesn't work but it would be great if someone is willing to lend a hand, it is a very high priority https://github.com/tripleblindmarket/covid-safe-paths/pull/282

sergesemashko:
By the way, the solution for large takeout files might be unzipping somewhere else and import only json files for the latest months manually into the app, e.g. `2020_APRIL.json`. The can be attached.

Under the hood, we can be checking the location timestamp freshness and let's say include only locations for the past 30 days.

Issue #73, opened 2020-03-16
Google Import should expand logged points for duration


penrods:
During the import of Google data the time is imported a single time/coordinate.  However, the Google data has durations, so we should use that to expand points.  This will be useful later in the intersections.

Basically, the Google data contains something like:
```json
[ 
 {"placeVisit": {
      "location" : {
        "latitudeE7" : 423936728,
        "longitudeE7" : -710802770,
      },
      "duration" : {
        "startTimestampMs" : "1583085476383",
        "endTimestampMs" : "1583086992817"
      },
   }
}
...
]
```

The start time is:  Sunday, March 1, 2020 5:57:56.383 PM
and the end time is:  Sunday, March 1, 2020 6:23:12.817 PM

That is 26 minutes of the user at the same point, so import it in 5 minute intervals so it looks like:
[
 { startTimestampMs, latitudeE7, longitudeE7 },
 { startTimestampMs + 5 minutes, latitudeE7, longitudeE7 },
 { startTimestampMs + 10 minutes, latitudeE7, longitudeE7 },
 { startTimestampMs + 15 minutes, latitudeE7, longitudeE7 },
 { startTimestampMs + 20 minutes, latitudeE7, longitudeE7 },
 { startTimestampMs + 25 minutes, latitudeE7, longitudeE7 },
]

Don't stress too much about overlap between places, we can work with having time segments that are narrower than 5 minutes.  But we will likely have issues intersecting in encrypted space with a timespan between points that is > 5 minutes.

Additionally, the transit points aren't being imported.  It would be nice to import waypoint data.

penrods:
Also, please verify that the Google timestamps are in UTC.  I fairly sure they are, but would like this verified.

harshvitra:
```
for each placeVisit we need to check 

if --> endTimestampMs - startTimestampMs is greater than **300000** 
then number of points = ( endTimestampMs - startTimestampMs ) / 300000
Finally Add these many points with adding 300000 to startTimestampMs of previous entry
```

(300000 is the difference for 5 mins)

jordajm:
@penrods @harshvitra I'd like to grab this one if it's not currently being worked on

gabalmat:
It doesn't look like the unzipped file is getting saved to the proper location:
```
LOG  Unzip Completed for /storage/emulated/0/Download and /storage/emulated/0/Download/takeout-20200327T014023Z-001.zip
[...]

PR #282, opened 2020-03-29
Change Import from webview to file picker

rparet:
Changes the Import screen and instructions from loading a web view to instructing the customer to get their takeout data from Google and then load it into PrivateKit using a file picker. 

Tested on iOS and Android via Google Drive / iCloud Drive.


popofibo:
This effectively Resolves #77 & #114

kenpugsley:
Once we have the dust fully settled from the rename to covid safe paths, we need to consider pulling this in, possibly for v1.0

sergesemashko:
@rparet, I started digging into issues #73 and #77 the code from this PR would need to be merged first before addressing them. 

Also, Import feature is currently not really working properly in develop branch (need to fetch data for last 2 months, exceptions are thrown if files do not exist, racing conditions as async code is written in a sync way). 

Also, I'd like to offer some help to make this PR merged faster as I sort of got your idea and have an understanding how to address mentioned import issues.

tremblerz:
@sergesemashko Can you branch off from the following one -
https://github.com/PrivateKit/private-kit/tree/import_picker
Once the issues are resolved we can test it and bring it inside the Safepaths

rparet:
@sergesemashko that condition also exists on the current importer. It only looks for this month's date file, which is problematic in 2 ways: 1. at the start of the month (like now), a customer will only be able to import a short amount of date. 2. the import fails silently for people who don't have this month's data, which can be a lot of people if they turned location history off at any time in the last 12-24 months. 
I put this up as a PR because using the picker is a better solution for most customers, but if you want to make those enhancements I'd recommend surfacing to the customer when ENOENT is raised, as well as naively importing this month and last month's files.

sergesemashko:
@rparet, yeah, picker solution seems to be friendlier. I can take a look at ENOENT and parsing latest 2 months of data and will open a PR soon.

tstirrat:
closing in favor of rebased #410

PR #283, opened 2020-03-30
Export fix for iOS

tremblerz:
The export of file in iOS does not work directly from base64 due to the encoding performed by sharing application in order to add a file as an attachment.
This PR solves it in the following way - 
1. Write file on the disk first
2. Share through the url of this file
3. Remove the file as soon as it is shared.


app/views/Export.js
```diff
@@ -60,50 +61,54 @@ function ExportScreen({ shareButtonDisabled }) {
...
+      }
+      await Share.open(options)
         .then(res => {
           console.log(res);
         })
         .catch(err => {
+          console.log(err);
           console.log(err.message, err.code);
         });
+      if (Platform.OS === 'ios') {
```
tremblerz: For removing file

tremblerz:
Tested on both Android and iOS

app/views/Export.js
```diff
@@ -60,50 +61,54 @@ function ExportScreen({ shareButtonDisabled }) {
   async function onShare() {
     try {
       let locationData = await new LocationData().getLocationData();
-
-      const jsonData = base64.encode(JSON.stringify(locationData));
+      var options = {};
+      var jsonData = JSON.stringify(locationData);
       const title = 'PrivateKit_.json';
-      const filename = 'PrivacyKit_.json';
+      const filename = 'PrivateKit_.json';
```
penrods: Let's make this a unique filename -- it makes the file management easier for a Redactor who is saving them all in the same folder.  This can be a simple Unix Timestamp.

PR #295, opened 2020-03-31
Improving pt br translation

MatheusAlvesAlmeida:
I realized that some improvements could be made in the Brazilian Portuguese translation. I hope it helps!

### DIFF
--- a/app/locales/pt_BR/exportscreen.json
+++ b/app/locales/pt_BR/exportscreen.json
@@ -1,11 +1,11 @@
 {
-    "export_para_1":"Você pode compartilhar sua trilha de localização com qualquer pessoa usando o botão Compartilhar abaixo. Depois de pressionar o botão, ele perguntará com quem e como você deseja compartilhá-lo.",
+    "export_para_1":"Você pode compartilhar sua trilha de localização com qualquer pessoa usando o botão Compartilhar abaixo. Depois de pressionar o botão, ele perguntará com quem e como você deseja compartilhá-la.",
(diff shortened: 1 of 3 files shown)


alpita-masurkar:
Hi Matheus,
Thank you for going through these and suggesting updates. While you're on this, there were two keys in the overlap.json whose values got updated:
These are the keys in en (English). Can you also translate them and update them in those files?
"overlap_para_1": "The green trail represents your location history\n\nThe light purple circles represent the public dataset",
  "show_overlap": "Click to view the public dataset",

MatheusAlvesAlmeida:
In the live app, the text content of these keys is different from the one you sent me. I translated and updated the ones you sent me. If this is not what you wanted, excuse me, I will correct it.

alpita-masurkar:
> In the live app, the text content of these keys is different from the one you sent me. I translated and updated the ones you sent me. If this is not what you wanted, excuse me, I will correct it.

Hi, as you can see in this file,
https://github.com/tripleblindmarket/private-kit/blob/develop/app/locales/en/overlap.json
the translations were updated for the keys that I asked you to translate. Thank you for doing the translations promptly.

kenpugsley:
Thank you for this translation.  The UI has changed significantly and has now finalized for the v1.0 release, which has resulted in a great deal of changes to the text in the final app.  Because of this we found that it was best to remove the existing translations and start again.  Even though we are also closing this PR, we hope you understand and will help us again.

The text in the app is now complete, so **we're asking for your help to redo this translation on the text for the v1.0 release**.  Our translation lead @alpita-masurkar will be coordinating this effort.

Thank you from the COVID Safe Paths dev team.

PR #296, opened 2020-03-31
Yoruba support from Ojedokun Seyi

alpita-masurkar:


kenpugsley:
Thank you for this translation.  The UI has changed significantly and has now finalized for the v1.0 release, which has resulted in a great deal of changes to the text in the final app.  Because of this we found that it was best to remove the existing translations and start again.  Even though we are also closing this PR, we hope you understand and will help us again.

The text in the app is now complete, so **we're asking for your help to redo this translation on the text for the v1.0 release**.  Our translation lead @alpita-masurkar will be coordinating this effort.

Thank you from the COVID Safe Paths dev team.

PR #312, opened 2020-04-01
Export fix for iOS

tremblerz:
Due to some reason url got changed to `RNFS.Bundle` from `RNFS.MainBundlePath`


penrods:
The change looks OK, but I don't have an iPhone to validate behavior.  Can you explain why the change from Bundle to MainBundlePath?  The comment in he suggests it causes problem on Android -- have you tried it on there, too?
https://github.com/itinance/react-native-fs#basic

tremblerz:
Has been fixed in another pr

PR #315, opened 2020-04-02
iOS Google Takeout file Download Fixes #77

### LINKED ISSUE #77: Import Locations from Google on iOS: Browser opens the Zip file in the Webview.

popofibo:
Fixes #77 

This existing react-native-webview bug will be auto-fixed in their next release with [this PR](https://github.com/react-native-community/react-native-webview/pull/1214). This fix will be a workaround to handle file downloads for iOS by:

- Intercepting WebView call to check for iOS and if url equates to takeout download url (onShouldStartLoadWithRequest)
- Creates a directory within AppData documents
- Saves the takeout file and unzips as normal

Tested on iOS and Android

### DIFF
--- a/app/helpers/GoogleTakeOutAutoImport.js
+++ b/app/helpers/GoogleTakeOutAutoImport.js
@@ -3,6 +3,7 @@
  */
 import { unzip, subscribe } from 'react-native-zip-archive';
 import { MergeJSONWithLocalData } from '../helpers/GoogleData';
+import { Platform } from 'react-native';
 
 // require the module
 let RNFS = require('react-native-fs');
@@ -12,6 +13,12 @@ let progress;
 
 // Google Takout File Format.
 let takeoutZip = /^takeout[\w,\s-]+\.zip$/gm;
+
+// Download directory based on Platform OS
+var downloadDir =
+  Platform.OS === 'ios'
+    ? RNFS.DocumentDirectoryPath + '/Downloads/'
+    : RNFS.DownloadDirectoryPath;
 
 // Gets Path of the location file for the current month.
 function GetFileName() {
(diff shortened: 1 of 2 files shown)


kenpugsley:
We need to review this to determine if it's required for the v1.0 release.  @popofibo - do you know if iOS export is fundamentally broken without this change?

popofibo:
> We need to review this to determine if it's required for the v1.0 release. @popofibo - do you know if iOS export is fundamentally broken without this change?

iOS _import_ with react native webview is definitely broken w/o this change; I referenced to the link in my comment above of the PR to core react native webview package that fixes it more gracefully. The core webview change essentially adds an optional callback `onFileDownload`, my code above essentially takes a similar approach but uses available `onShouldStartLoadWithRequest` to intercept and create a downloads folder within the application, and save the file to be utilized.

popofibo:
cc: @kenpugsley @tremblerz if this is to be merged before it starts to conflict

kenpugsley:
Leaving this PR, but moving to "next next" the version right after v1.0.  Since final translations are already underway, I think the best plan is to merge this in to develop right after the rc-1.0 branch is created.  That will put this functionality into the first update after v1.0.

tstirrat:
This will be superseded by #410 as we will remove the webview.

sergesemashko:
I think this PR can be closed in favor of https://github.com/tripleblindmarket/covid-safe-paths/pull/410 as it also takes care of issue #77.

PR #329, opened 2020-04-03
Make some hardcoded texts translatable

filkaris:
I found some texts lying around that were hardcoded and not under a translation key, so they were not translatable

I skipped `Welcome.js` because as far as I can tell it's not used anywhere (and could be deleted?)

Also, as a topic of discussion, maybe translations should all be in one flat json file?

- `app/locales/en/index.js` is going to keep growing without offering any real value
- It's not clear where a key is located, so you will have to check multiple files
- If one key is used twice, the latter will overwrite the former, which will be weird to debug for someone who is not familiar with how JS object destructuring works


harshvitra:
> * If one key is used twice, the latter will overwrite the former, which will be weird to debug for someone who is not familiar with how JS object destructuring works

The points are valid, but modularity is also needed. 

This is a simple developer experience challenge. Some devs will find this easy and some devs will find separate files easier.

harshvitra:
Resolved the conflicts with Develop

LGTM + 

@tremblerz ready for merge

app/locales/en/locationService.json
```diff
@@ -0,0 +1,13 @@
+{
+    "yes": "Yes",
+    "no": "No",
+    "location_enabled_title": "Private Kit Enabled",
+    "location_enabled_message": "Private Kit is securely storing your GPS coordinates once every five minutes on this device.",
```
tstirrat, line 5: Please update these to `COVID Safe Paths` like in #331 , if you have any specific Q's about a specific translation, @penrods can help.

PR #331, opened 2020-04-03
"Rebranding" from Private Kit to: COVID Safe Paths

penrods:
Beginning the process of publicly being COVID Safe Paths


tstirrat:
#327 for reference

PR #341, opened 2020-04-04
Renaming the project and project files

harshvitra:
Renamed all the assets from Private Kit to COVID Safe Paths

The changes will be visible on installing the application. 
It will be installed with a new Package name

IMG_20200404_130658


tstirrat:
Getting way out of date, and we're doing parts of this in smaller CLs. Closing for now

PR #355, opened 2020-04-05
Add unobtrusive locale selection dropdown to onboarding (WIP)

StartupSam:
A dropdown locale / language select for areas where phone language settings doesn't usually match spoken language. Primarily in prep for our pilot program. This PR is a work in progress. Brief summary:
- Set up with the native picker components for best ease of use. A bit annoying to handle on iOS (especially since I can't debug there)
- Uses a list of all prepared languages pulled from the i18next configuration
- If a language is manually selected, that choice is put in async storage and the app will override the phone setting with the user setting

Still to do:
- I don't have easy access to a macbook/iPhone, so I need some help to verify that the Picker component on iOS is working as expected
- Add unicode flags to the language labels
- Style both Android and iOS components to match v2 style in incoming PR (see Figma mockup for target)

Screenshots:
Screenshot from 2020-04-05 06-07-10
Screenshot from 2020-04-05 06-07-28

### DIFF
--- a/app/components/NativePicker.js
+++ b/app/components/NativePicker.js
@@ -1,0 +1,120 @@
+import React, { Component } from 'react';
+import {
+  View,
+  Text,
+  TouchableOpacity,
+  TextInput,
+  TouchableWithoutFeedback,
+  Picker,
+  Modal,
+  Platform,
+  StyleSheet,
+} from 'react-native';
+
+// Code for the language select dropdown, for nice native handling on both iOS and Android.
+// TODO: Move component to a separate file
+export default class NativePicker extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      modalVisible: false,
+    };
+  }
+
+  render() {
+    if (Platform.OS === 'android') {
+      return (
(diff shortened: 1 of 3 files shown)


penrods:
Also, can you add unicode to these?  If so, just put a Unicode Flag as the first character on each of these.

http://xahlee.info/comp/unicode_flags.html

StartupSam:
Quick update: After some research, it seems to make a lot more sense to ditch the flags in the language picker. Since we're trying to be as politically acceptable as possible, there are too many variables and contentions around flags. This mostly stems from the fact that they represent countries, not languages.

See here for further reading: http://www.flagsarenotlanguages.com/blog/why-flags-do-not-represent-language/

Couldn't even find an unofficial list somewhere. Am going to leave out the flags for this PR.

penrods:
I agree with leaving out the Flag.  Makes sense.

tremblerz:
Where I come from almost every state has its own language :)

penrods:
#367 redid this in the new UI, closing this one

PR #356, opened 2020-04-05
Merge translation strings into a single file

penrods:
The translation scheme had a few downfalls:
* It was easy to create overlapped keys when the files were independent
* Translators had to manage multiple files, transmission was tricky
* Reviewers had to jump around, also.  Especially with global strings.

This merges them into a single file to solve those problems.

 ### Localization Notes ###
All of the language files can be mirrored off of this.  Given the major
UI changes, I think we really should throw away the old file and start 
from scratch on each language, unfortunately.


alpita-masurkar:
@penrods I agree with you. I think having a single global file for all strings is a great idea.
I was always creating a single file out of all the files for the translators whenever I gave them the content.
This will really solve a lot of the issues that you mentioned and create lesser code and content to manage.

But related to this, I just wanted to mention something and see what you think about it:
We have a lot of content and we have to consider the possibility that it will increase or get updated as new features get added. It might be trickier at some point to figure out sections of the file or keys that you want to work with.

My point is and there's probably a lot of disagreement about this-- pure json doesn't support commenting, but if this file grows really long, it might be helpful to have comments at some point (JSON5, HOCON or some human readable format and a parser or maybe even something like yaml...). If this is not the case, we can safely ignore this, but if there's a possibility that this file grows long, we should think about this.

kenpugsley:
Created an issue to track this need.  #357

PR #359, opened 2020-04-05
Separate and parallelize builds, build APK for easy testing

tstirrat:
* Separate iOS and Android builds, so that they can run in parallel
  * 10-14 mins down to 4-6 mins
  * more clear if only one of ios or android is failing
* Allow download of apk files for drag and drop into emulator
  * Screen Shot 2020-04-06 at 10 04 55 PM

* Cache `node_modules/` for faster builds (10m builds down to 4-6 mins, unit tests reduced by 56 seconds)


tstirrat:
@kenpugsley this is ready to be merged. 

We can and should enable branch protection for all 3 of these new builds now that they're faster and separate

tstirrat:
the ios build can be improved but it's a much more complicated setup with release signing. Will look into this in a later pr

tstirrat:
Yep, agreed, let's let it bake and see re apk builds

PR #366, opened 2020-04-06
Start of news from Healthcare Authorities

penrods:
When you subscribe to a Healthcare Authority, it should show a webpage
from that authority in the web view.  Since an individual might be
interested in multiple Healthcare Authorities, this should support having
multiple news pages.

This implementation uses a swiper with a titlebar that displays the
healthcare authority name.  The last entry should always be the
Safe Paths default page.

* Tweak look to match the new UI


penrods:
NOTE:  This is currently hard-coded to show the Haitian website as a 2nd.

PR #367, opened 2020-04-06
Add locale picker, Haitian Creole translation

penrods:
Add a language picker on the first screen of the Welcome sequence.  This
allows nonstandard languages to be easily selected.

This is based on #355 by @StartupSam, but merged into the new UI

 ### Testing Notes ###

Start the welcome sequence after startup.  Select Haitian Creole (Angajman)

 ### Localization Notes ###

This adds Haitian Creole from Gerald

Issue #263, opened 2020-03-28
FEATURE: Select locale


penrods:
The current localization scheme uses the language of the phone to pick a locale.  However, some locations have a legal language, but many in the population actually speak a variation -- e.g. a creole.

Support this functionality by allowing the user to pick the Locale in the first page of the welcome sequence.  The local should be indicated by both text and (ideally) a Flag icon.  This is a small dropdown addition to the existing first page, not an extra page.

The Settings page should also allow switching locales.

rparet:
I understand the need for this in places like Haiti. However, are we imposing this across the app? If yes, can we justify the additional onboarding friction for most customers? Is it possible to do this in a more elegant way than in the welcome sequence?

penrods:
It will be a little dropdown, much like you see in websites.  Kinda like language picker in the upper right of this website:  https://www.similarweb.com/ja/website/nicovideo.jp

penrods:
I updated the description to clarify that this isn't a completely new page, just a small addition to the existing first page.

andreanuzzo:
How about a choice at the first launch of the app?

saranyaharidas:
Looks like a duplicate issue ?

#355

PR #368, opened 2020-04-06
Add v2 UI styled locale picker, Haitian Creole translation

StartupSam:
Building off of my earlier PR #355 and @penrods merge into the new UI with #367 (on issue #263), I've styled the locale select picker component to match the Figma design on both iOS and Android.

I still need to make sure the picker component works as expected on iOS, could use someone's help checking that.

image


StartupSam:
@kenpugsley @penrods This was merged into Steve's branch, I think it still needs to be merged into develop

PR #369, opened 2020-04-06
Fix the tests with all the UI changes

tremblerz:


app/views/LocationTracking.js
```diff
@@ -86,8 +86,12 @@ class LocationTracking extends Component {
       isLogging: '',
       currentState: StateEnum.NO_CONTACT,
     };
-
-    this.checkCurrentState();
+    try {
```
tremblerz, line 89: If we don't add this try catch then tests don't work as `PERMISSIONS` is not imported in the jest mock

PR #378, opened 2020-04-07
Migrate intersection logic into background service

tremblerz:
Summary:
- Fixes the intersection tick by sending it in the background fetch service.
- Applies fix to some of the intersection scheme code.

Runs fine on Android but the iOS is a tricky bit as it runs sporadically in the background fetch service.


app/services/BackgroundTaskService.js
```diff
@@ -0,0 +1,123 @@
...
+              });
+
+            SetStoreData('AUTHORITY_NEWS', name_news)
+              .then(() => {
+                // TODO: Anything after this saves?  Background caching of
+                //       news to make it snappy?  Could be a problem in some
+                //       locales with high data costs.
+              })
+              .catch(error =>
+                console.log('Failed to save authority/news URL list'),
```
denispapakul, line 78: Probably, make sense to pass is "error" prop to log for understanding, which kind of error we have

app/services/BackgroundTaskService.js
```diff
@@ -0,0 +1,123 @@
+import BackgroundFetch from 'react-native-background-fetch';
+import PushNotification from 'react-native-push-notification';
+import { IntersectSet } from '../helpers/Intersect';
+import { GetStoreData, SetStoreData } from '../helpers/General';
+import { isPlatformiOS } from './../Util';
+
+const INTERSECT_INTERVAL = 30 / 60; // In Minutes
```
denispapakul, line 7: I think to make sense a create like 'constants.js' file for all constants of the app.

tremblerz: that's a good point, makes it much more consistent. I think a good idea would be to port all of them in a separate PR

app/services/BackgroundTaskService.js
tremblerz: Makes sense, will add.

app/services/BackgroundTaskService.js
denispapakul: Yeah, I will create a new issue for it and will resolve it in a separate PR. Thanks!

app/services/BackgroundTaskService.js
```diff
@@ -0,0 +1,135 @@
...
+import PushNotification from 'react-native-push-notification';
+import { IntersectSet } from '../helpers/Intersect';
+import { GetStoreData, SetStoreData } from '../helpers/General';
+import { isPlatformiOS } from './../Util';
+import languages from '../locales/languages';
+
+const INTERSECT_INTERVAL = 30 / 60; // In Minutes
+
+export function executeTask() {
+  check_intersect = () => {
```
kenpugsley, line 11: I think `check_intersect` should be defined over in `Intersect.js`.  I think it's OK to keep the executeTask wrapper for the background service to call ... keeps it a bit cleaner by separating out what will execute from the service definition, which I think is good.

app/views/LocationTracking.js
```diff
@@ -25,8 +25,11 @@ import BackgroundImageAtRisk from './../assets/images/backgroundAtRisk.png';
 import Colors from '../constants/colors';
 import LocationServices from '../services/LocationService';
 //import BroadcastingServices from '../services/BroadcastingService';
+import BackgroundTaskServices, {
+  executeTask,
```
kenpugsley, line 29: See the other note about this.  I think `check_intersect` could be in `Intersect.js`, and used here directly.  Would help with readability for sure.

app/services/BackgroundTaskService.js
```diff
@@ -0,0 +1,135 @@
...
+                //       notification if that changes, i.e. if there is a newsflash?
+                name_news.push({
+                  name: responseJson.authority_name,
+                  news_url: responseJson.info_website,
+                });
+
+                // TODO: Look at "publish_date_utc".  We should notify users if
+                //       their authority is no longer functioning.)
+
+                IntersectSet(responseJson.concern_points, dayBin => {
```
kenpugsley, line 67: Is there an issue here with calls that can happen via the background service happening concurrently with the UI?  Seems possible that the user could have the background service chugging away at the Intersect, while the user opens the app and the UI thread _also_ kicks off doing it.  Might end up with a race condition and the user getting notified twice in succession.  I guess it all depends upon how long the `IntersectSet` takes.

PR #382, opened 2020-04-07
New versions of the Safe Paths icon -- linked location pointers

penrods:
This starts the work to implement the new application icon throughout the
application on both iOS and Android.  Still more to be done.

P.S. I <3 the new icon.  :)


penrods:
They changed the icon on me!  But the new new one is better anyway, so totally worth it.

Implemented in #386, closing this unmerged.

PR #389, opened 2020-04-08
Location saving at proper rate; disable bluetooth

kenpugsley:
Fixes for getting the location saving to run at sane rates.   Includes:
- mimimum time allowed between saves (to protect from flooding the location log)
- maximum allowed backfill time
- Fixed backfill logic
- removed duplicate listeners, also the stationary listener that leads to duplicate location updates

Also disabled the start and stop of the BroadcastService to turn off bluetooth for v1


vitorpamplona:
Bluetooth Disabling code is Good. Not sure about the Location Tracking. Abhi?

PR #393, opened 2020-04-08
Use best practices for variables #385

### LINKED ISSUE #385: Store all constants in one file
I suggest storing all constants in one file in order to avoid code repetition and to use re-use of variables. This problem looks like this - 
Screen Shot 2020-04-08 at 12 57 00 AM
Can implement it.

denispapakul:
Fixed #385


app/helpers/Constants.js
```diff
@@ -0,0 +1,3 @@
+import { Dimensions } from 'react-native';
+
+export const { width, height } = Dimensions.get('window');
```
tstirrat, line 3: Dimensions could change:

> Although dimensions are available immediately, they may change (e.g due to device rotation, foldable devices etc) so any rendering logic or styles that depend on these constants should try to call this function on every render, rather than caching the value (for example, using inline styles rather than setting a value in a StyleSheet).

app/helpers/config.js
```diff
@@ -0,0 +1,5 @@
+export const public_data =
```
tstirrat, line 1: use CONST_CASE for consts while we're here.

move to `constants/authorities.js`

app/helpers/Constants.js
```diff
@@ -0,0 +1,3 @@
+import { Dimensions } from 'react-native';
```
tstirrat, line 1: helpers suggests "test helpers", prefer `constants`

we already have a `constants/`folder, suggest `constants/layout.js`

tstirrat: Oops, obsolete comment.

tstirrat:
Can you import the `useWindowDimensions` hook where possible?

app/helpers/config.js
denispapakul: Sure, thanks

app/constants/constants.js
```diff
@@ -0,0 +1,6 @@
+export const LOCATION_DATA = 'LOCATION_DATA';
```
tstirrat, line 1: `constants/constants.js` is vague enough that any constant can go here. and that means this file will quickly become dumping ground of every constant.

suggest `constants/storage.js`

app/helpers/Constants.js
denispapakul: Will do it a separate PR, thanks for advice

app/helpers/authorities.js
```diff
@@ -0,0 +1,5 @@
+export const PUBLIC_DATA_URL =
```
tstirrat, line 1: move to constants/authorities.js

tstirrat:
One minor change, otherwise great

app/helpers/authorities.js
denispapakul: Sure, thanks

tstirrat:
Please `git merge develop` in _your branch_ and then push again. It seems to have been a bad merge.

PR #409, opened 2020-04-08
Update iOS icon

tstirrat:
Screen Shot 2020-04-08 at 4 29 42 PM


tstirrat:
I reverted the splash screen change because it looked horrible on smaller devices

PR #410, opened 2020-04-09
(Fix) Improve Google Location Data import Experience, fix import logic

### LINKED ISSUE #69: Improve Google Location Data import experience
The Import functionality is great, but still tricky to use.  It might take a few iterations to get it to what we want, but here are some thoughts:

Break the Import into three stages, kinda like an old-school Wizard.
  1) "Intro" page.
     Explain what this does and give instructions to log in and start the process w/Google.
     Q: Can we explore the DOM on the webview?  It would be cool to watch for an indicator that the request for a takeout has happened, then move to the "Waiting" step described below.
  2) "Waiting on Google takeout" page
      If the user returns to Import page, show a message that we are still waiting on the takeout file.
      Include a "Start over" button to go back to Step 1 in case something fails.
  3) "Import Complete"
      After an import completes, the results of the import should be remembered and displayed.  E.g.
      "28 day of data imported from Mar 1 - Mar 28".
      Also have an "Import again" button that goes back to step 1.

During Step 2 the user should be be able to leave the app or go back to the main page.  We should have a background task look for the takeout file when one is pending, so you don't have to go back to Import.  The app should show a notification when the takeout file is found and process completes.

Discussion is welcome.  I expect the implementer will find challenges and opportunities to improve this sequence.  For instance, there might be a step 2.5 that helps with the download of the Takeout file.

### LINKED ISSUE #77: Import Locations from Google on iOS: Browser opens the Zip file in the Webview.

sergesemashko:
Rebase of https://github.com/tripleblindmarket/covid-safe-paths/pull/282.

This PR addresses 2 issues:
- [Improve Google Location Data import experience](https://github.com/tripleblindmarket/covid-safe-paths/issues/69)
- [Import Locations from Google on iOS: Browser opens the Zip file in the Webview.](https://github.com/tripleblindmarket/covid-safe-paths/issues/77)

Other things done besides fixes above: 
- Add logic to import last 2 months from Google takeout as safe paths is concerned only about  locations for recent couple of weeks
- Add checks if files exist before reading
- Add info messages on import complete:

Success:
Screen Shot 2020-04-08 at 5 39 29 PM

Already imported:
Screen Shot 2020-04-08 at 5 37 41 PM

No recent locations:
Screen Shot 2020-04-08 at 5 37 35 PM

Select file screen if downloaded locally:
Screen Shot 2020-04-08 at 5 29 13 PM

Closes #77 
Closes #69 
Closes #315


app/services/LocationService.js
```diff
@@ -17,8 +17,7 @@ export class LocationData {
     this.minLocationSaveInterval = Math.floor(this.locationInterval * 0.8); // Minimum time between location information saves.  60000*4 = 4 minutes
```
sergesemashko, line 17: this file contains only auto fixes by prettier

app/helpers/GoogleData.js
```diff
@@ -3,67 +3,54 @@
...
-    data,
-    //index
-  ) {
-    // Only import visited places, not paths for now
-    if (data.placeVisit) {
-      let loc = BuildLocalFormat(data.placeVisit);
-      InsertIfNew(localDataJSON, loc);
-    }
-  });
+function extractNewLocations(storedLocations, googleLocationHistory) {
```
tstirrat, line 28: Can you add some unit tests for the changes in here? I know there are none yet, but now's a good time to get something there for others to build on.

app/helpers/GoogleTakeOutAutoImport.js
```diff
@@ -2,107 +2,109 @@
...
+    );
+  });
 }
 
-export async function SearchAndImport() {
-  //googleLocationJSON
-  console.log('Auto-import start');
+// Imports any Takeout location data
+// Currently works for Google Takeout Location data
+export async function ImportTakeoutData(filePath) {
```
tstirrat, line 49: Add some unit tests, thanks.

app/views/Import.js
```diff
@@ -1,115 +1,110 @@
...
-  };
+const makeImportResults = (label = '', error = false) => ({
+  error,
+  label,
+});
 
-  hideSpinner() {
-    this.setState({ visible: false });
-  }
+const ImportScreen = props => {
```
tstirrat, line 29: Changing to FC  and using hooks = ❤️ thanks

app/views/__tests__/Import.spec.js
```diff
@@ -5,9 +5,13 @@ import React from 'react';
...
 import Import from '../Import';
 
+const navigationMock = {
+  goBack: jest.fn()
+};
+
 it('renders correctly', () => {
   const tree = renderer
-    .create(<Import />)
+    .create(<Import navigation={navigationMock} />)
```
tstirrat, line 14: A simple test of clicking  the right button opening a picker would be helpful, thanks.

app/locales/en/english.json
```diff
@@ -32,10 +32,15 @@
...
 
 
-    "import_title": "Import Locations",
-    "import_step_1": "1. Login to your Google Account and Download your Location History",
-    "import_step_2": "2. After downloaded, open this screen again. The data will import automatically.",
-
+    "import_title":"Import Locations",
+    "import_step_1":"Adding location data from Google will give you a head start on building your recent locations.",
+    "import_step_2":"Before you can import, you must first 'Take out' your location data from Google.",
+    "import_step_3":"Visit Google Takeout and export your Location History using following settings: \n1. Delivery method: \"Google Drive\" \n2. Frequency: \"Export once\" \n3. File type & size: \".zip\" and \"1GB\" \n4. Return here to import the archive",
```
tstirrat, line 38: I think there is an extra step of "Google sends you an email to tell you the export is ready", right?

tstirrat:
Linking #77 & #114 from closed #282

tstirrat:
I think it would be good to disable the import button if we already know import is useless (i.e. already imported, or no import files available)

kenpugsley:
To keep the codeline (including translations) stable for v1.0, we won't merge this into develop until the rc-1.0 branch is created.  Would like to pull this in right after that happens so that this does into the first point release after v1.0.

tstirrat:
I'll update and get this merged now

sergesemashko:
@tstirrat, could you hold on, I'm about to push the unit tests

tstirrat:
Yep, please do :)

sergesemashko:
@tstirrat , thanks for your time reviewing this PR. The feedback has been reflected. Could you take a look one more time?

android/app/build.gradle
```diff
@@ -161,7 +161,6 @@ android {
                 project.logger.lifecycle('Could not find release key')
             }
         }
-
```
tstirrat, line 164: Remove this diff

tstirrat:
Yep, looks good.

app/locales/en.json
```diff
@@ -33,9 +33,16 @@
     "home_no_contact_subtext": "Based on available data you haven’t been near anyone reported positive for COVID-19.",
     "home_unknown_header": "Unknown",
     "home_unknown_subtext": "We can’t tell if you’re at risk unless you enable the app to access your location.",
-    "import_step_1": "1. Login to your Google Account and Download your Location History",
-    "import_step_2": "2. After downloaded, open this screen again. The data will import automatically.",
-    "import_title": "Import Locations",
+    "import_title":"Import Locations",
+    "import_step_1":"Adding location data from Google will give you a head start on building your recent locations.",
+    "import_step_2":"Before you can import, you must first 'Take out' your location data from Google.",
+    "import_step_3":"Visit Google Takeout and export your Location History using following settings: \n1. Delivery method: \"Google Drive\" \n2. Frequency: \"Export once\" \n3. File type & size: \".zip\" and \"1GB\" \n4. Google sends an email when the export is ready \n5. Return here to import the archive",
```
tstirrat, line 39: Delivery method is "Add to Drive", not Google Drive

Screen Shot 2020-04-10 at 1 51 59 PM

tstirrat:
Actually one minor text change, then we're good to go

sergesemashko:
@tstirrat , the text is updated.

tstirrat:
In my testing on a pixel 2 emulator, it cannot select the zip files even though I found them in the drive. 

The picker is not totally straight forward.

tstirrat:
I suspect this needs more explanation in the 1 - 3 steps. I'll try testing it on another android device now

tstirrat:
I was able to select the zip by going through a different Drive picker, and now there's no indication that anything happened:

Screen Shot 2020-04-10 at 2 42 42 PM


What should happen directly after selecting the zip file?

sergesemashko:
@tstirrat, thanks for testing, there should be either success or error message on completion. I'll double check.

sergesemashko:
@tstirrat, I fixed the Android behavior and slightly updated the Import instructions. 

Let me know what do you think.
Screen Shot 2020-04-12 at 4 00 35 PM

diarmidmackenzie:
I'm not able to tell from this issue quite how substantial a rework this is, and therefore whether any problems from the previous UX might persist.

Here are some notes on flaws with the existing experience.  To save further test iterations, could we review & confirm that all these issues will be addressed by this rework?  Thanks!

https://github.com/tripleblindmarket/covid-safe-paths/issues/69#issuecomment-612687784

app/views/__tests__/Import.spec.js
```diff
@@ -1,13 +1,51 @@
...
 import 'react-native';
+import {Linking} from 'react-native';
 import React from 'react';
 import renderer from 'react-test-renderer';
 import Import from '../Import';
 
-it('renders correctly', () => {
-  const tree = renderer
-    .create(<Import />)
+jest.spyOn(console, 'log').mockImplementation(() => {});
```
tstirrat, line 8: Does this act globally?

I think these `jest.spyOn()` should be in a `beforeAll()` or `beforeEach` block

app/views/__tests__/Import.spec.js
```diff
@@ -1,13 +1,51 @@
...
+};
+const openURLSpy = jest.spyOn(Linking, 'openURL').mockImplementation(() => Promise.resolve());
+const importTakeoutDataSpy = jest.spyOn(GoogleTakeOutAutoImport, 'importTakeoutData')
+  .mockImplementation(() => Promise.resolve([]));
+const pickFileSpy = jest.spyOn(General, 'pickFile')
+  .mockImplementation(() => Promise.resolve('file://tmp/test.zip'));
+
+describe('Import component', () => {
+  afterAll(() => {
+    openURLSpy.mockReset();
```
tstirrat, line 24: given they are here in an afterAll,put the mocks in a beforeAll

app/helpers/tests/GoogleTakeOutAutoImport.spec.js
```diff
@@ -0,0 +1,193 @@
+import { makeTimelineObject } from './GoogleTakeoutUtils';
+const {
+  location1,
+  location2,
+  location3,
+  location4,
+} = require('./fixtures/googleImport');
+
+jest.spyOn(console, 'log').mockImplementation(() => {});
```
tstirrat, line 9: put in beforeAll or beforeEach

app/helpers/tests/GoogleData.spec.js
```diff
@@ -0,0 +1,47 @@
...
+const GOOGLE_IMPORT = {
+  timelineObjects: [
+    makeTimelineObject(location2),
+    makeTimelineObject(location3),
+    makeTimelineObject(location4),
+  ],
+};
+
+describe('mergeJSONWithLocalData', () => {
+  beforeEach(() => jest.resetModules());
```
tstirrat, line 13: Seems like a code smell. is this needed?

Perhaps this is because of the module scoped jest.spyOn in other files?

ten-sis: this is needed because each test is mocking the same dependency, probably better approach is to reset mocked module explicitly

ten-sis: there is `jest.doMock('../../helpers/General'` in every test and mocks has to be reset in order to apply a new mock implementation.
BTW, from [.resetModules() docs](https://jestjs.io/docs/en/configuration#resetmodules-boolean):

> By default, each test file gets its own independent module registry. Enabling resetModules goes a step further and resets the module registry before running each individual test. This is useful to isolate modules for every test so that local module state doesn't conflict between tests. This can be done programmatically using jest.resetModules().

Which means .spyOn() called in other tests should not affect anything here

app/views/__tests__/Import.spec.js
ten-sis: looks like this is applied only within the context of this file. [more details in .resetModules() docs](https://jestjs.io/docs/en/configuration#resetmodules-boolean)

app/helpers/tests/GoogleData.spec.js
tstirrat: ok great :)

ten-sis:
@tstirrat, I addressed the feedback and slightly updated intructions wording. could you please re-review?

app/helpers/GoogleData.js
```diff
@@ -4,67 +4,73 @@ import { LOCATION_DATA } from '../constants/storage';
...
-    time: placeVisit.duration.startTimestampMs,
-  });
+/**
+ * Rounds float number to a desired number of decimal places and returns a float
+ * number. NOTE: .toFixed() returns a string, but number is required.
+ * @param num - number
+ * @param digits - amount of digits to round
+ * @returns {number}
+ */
+function toFixedNumber(num, digits) {
```
tstirrat, line 14: [toPrecision](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision) returns a string, but [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) returns a number.

app/helpers/GoogleTakeOutAutoImport.js
```diff
@@ -2,108 +2,141 @@
...
+/**
+ * Safe paths is interested in locations for latest a couple of weeks.
+ * Date for latest 2 months should be sufficient to cover all cases,
+ * especially the case when we are in the early days of the current month.
+ * @returns {string[]} - array of files for latest 2 months from google takeout archive
+ */
+export function getFilenamesForLatest2Months(rootPath) {
+  const now = new Date();
+  const previousMonth = new Date();
+  previousMonth.setMonth(now.getMonth() - 1);
```
tstirrat, line 37: This totally works. I had to check manually 👍 . Can you add a test that checks this year wrapping case  (i.e its correct in Jan), so that any mods to this code are covered?

app/helpers/GoogleData.js
sergesemashko: they actually both return "string", I think MDN also says the returning value is a string:

Untitled 2

app/helpers/GoogleTakeOutAutoImport.js
sergesemashko: good catch, thanks. I'll update the logic to rather use MomentJS and cover it in unit test

app/helpers/GoogleData.js
tstirrat: What a trip! Chrome was a number for me. What browser is that? 

Or perhaps it's transpiled incorrectly...

app/helpers/GoogleTakeOutAutoImport.js
tstirrat: We now use dayJS, which is an immutable momentJS clone

app/helpers/GoogleData.js
tstirrat: Bizarre... resolving this.

tstirrat:
Once we get tests around the 2-month window I am ready to merge this. I don't see any other blockers.

sergesemashko:
@kenpugsley , @tstirrat , I address the January edge case with last 2 months calculation, could you re-review?

kenpugsley:
Nice work!  Thanks for pushing to get this one completed!

PR #417, opened 2020-04-09
iOS splash screen update

tstirrat:
thanks @christospappas


tstirrat:
Fixing up the font before merge

tstirrat:
Even with the font fix, there was a flash/pop when the font is not loaded so it looked weird.  @christospappas is going to encode the text into the bg image and let us know when ready

PR #419, opened 2020-04-09
Add helper for i18n extraction

tstirrat:
* Consolidated all translations into single files using `jq`;
* Adds an extraction tool: `npm run i18n:extract`
  * adds new keys to translation files like so:

    ```
    {
      "labels": {
        "new_key": "",
    ```

  * removes old keys into a backup `_old.json` file, in case they are still useful.

To keep this PR from reaching everywhere, I have not changed the keys we use in the app, yet. In a follow up PR it would be good to change the namespace from `label.x` to `settings.x` so that messages might be grouped better in the json file (currently they are sorted asc)

TODO: update/add i18n docs


tstirrat:
Build failure should be fixed in #420 🚬

app/locales/en/index.json
```diff
@@ -1,163 +1,98 @@
 {
```
penrods, line 1: I personally liked having "english.json" so if I got files back from someone they weren't all called "index.json".  Not a showstopper, but if that is possible I think it helps @alpita-masurkar manage this better.

app/locales/es/index.json
```diff
@@ -0,0 +1,23 @@
+{
```
penrods, line 1: Are these strings that survived from a previous translation?  I was thinking it might be most useful to have the translation files full of the full English phrases -- then the translator just needs to walk through the single file and translate as they go.  But I am presuming about how this gets done. @alpita-masurkar ?

app/locales/languages.js
```diff
@@ -3,29 +3,32 @@ import { getLanguages } from 'react-native-i18n';
...
+// Adding/updating a language:
+// 1. Update i18next-parser.config.js to ensure the xy language is in "locales"
+// 2. run: npm run i18n:extract
+// 3. All known/new keys will be added into xy/index.json
+//    - any removed keys will be put into index_old.json, do not commit this file
+// 4. Update translations as needed
+// 5. REMOVE all empty translations. e.g. "key": "", this will allow fallback to the default: English
+// 6. import xyIndex from `./xy/index.json` and add the language to the block at the bottom
+
+import enIndex from './en/index.json';
```
penrods, line 16: See my comment above.  This could be
```
import enIndex from './en/english.json';
import deIndex from './de/german.json';
import frIndex from './fr/french.json';
...
```

I think that help in a few ways:
* Managing files being sent to translators
* Possibly making search results easier to understand in global searches (depends on the editor)

app/locales/mr/index.json
```diff
@@ -0,0 +1,23 @@
+{
```
penrods, line 1: @alpita-masurkar should we just throw away all old translations and start from scratch?  Thoughts?

app/locales/es/index.json
tstirrat: 1. Yep, these are those strings that survived.
2. I too would much prefer the english as the key.

I could keep keys as they are for now in this PR, and do (2) in a follow up if @alpita-masurkar is ok with that.

app/locales/languages.js
tstirrat: The extractor has output path vars for the language code, `en`, but not `english`, so we could do:

`./en.json`, is that enough?

```
// Supports $LOCALE and $NAMESPACE injection
```

app/locales/mr/index.json
kenpugsley: I think it's probably pretty late for @alpita-masurkar ... but I did discuss this in a chat with her earlier today ...

> What do you think about asking translators to start again.  Do you see any issue with that?
> 
> Alpita Masurkar  12:49 PM
> not really,
> 12:49
> this was one of the first things I had discussed with them
> 12:49
> we have some translators who are familiar with github and directly create their prs
> 12:50
> the others had already shown their willingness to help out and are aware that this is an actively developing app, so the content will keep getting updated

tstirrat:
@kenpugsley I've fixed the last two remaining extraction errors and scoured the code base for places where this could have happened. I think we're good:

Screen Shot 2020-04-09 at 4 48 11 PM
Screen Shot 2020-04-09 at 4 48 18 PM

app/locales/languages.js
kenpugsley: Do you mean `./en/en.json`?

tstirrat: either or, both are possible. I'll go with `./en/en.json` for now

tstirrat:
Removed all but en and ht locales

app/locales/en/index.json
tstirrat: now named `en.json`

app/locales/es/index.json
tstirrat: Removed all but ht and en. Will try english keying in a follow up

app/locales/mr/index.json
tstirrat: Thanks, we'll reset for now. Any old strings are here in the history.

PR #422, opened 2020-04-10
Exact location distance calculation for intersect and backfill

kenpugsley:
1. Added exact calculation to find nearby locations with shortcuts to speed determination
2. Use the new exact nearby calculation for the intersection and backfill calculations
3. Update backfill to fill up to 24 hours max, and only when the next position is nearby the last


app/helpers/Intersect.js
```diff
@@ -100,6 +101,54 @@ export async function IntersectSet(concernLocationArray, completion) {
   });
 }
 
+// Function to determine if two location points are "nearby".
+// Uses shortcuts when possible, then the exact calculation.
+// returns boolean
+export function areLocationsNearby(lat1, lon1, lat2, lon2) {
```
tstirrat, line 107: A couple of simple tests for general and edge cases would be super helpful.

app/services/LocationService.js
```diff
@@ -80,35 +80,39 @@ export class LocationData {
...
 
-        // Figure out the time to start the backfill from.  Default is
-        // the time of the last entry in the location set.
-        // To protect ourselves, we won't backfill more than the
-        // maxBackfillTime
-        let lastRecordedTime = lastLocationArray['time'];
-        if ((unixtimeUTC - lastRecordedTime) > this.maxBackfillTime) {
-          lastRecordedTime = unixtimeUTC - this.maxBackfillTime;
-        }
+        var nearby = areLocationsNearby(
```
tstirrat, line 91: Convention seems to be `isX` for booleans

app/services/LocationService.js
```diff
@@ -80,35 +80,39 @@ export class LocationData {
...
-            latitude: lastLocationArray['latitude'],
-            longitude: lastLocationArray['longitude'],
-            time: newTS
-          };
-          console.log('[INFO] backfill location:',lat_lon_time);
-          curated.push(lat_lon_time);
+        // Actually do the backfill if the current point is nearby the previous
+        // point and the time is within the maximum time to backfill.
+        let lastRecordedTime = lastLocationArray['time'];
+        if (nearby && unixtimeUTC - lastRecordedTime < this.maxBackfillTime) {
```
tstirrat, line 102: Code looks good. Tests would be good for this change too.

app/services/LocationService.js
kenpugsley: yeah, I know.  Always bothers me when talking about plurals.  I'll change.

tstirrat: `areX` would be fine too :)

It's the difference of nearby might be nearby points, vs areNearby is clearly a boolean

app/helpers/Intersect.js
kenpugsley: Added testing on the renamed function

app/services/LocationService.js
kenpugsley: Agreed.  Not ideal, but not adding tests for the LocationService in this PR.  Will address in a followup.

app/services/LocationService.js
kenpugsley: it's now `isLocationsNearby`

app/helpers/Intersect.js
```diff
@@ -101,14 +102,62 @@ export async function IntersectSet(concernLocationArray, completion) {
   });
 }
 
+// Function to determine if two location points are "nearby".
```
tstirrat, line 105: use jsdoc:

```
/**
 * 
 */
```

That way many IDEs will show the method description on hover of the method anywhere in the code base. (IDEA and VScode)

app/services/LocationService.js
```diff
@@ -79,35 +79,39 @@ export class LocationData {
...
-            latitude: lastLocationArray['latitude'],
-            longitude: lastLocationArray['longitude'],
-            time: newTS,
-          };
-          console.log('[INFO] backfill location:', lat_lon_time);
-          curated.push(lat_lon_time);
+        // Actually do the backfill if the current point is nearby the previous
+        // point and the time is within the maximum time to backfill.
+        let lastRecordedTime = lastLocationArray['time'];
+        if (nearby && unixtimeUTC - lastRecordedTime < this.maxBackfillTime) {
```
tstirrat, line 101: Hah, sorry I wasnt clear this was the `isX` I was talking about.

areLocationsNearby was fine

PR #437, opened 2020-04-11
Add a debug mode for intersections

### LINKED ISSUE #423: FEATURE REQ: Test/demonstrate intersection with infected trails
While testing and demonstrating it would be useful to be able to see the interface that is shown when you cross paths with an infected individual.  The following is a decent scheme for achieving this:

* Add a handler for clicking on the Team icon in the About box.  Have the page-open reset a state variable to 0, then increment it once each tap.  If they get to 3(?) taps, toggle the test mode described here.  Perhaps change the displayed icon?
* In "test" mode, run code similar to the [DEBUGGING CODE](https://github.com/tripleblindmarket/covid-safe-paths/blob/develop/app/views/Notification.js#L90) in Notification.js
* Have the code save fake counts to the CROSSED_PATHS StoreData.  This information gets regenerated once every 12 h ours.  It should also set a TEST_DATA boolean to true
* On the main page, if TEST_DATA is set you should not run the intersection that occurs on page load.
* On the Exposure History page you should see the results
* Going back to About and triple-tapping can immediately reset the flag and run another intersection to fill the CROSSED_PATHS array correctly.

penrods:
Added the ability to enter a "debug" mode where an artificial intersection is
created.  This happens by:
* Going to the About screen
* Tapping on the Team icon 4 times
* You can also return and tap 4 more times to toggle.

When in this debug mode, faux intersection data is placed in the crossing
arrays.  This will either clear automatically during the 12-hour calculation,
or can be cleared by the tester by returning to the About box and repeating
the 4-tap.

 ### Testing Notes ###
Go to About, tap the icon 4 times.  Close and restart the app to see the 
faux positive results.

 ### Fixed Issues ###
The resolves #423, providing a debug mode.


app/views/About.js
```diff
@@ -23,18 +25,60 @@ class AboutScreen extends Component {
...
+  handleTapTeam = () => {
+    this.setState({ tapCount: this.state.tapCount + 1 });
+    if (this.state.tapCount >= 3) {
+      if (this.state.tapCount == 3) {
+        // Debug mode on
+        SetStoreData(DEBUG_MODE, 'true');
+
+        // Create faux intersection data
+        let pseudoBin = [];
+        for (let i = 0; i < 28; i++) {
```
tstirrat, line 56: magic numbers, use a const.

app/views/About.js
```diff
@@ -23,18 +25,60 @@ class AboutScreen extends Component {
...
+          pseudoBin.push(intersections);
+        }
+        let dayBin = JSON.stringify(pseudoBin);
+        SetStoreData(CROSSED_PATHS, dayBin);
+      } else if (this.state.tapCount == 7) {
+        // Debug mode off
+
+        // Create faux intersection data
+        let pseudoBin = [];
+        for (let i = 0; i < 28; i++) {
```
tstirrat, line 70: Same here re magic numbers.

app/views/About.js
```diff
@@ -23,18 +25,60 @@ class AboutScreen extends Component {
...
+          const intersections = Math.max(
+            0,
+            Math.floor(Math.random() * 50 - 20),
+          );
+          pseudoBin.push(intersections);
+        }
+        let dayBin = JSON.stringify(pseudoBin);
+        SetStoreData(CROSSED_PATHS, dayBin);
```
tstirrat, line 64: Does this overwrite real data? perhaps we can store fake data in `CROSSED_PATHS_DEMO`

`GetStoreData(CROSSED_PATHS)` can check for `DEMO_MODE` and return `CROSSED_PATHS_DEMO` if true.. this would allow demo mode to work in all locations.

app/views/About.js
```diff
@@ -53,7 +97,19 @@ class AboutScreen extends Component {
             {languages.t('label.commitment_para')}
           </Text>
 
-          <SvgXml style={styles.aboutSectionIconTeam} xml={team} />
+          <SvgXml
+            onPress={this.handleTapTeam.bind(this)}
```
tstirrat, line 101: it's already this bound because it used `handleTapTeam = () => {}` you can remove .bind(this)

app/views/About.js
```diff
@@ -23,18 +25,60 @@ class AboutScreen extends Component {
...
   componentWillUnmount() {
     BackHandler.removeEventListener('hardwareBackPress', this.handleBackPress);
   }
 
+  handleTapTeam = () => {
+    this.setState({ tapCount: this.state.tapCount + 1 });
+    if (this.state.tapCount >= 3) {
+      if (this.state.tapCount == 3) {
+        // Debug mode on
+        SetStoreData(DEBUG_MODE, 'true');
```
tstirrat, line 52: pop up a toast notification (if we have that?) saying demo mode has been enabled.

app/views/About.js
```diff
@@ -23,18 +25,60 @@ class AboutScreen extends Component {
...
+        for (let i = 0; i < 28; i++) {
+          const intersections = Math.max(
+            0,
+            Math.floor(Math.random() * 50 - 20),
+          );
+          pseudoBin.push(intersections);
+        }
+        let dayBin = JSON.stringify(pseudoBin);
+        SetStoreData(CROSSED_PATHS, dayBin);
+      } else if (this.state.tapCount == 7) {
```
tstirrat, line 65: Suggestion... instead of counting more taps to disable:

Add a label next to the Team icon or title:

`Demo mode is enabed, tap to disable`

Now its very clear if you're in or out of demo mode.

tstirrat:
I used this yesterday, it was very helpful!

It requires the app to be closed/ reopened to see the change on the main screen. It's still extremely useful.

app/views/About.js
penrods: I based this on the code above for this.backToMain.  But dropping to just
```javascript
onPress={this.handleTapTeam}
```
Seems to work.

app/views/About.js
penrods: Modified to restore data on exit.

app/views/About.js
penrods: Done

app/views/About.js
penrods: Done

app/views/About.js
penrods: Don't know about this.  Sounds pretty fancy for a debugging thingy...  :)

app/views/About.js
penrods: I added a label.  It is English only, but I think that is reasonable for a hidden debug/demo feature.

kenpugsley:
@penrods are we going to try and get this in before v1?  Looks like it needs conflicts resolved again.

tstirrat:
> @penrods are we going to try and get this in before v1? Looks like it needs conflicts resolved again.

I vote yes! I can take over and get it merged if Steve is busy

tstirrat:
Made the label red  and the text beside it tappable:

Screen Shot 2020-04-18 at 1 11 19 PM

tstirrat:
Made the text more clear for tapping:

Screen Shot 2020-04-18 at 3 46 06 PM

Issue #439, opened 2020-04-11
Cocoapods gem should be provisioned in the development setup script if users do not have it set up


efalkner:
Cocoapods should be provisioned in the development setup scripts if they are not.
Suggest including ruby `bundler` package to properly manage versions of ruby in the environment (and avoid inconsistencies in cocoapod installs)

PR #440, opened 2020-04-11
Update Cocoapods Provisioning

efalkner:
Addresses #439 

- Globally install `bundler` package management for ruby in the dev_setup.sh script (do not add this to the dev_setup.bat script, as it's required for iOS development only).
- Add a Gemfile to specify a cocoapods gem version (use the same version as what was in the develop branch Podfile.lock)
- Specify bundler version in Gemfile
- Add Gemfile.lock
- Update package.json command to use `bundle exec pod install`, which specifies using the gems in the local vendor/bundle cache, rather than a global install.


dev_setup.sh
```diff
@@ -184,6 +184,12 @@ if ! found_exe android-studio ; then
     fi
 fi
 
+if ! gem list '^bundler$' -i --version 2.1.4; then
```
tstirrat, line 187: Will this work on ubuntu too?

ios/Gemfile.lock
```diff
@@ -0,0 +1,94 @@
+GEM
```
tstirrat, line 1: Add a github actions cache for bundler to `.github/workflows/build.yml`

https://github.com/actions/cache/blob/master/examples.md#ruby---bundler

dev_setup.sh
tstirrat: Looks like it might need an executable check on `gem`, and `sudo apt-get install rubygems` if not found

efalkner: For now, we don't need ruby installed on anything but a macOS. I don't think people can develop on ubuntu for apple (since xcode tooling requires macOS to be installed). I can definitely throw a check around this so it doesn't error out for users on ubuntu though.

efalkner: Added an OS check for the install

ios/Gemfile.lock
efalkner: Done ✅

PR #465, opened 2020-04-11
Final removal of Private Kit (leaving only attributions)

kenpugsley:
Final removal of Private Kit from the repo, except for the attributions.  Should set us on the path to be able to produce final app store release builds.

I've also included the rewording on the location/accelerometer permission description (to take out the refugee wording).

What I've tested (on Mac ... not Linux - sorry @penrods 

- pulled branch fresh the repo
- yarn install, yarn install:pod
- running on both Android and iOS simulators
- building and installing on iPhone from xcode
- building Android .apk from command line and sideloading on Android phone

All seems to work.


android/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,13 @@
+arguments=
```
tstirrat, line 1: .gitignore these elipse specific file additions

android/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tstirrat, line 1: .gitignore eclipse stuff

android/.settings/org.eclipse.buildship.core.prefs
kenpugsley: Hmmm... not sure where this came from.  Will remove.

android/.project
kenpugsley: same as above ... not sure where this came from

PR #480, opened 2020-04-12
Auto sort import statements with Prettier

tstirrat:
Used this in another project and its very nice.

Sorts imports by:

1. side effect imports like `import './a';` 
1. npm packages ascending on package name `import x from '@emotion';`
1. same folder, ascending on file name `import x from './dep';`
1. different folder relative, ordered by number of `../` and then asc: `import x from '../../common/x'`


penrods:
I'm fine with "opinionated" formatters.  Black drove me crazy for a while in the Python world, but I think overall it is worth it.

tstirrat:
Yeah, I like the fact that it makes duplicate imports easy to see, and parent/child file relationships easier to see, in one place.

PR #481, opened 2020-04-12
Automated dev_setup.sh testing

tstirrat:
Runs a couple of tests on after dev_setup.sh to ensure on mac and linux that all the required bits are there (like brew, yarn, watchman, env vars)


__tests__/dev_setup.test.bats
```diff
@@ -0,0 +1,89 @@
...
+}
+
+@test "android-studio is installed" {
+  linux_only
+
+  run found_exe android-studio
+  [ "$status" -eq 0 ]
+}
+
+@test "~/.profile_mobileapp exists" {
```
kenpugsley, line 44: Not sure this is actually needed.  Technically this only happens if the user selected to have the dev_setup script install android studio and set up the environment vars.  Probably OK to not check this I think.

tstirrat: In non-interactive mode `-ni` the file is always made, so I think a test for it is warranted, unless we decide to conditionally write the file.

tstirrat:
For now, this tests `dev_setup.sh` _as it is today_, and we can improve on the script and github action in future PRs.

Issue #496, opened 2020-04-13
Android StatusBar styles not applied when app opened from notification.


troach-salesforce:
## Version

- Type of phone: Android
- OS Version: Multiple
- Model: Multiple

## Steps to reproduce

When clicking on the a Safe Paths notification in the drawer, the app opens to the LocationTracking.js view. However, the StatusBar style (light-content, transparent, translucent) set inside render does not get applied to the status bar. This. seems to be a known React Native bug, which can be worked around by using the static StatusBar methods.

## Expected Behavior

When I click on a notification in the drawer, I expect the status bar in LocationTracking.js screen to be styled with light content, transparent, and translucent.

## Upload any files to the Issue useful in helping us to investigate

The screenshot is difficult to see because it runs into the white background, but the status bar is white from notification click. 

![Screenshot_20200413-124326](https://user-images.githubusercontent.com/63416413/79140708-644bcb80-7d86-11ea-803f-f9fa0be9ebcd.png)

troach-salesforce:
Work has been merged, closing.

PR #497, opened 2020-04-13
Android StatusBar Style Fix

troach-salesforce:
## Description

Fixes issue #496 where status bar styles do not get set when the app is opened from an android notification.

## How to test

Click to open the app from the Safe Path notification in the drawer.


app/constants/colors.js
```diff
@@ -9,6 +9,7 @@ const colors = {
   PRIMARY_TEXT: '#000',
   GREEN: '#32A852',
   VIOLET: '#4051DB',
+  TRANSPARENT: '#00000000',
```
tstirrat, line 12: does `transparent` work here?

troach-salesforce: @tstirrat, changed to 'transparent' and resolved conflicts

PR #499, opened 2020-04-13
Remove hardcoded link to Haitian health authority

penrods:
The News page had a hardcoded "Haitian Ministry of Health" entry.  That has
been removed.  Also cleaned up a few other things:
* Removed some console.log() messages
* Added several variables to the storage.js list (if we are going to do that,
  we should do it everywhere).
* Fixed the AUTHORITY_NEWS population (it was always serializing an empty list)
* Add/Remove of a Healthcare Authority now kicks off an intersect calc
  immediately.  This also updates the links to News, etc (but it can take a
  few seconds to complete in the background).

 ### Testing Notes ###
Add/Remove authorities.  You can see the results visibly in the News page.
NOTE: Jumping in/out of News too quickly might not give it time to repopulate
the list, so only the default page shows.  This is expected.


app/helpers/Intersect.js
```diff
@@ -248,13 +248,24 @@ export function checkIntersect() {
               //      { "time": 456, "latitude": 12.34, "longitude": 12.34}
               //    ]
               // }
-
-              // Update cache of info about the authority
+              // TODO: Add an "info_exposure_url" to allow recommendations for
+              //       the health authority driectly on the Exposre History
```
tstirrat, line 252: typo: Exposure

PR #504, opened 2020-04-14
[RC] fix function call

tremblerz:
## Description

Fix the function call

## How to test

Run on xcode and press simulate background fetch and you will see background task getting called in terminal


tstirrat:
Same @kenpugsley for merging

PR #507, opened 2020-04-14
Fix usage of URL from a Healthcare Authority

penrods:
The webview was using the wrong name for the webview ("url" instead of
"news_url").

 ### Testing Notes ###
The "Example Authority" was not showing a webpage.  It does now.


app/views/News.js
```diff
@@ -28,7 +28,7 @@ class NewsScreen extends Component {
     super(props);
     let default_news = {
       name: languages.t('label.default_news_site_name'),
-      url: languages.t('label.default_news_site_url'),
+      news_url: languages.t('label.default_news_site_url'),
```
tstirrat, line 31: *mumbles something about Typescript would have caught this*

PR #508, opened 2020-04-14
[RC] Notification fix

tremblerz:
## Description

Fixes the issue of overlap being found but screen not being updated accordingly. Gets solved by using react's navigation focus listener.

## How to test

Fire up intersection logic to update the state to intersection found (by adding healthcare authorities) and the screen will turn black.

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -111,7 +111,7 @@ class LocationTracking extends Component {
         switch (result) {
           case RESULTS.GRANTED:
             LocationServices.start();
-            this.checkIfUserAtRisk();
+            this.spawnIntersect();
             return;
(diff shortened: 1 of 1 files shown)


app/views/LocationTracking.js
```diff
@@ -125,11 +125,13 @@ class LocationTracking extends Component {
       });
   }
 
-  checkIfUserAtRisk() {
+  spawnIntersect() {
```
tremblerz, line 128: Splitted the function into two parts, `spawnIntersect()` is the heavyweight intersection algo call while `checkIfUserAtRisk()` only reads from asyncstorage and changes the results based on the state.

tstirrat:
Will leave for @kenpugsley to merge into RC

PR #467, opened 2020-04-12
Event history UX update

tstirrat:
_Sorry for the size of this PR, doing the Theming blew it out. I am willing to split theme stuff into a separate PR if needed._

- Replaces moment js with dayjs (immutable = way less potential for mutation/state bugs)
- Added emotion's styled components, for cleaner styles, and ability to create a powerful theming system
- Developed a set of themes and a `<Theme>` component with these known themes
  - main (settings screen)
  - dark (when exposure is found, white on gray)
  - inverted (main screen, white on violet)

- Added a Typography component that encodes the design system from Upstatement, and is theme color aware
  - Currently this is not migrated to be used outside this PR, but I do plan to migrate theming and update other components (like Button) after this PR
- Tests for Typography, Theme,  and places with date math

Screenshots:

Screen Shot 2020-04-13 at 2 22 17 PM

Screen Shot 2020-04-13 at 2 22 41 PM

Fixes #495

PR #510, opened 2020-04-14
Fix untokenized strings

### LINKED ISSUE #495: Missing Key in the en.json file?
This was brought up by @diarmidmackenzie in an issue in another case https://github.com/tripleblindmarket/covid-safe-paths/issues/483

On the Exposure History page, there is some English text that goes like this:
What does this mean? Individuals who don't exhibit symptoms can sometimes still carry the infection and infect others. Being careful about...

This text is not getting translated as it is not included in the en.json file. Is there a missing key for this?

denispapakul:
## Description

Added untokenized strings in Notifications.js for all langs. Resolves #495


tstirrat:
Thanks for the PR, however the wording on these strings will be changed in #467, and I did the i18n extraction there too .. I'll close this for now.

PR #518, opened 2020-04-15
[RC] final wording and flow changes, also fix edge case on removing authorities

kenpugsley:
## Description

Final wording for the RC.  Also fixes an edge case where removing all authorities would not clear the "There is a Data Match" state.  

## How to test

Run the app, check the wording


tremblerz:
@kenpugsley just update the snapshots with new language change

kenpugsley:
fixed unit test.  will merge after tests pass

Issue #523, opened 2020-04-15
Android doesn't resume location tracking on restart


troach-salesforce:
## Version

- Type of phone: Android
- OS Version: 10 (Assume most devices)
- Model: Pixel 3a (Assume most devices)

## Steps to reproduce

Restart the device, location tracking does not resume until the app is manually opened.

## Expected Behavior

When I restart the device, location tracking should resume in the background without requiring the user to open Safe Paths

## Research

This appears to be a bug in https://github.com/mauron85/react-native-background-geolocation.
See: 
https://github.com/mauron85/react-native-background-geolocation/issues/450
https://github.com/mauron85/react-native-background-geolocation/issues/345

It doesn't appear that this library has updated since Sept 2019 and isn't under much active development to fix bugs. I can work on the solution, but we should agree on a path forward. Submit a PR, and in the meantime fork the library?

troach-salesforce:
I'll take this task if an admin wants to add me as the assignee.

troach-salesforce:
Fix merged. Closing

PR #526, opened 2020-04-15
Patch for android restart bug in react-native-background-geolocation

troach-salesforce:
## Description

react-native-background-geolocation bug fix that described in #523 

## How to test

Restart your device. You should see the SafePaths notification in drawer within a minute or two, indicating that react-native-background-geolocation is pulling location again.

Fixed the issue using patch-package. The patch show many lines changed due to jettifier, but the patch itself was a single line (L248)


tstirrat:
Forgive me, perhaps I need it explained. I don't quite understand the manual patching. Isn't that what _jetifier already does_?

troach-salesforce:
@tstirrat see L248 with the description of the bug: https://github.com/tripleblindmarket/covid-safe-paths/issues/523

There was a separate need for matching, not related to jettifier. Running the patch command also made jettifier changes.

tstirrat:
ok, interesting.. Is there a way to patch before jetifier so that your patch diff is minimal?

The problem is, it's unclear what you're actually patching.

troach-salesforce:
Sure, I'll see if I can apply the patch before jettifier is run and see if jettifier is still able to patch at a later time.

tstirrat:
The other Q is, is there something preventing this from being updated upstream?

Have you tried proposing the fix to the original repo?

Here's another suggestion: fork the repo and create the fix on a branch, use that branch in package.json (we've done this for others)..

Then, we can re-integrate once the fix is upstream.

troach-salesforce:
I will submit a PR there as well, but this bug and solution have been reported for 6+ months now. Unfortunately, the project doesn't look very active. Patching seemed to be the fastest solution, and a decent one for a library that isn't undergoing many updates. If we go with a fork solution, we should do so on a repo owned by SafePaths. I didn't want to fork to my personal.

tstirrat:
> If we go with a fork solution, we should do so on a repo owned by SafePaths. I didn't want to fork to my personal.

Yes I agree there

@penrods if you agree too, can you fork https://github.com/mauron85/react-native-background-geolocation into this org?

Personally, I'd rather a fork and have an obvious pending PR back to the original repo, rather than have a patch file here and no way to easily tell if it's integrated upstream

troach-salesforce:
@tstirrat @penrods We need multiple forks. The core of the issue is in the submodule that uses found here: https://github.com/mauron85/background-geolocation-android

tremblerz:
The original repo is indeed outdated, probably fork would be a better option. This would be useful in future also if we want to apply changes on top of this plugin given that it is a central part of what we're doing

troach-salesforce:
If you aren't opposed. I have a much cleaner patch going if we can temporarily use this and create a ticket to create the fork. I'd like to focus working on part 2 which is another separate issue preventing location working on background restarts. https://github.com/tripleblindmarket/covid-safe-paths/issues/527

troach-salesforce:
@tstirrat I'm pushed a commit with a minimal patch. If we are opposed to merging in a patch, I'll coordinate with @tremblerz and others to get changes pushed out to a fork. I'd like to merge this in as a temporary fix if at all possible because of #527, which I've also begun working on. We can create another ticket for me to prep the fork after 527 is fixed. It's the end of the day for me here today so I'll jump back on tomorrow and adjust to whatever is decided.

tremblerz:
works on restart.

ehsanaskari74:
> For our tracking purposes, the solution is here: [tripleblindmarket#526](https://github.com/tripleblindmarket/covid-safe-paths/pull/526)

Dear @troach-sf it's not enough because it doesn't start Headless Task in Android and we have this issue > https://github.com/mauron85/react-native-background-geolocation/issues/473#issuecomment-620850714

troach-sf:
@ehsanaskari74, thank you Ehsan. I am aware that there is a part 2 to this fix. This first fix was necessary to prevent a silent crash in the library, and now a second step is necessary to improve tracking on our end since the listeners no longer exist on reboot. We have a couple of upcoming requirements so the fix to this will all be contained in 1 PR.

1. We are going to start storing user locations in an encrypted database.
2. We want to insert locations in Android and iOS without requiring the app to come to the foreground.
3. We need to prevent the geolocation library from also storing locations in its own db. 

The work for all of this is probably 2-3 days out, and can be tracked here:
https://github.com/safe-paths-contrib/covid-safe-paths/tree/feature/realm

lyseiha:
anything update on it?

PR #529, opened 2020-04-16
Fix intersect to handle multiple authorities

kenpugsley:
## Description

Fixes the intersection logic to better handle multiple authorities. Turns down the async->async stuff a bit to help ensure we get correct results.

This is the backfit to `develop` for #517

## How to test

Put in more than 1 health authority that would have intersecting data, and verify the bin counts go up as expected (that is still in the logging)


kenpugsley:
ready for review

app/helpers/Intersect.js
```diff
@@ -210,86 +179,150 @@ function binarySearchForTime(array, targetTime) {
...
+  // Last checked key is not being used atm. TODO check this to update periodically instead of every foreground activity
+  SetStoreData(LAST_CHECKED, unixtimeUTC);
+
+  // looks like we did something... return true to signal OK
+  return true;
+}
+
+// simple function to get an empty dayBin array when needed.
+function getEmptyDayBin() {
+  let dayBin = [
```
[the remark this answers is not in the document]
kenpugsley: Yeah... good point.  Not sure exactly why we do a 28 day match.  Will change.

app/helpers/Intersect.js
```diff
@@ -210,86 +179,150 @@ function binarySearchForTime(array, targetTime) {
...
+    }
+  }
+  console.log('Crossing results: ', dayBin);
+
+  // store the results
+  SetStoreData(CROSSED_PATHS, dayBin); // TODO: Store per authority?
+
+  // the current time is the last time checked
+  let nowUTC = new Date().toISOString();
+  let unixtimeUTC = Date.parse(nowUTC);
```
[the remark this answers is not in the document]
kenpugsley: Yeah, much cleaner.

app/helpers/Intersect.js
```diff
@@ -210,86 +179,150 @@ function binarySearchForTime(array, targetTime) {
...
+  // Last checked key is not being used atm. TODO check this to update periodically instead of every foreground activity
+  SetStoreData(LAST_CHECKED, unixtimeUTC);
+
+  // looks like we did something... return true to signal OK
+  return true;
```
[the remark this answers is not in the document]
kenpugsley: Yeah... my original intent here was to have it return false if something fails.  Not currently doing that, so I'll remove.

app/helpers/Intersect.js
```diff
@@ -8,100 +8,64 @@ import PushNotification from 'react-native-push-notification';
...
-        }
-
-        i++;
+/**
+ * Intersects the locationArray with the concernLocationArray, putting the results
+ *   into the appropriate bins in dayBin.
+ *
+ * @param {array} locationArray - array of the local locations
+ * @param {array} concernLocationArray - superset array of all concerning points from health authorities
+ * @param {array} dayBin - bins to be populated with the intersection results
```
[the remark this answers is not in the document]
kenpugsley: I'll probably leave this semantically as is, but will be refactoring some (previous comments).  The long term idea is to be able to break up the intersection into multiple smaller parts that we do in order, so having a generic way to intersect into a partially intersected set of bins makes sense in that case.

app/helpers/Intersect.js
```diff
@@ -210,86 +179,150 @@ function binarySearchForTime(array, targetTime) {
...
+  // Set the news for the authorities found.  There's an implicit assumption that news only
+  //  comes from the configured authorities
+  SetStoreData(AUTHORITY_NEWS, name_news);
+
+  // now, get the location data, normalize and do the intersection
+  let locationArrayString = await GetStoreData(LOCATION_DATA);
+  if (locationArrayString !== null) {
+    let locationArray = JSON.parse(locationArrayString);
+    let concernArray = concernResultsCombined;
+    dayBin = intersectSetIntoBins(locationArray, concernArray, dayBin);
```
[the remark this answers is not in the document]
kenpugsley: Will be more obvious after come code cleanup.

app/helpers/Intersect.js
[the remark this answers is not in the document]
tstirrat: JS dates don't maintain the timezoneOffset.. meaning.. if you set a date at now (UTC) it will still have your local getTimezoneOffset

We definitely need tests on that, and I suspect dayjs().utc() might be required to keep things sane, code wise.

app/helpers/Intersect.js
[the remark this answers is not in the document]
tstirrat: ok, so it's iterative/progressive. 

In that case, **definitely clone** the input array, _never mutate_ it, and return a new array each time..  or we'll have a bad time later when its harder to get a fix out quickly.

app/helpers/Intersect.js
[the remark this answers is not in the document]
tstirrat: relevant link that @efalkner sent to me: https://www.youtube.com/watch?v=EUzPpXhFmG8

Takeaway:

> time math is more complicated than we think, use a lib.

kenpugsley:
@tstirrat @penrods  ... A fair number of changes here for such a small number of files.  Should have the edge cases covered now along with testing.  Ready for you guys to take a look again.

Also ... including small updates to ExposureHistory and the debug mode to account for the change to the bins having the exposure time millis.

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -1,19 +1,569 @@
...
+      emptyLocations,
+      emptyConcernLocations,
+    );
+    expect(resultBins.every(a => a == -1)).toBe(true);
+  });
+
+  /**
+   * Exact matches in the baseLocations and the concernLocations
+   */
+  it('intersect of known location intersections at default intervals has known result', () => {
```
tstirrat, line 129: extract the grouping into a describe block:

```js
describe('intersect of known location', () => {
  it('intersections at default intervals has known result', ...)
});
```

app/views/ExposureHistory/__tests__/ExposureHistory.spec.js
```diff
@@ -23,7 +23,9 @@ describe('convertToDailyMinutesExposed', () => {
   });
 
   it('converts day bins to minutes', () => {
-    expect(convertToDailyMinutesExposed('[0, 1, 2, 3, 0]')).toEqual([
+    expect(
```
tstirrat, line 26: check non even spaced minutes to see what we get

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -1,19 +1,569 @@
...
+    // 5 locations, spread over 17 days.  Note, the final location is over the 14 days that intersect covers
+    let baseLocations = [
+      ...generateBackfillLocationArray(
+        TEST_LOCATIONS.kansascity.base,
+        TEST_MOMENT_MS,
+      ),
+      ...generateBackfillLocationArray(
+        TEST_LOCATIONS.hammerfest.base,
+        //TEST_MOMENT_MS - 3 * 24 * 60 * 60 * 1000,
+        TEST_MOMENT_MS - dayjs.duration(3, 'days').asMilliseconds(),
```
tstirrat, line 139: `TEST_MOMENT.subtract(3, 'days')` is clearer.

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -1,19 +1,569 @@
...
+    ];
+
+    // normalize and sort
+    baseLocations = normalizeAndSortLocations(baseLocations);
+    concernLocations = normalizeAndSortLocations(concernLocations);
+
+    let resultBins = intersectSetIntoBins(baseLocations, concernLocations);
+
+    let expectedBins = getEmptyLocationBins();
+    expectedBins[0] = dayjs
```
tstirrat, line 185: in a lot of these tests you test that the "hits" are correct. but you don't test that the spaces in between have 0 or -1

You have a general "empty" test, which is great. but also you want to check when doing the math on real points, that it also calculates 0, -1 in the right gaps either side

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -1,19 +1,569 @@
...
+  });
+
+  afterEach(() => {
+    MockDate.reset();
+  });
+
+  /**
+   * Simplest case, empty data
+   */
+  it('intersect of empty locations produces bins with no intersections', () => {
```
tstirrat, line 116: Suggest a little more testing on the possible edge cases around the 0 daybin (today).. like end of day, start of day, middle of day.

tstirrat:
Actually a timezone test - one that specifically sets today to a date where it would be yesterday/tomorrow in UTC. We should know what's happening in that case. because the bin alignment is not the same as the user's day alignment.. and perhaps this is a UX issue too.. but we should know the edges and have tests on it

app/helpers/__tests__/Intersect.spec.js
kenpugsley: The 0's and -1 are in fact being tested... all tests are using `expect(resultBins).toEqual(expectedBins);` to do a deep compare.  That won't pass unless *all* of the bins match the expected.

app/helpers/__tests__/Intersect.spec.js
kenpugsley: Agree... will change.

app/helpers/__tests__/Intersect.spec.js
kenpugsley: Yeah, a test that crosses midnight does make sense.  I'll add.

kenpugsley:
> Actually a timezone test would be great - one that specifically sets today to a date where it would be yesterday/tomorrow in UTC. We should know what's happening in that case. because the bin alignment is not the same as the user's day alignment.. and perhaps this is a UX issue too.. but we should know the edges and have tests on it

I do think this is mostly on the UX side.  The bins and intersection is all based on the unix timestamp -  so timezone comes up only indirectly around figuring the determination of midnight for the switch to the next day bin.  I'll add in some testing for that, but I think we should not try to pull in the UX testing into this work.

app/views/ExposureHistory/__tests__/ExposureHistory.spec.js
kenpugsley: I'm going to leave this alone for now ... a little beyond the scope of what this is trying to fix.

app/helpers/__tests__/Intersect.spec.js
tstirrat: Ah, totally missed that! Great. best of both worlds highlighting specific points, then testing everything in between 👍

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -513,6 +699,92 @@ describe('intersectSetIntoBins', () => {
...
+/**
+ * These are tests running at interesting times, such as near midnight or at a 
+ *   daylight savings change.  
+ * 
+ * TODO: daylight savings not yet tested
+ */
+describe('instersect at interesting times', () => {
+  dayjs.extend(duration);
+
+  afterEach(() => {
```
tstirrat, line 711: no before each, seems fishy?

tstirrat:
Thanks for doing the legwork on all the tests! It's not immediately rewarding work, but it drastically reduces stress /risk merging any prs in this code moving forward.

I'll think about the "pre intersect tests" era every time I see green unit tests :)

kenpugsley:
> Thanks for doing the legwork on all the tests! It's not immediately rewarding work, but it drastically reduces stress /risk merging any prs in this code moving forward.
> 
> I'll think about the "pre intersect tests" era every time I see green unit tests :)

Honestly, me too.

PR #538, opened 2020-04-16
Don't translate the default news URL

tstirrat:
## Description

Stop the confusion each time on new translation files about this url and that the key should not be included in .json files.

## How to test

```
yarn run i18n:extract
```

This should not put the key in a new or existing json file.

### DIFF
--- a/app/constants/authorities.js
+++ b/app/constants/authorities.js
@@ -3,5 +3,3 @@ export const PUBLIC_DATA_URL =
 
 export const AUTHORITIES_LIST_URL =
   'https://raw.githubusercontent.com/tripleblindmarket/safe-places/develop/healthcare-authorities.yaml';
-
-export const DEFAULT_NEWS_SITE_URL = 'https://covidsafepaths.org/in-app-news';
--- a/app/constants/authorities.js
+++ b/app/constants/authorities.js
@@ -3,3 +3,5 @@ export const PUBLIC_DATA_URL =
 
 export const AUTHORITIES_LIST_URL =
   'https://raw.githubusercontent.com/tripleblindmarket/safe-places/develop/healthcare-authorities.yaml';
+
+export const DEFAULT_NEWS_SITE_URL = 'https://covidsafepaths.org/in-app-news';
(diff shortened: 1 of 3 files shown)


app/constants/authorities.js
```diff
@@ -3,3 +3,5 @@ export const PUBLIC_DATA_URL =
 
 export const AUTHORITIES_LIST_URL =
   'https://raw.githubusercontent.com/tripleblindmarket/safe-places/develop/healthcare-authorities.yaml';
+
+export const DEFAULT_NEWS_SITE_URL = 'https://covidsafepaths.org/in-app-news';
```
alpita-masurkar, line 7: since it now contains more than authorities url now, maybe refactor this class name to reflect what it contains...

tstirrat: hrm. yeah. good point.

I think like in the other one I will leave this in the only file that uses it.. until it's used somewhere else.

penrods:
Actually this was intentional.   I suspected that we might have a translated landing-page for all of the major languages.  @alpita-masurkar , thoughts?  Are there better ways to do this?

kenpugsley:
We wouldn't want to try and maintain multiple language default news urls, would we?  Seems fairly hard to maintain that way.  Ideally the embedded browser would be sending the locale of the phone and on the website we'd detect and change (if we even intend to do so).

alpita-masurkar:
> Actually this was intentional. I suspected that we might have a translated landing-page for all of the major languages. @alpita-masurkar , thoughts? Are there better ways to do this?

@penrods I think that's a valid thought.
Which countries/ regions are a significant proportion of our downloads coming from? I think we've already taken that into account wrt Haiti or Vietnam for example..
We shouldn't have a blanket implementation for all languages, but if there's a demand from a certain region, then we should have a mechanism for doing a bit extra on localization wrt:

1. some default news/ authorities URLs (ask the translator to include them on some keys?)

2. Localization, not literal translation. Some sentences might be longer or shorter in English compared to other languages. The hard placement of \n in the sentence may work in some cases and may lead to an overflow or look strange in another case. If it's a possibility, for high priority languages, maybe a pre-release copy can be given to the translator to see if it looks right on the screen. (Let the translator see for themselves and catch some bug before it's released).

3. If we are planning to change content often in the next few weeks, one of the problems that will happen is some content might appear in English and some in the local language (as we have seen before). For the high-demand languages, localization should not look like an afterthought.
If it's possible, we should probably explore solutions like locize or Crowdin that @tstirrat suggested on the slack language support channel before. There's a free trial, but it's not free.
This will help keep the translation process continuous.

PR #539, opened 2020-04-16
No translation for Mayo clinic link

tstirrat:
This link so far has not changed for any translation.

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -49,7 +49,7 @@ const StateEnum = {
   NO_CONTACT: 2,
 };
 
-const StateIcon = ({ title, status, size, ...props }) => {
+const StateIcon = ({ status, size }) => {
   let icon;
   switch (status) {
     case StateEnum.UNKNOWN:
@@ -67,7 +67,6 @@ const StateIcon = ({ title, status, size, ...props }) => {
   );
 };
 
-const width = Dimensions.get('window').width;
 const height = Dimensions.get('window').height;
 
 class LocationTracking extends Component {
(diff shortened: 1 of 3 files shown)


app/views/LocationTracking.js
```diff
@@ -38,17 +38,18 @@ import { PARTICIPATE } from '../constants/storage';
 import { GetStoreData, SetStoreData } from '../helpers/General';
 import { checkIntersect } from '../helpers/Intersect';
 import languages from '../locales/languages';
-//import BroadcastingServices from '../services/BroadcastingService';
 import BackgroundTaskServices from '../services/BackgroundTaskService';
 import LocationServices from '../services/LocationService';
 
+const MAYO_COVID_URL = 'https://www.mayoclinic.org/coronavirus-covid-19';
```
alpita-masurkar, line 44: a constant's perfect for this case.
just one question... is there a likelihood of this constant MAYO_COVID_URL being used outside of locationTracking? 

If yes, then it might not be intuitive to lookup this URL in locationTracking.
I have seen a small URL constants class in our repo . Although I'm not a big fan of having one URL constants class, if there's a larger use for this constant and if it serves the purpose, then it can be moved there.
I think it is called authorities.js (but should be refactored at some point if it grows because it doesn't contain just links to authorities).

tstirrat: I like your thinking. If someone else needs it, we can move it outside then, but for now, only this page uses it. What do you think?

alpita-masurkar: okay. that sounds perfect. :)

PR #549, opened 2020-04-17
Add location tracking toggle and language picker to Settings screen

StartupSam:
## Description

Per the Figma design, I added two new settings: a location on/off toggle, and a language picker.

The app wasn't equipped for displaying the state of a user who turned off tracking, so I had to make some additions to the LocationTracking.js screen as well to both show correct information and to update when the user changed the setting (without necessitating an app restart). For similar reasons, made a slight modification to Entry and Onboarding1 so returning users wouldn't be subjected to onboarding on every open.

The other changed files are new icons and translation strings.

## How to test

We need to make sure that the PR 
- Works as expected on high-resolution screens (it should, but would love to verify)
- Language picker works the same as the onboarding one on iOS

## Still to do
On startup, there's a small bug where the LocationTracking screen will show the status unknown icon/text for a brief moment before showing the correct status.

Also, I may need some help merging back into develop - didn't make friends with git the last time I tried. (UPDATE: Just rebased with develop and nothing seems to be on fire 👍)

## Screenshots
Screenshot from 2020-04-17 08-28-17
Screenshot from 2020-04-17 08-28-30
Screenshot from 2020-04-16 09-21-55


app/views/Settings.js
```diff
@@ -24,8 +32,22 @@ import Colors from '../constants/colors';
 class SettingsScreen extends Component {
   constructor(props) {
     super(props);
+
+    // Get locales list from i18next for locales menu
+    let localesList = [];
+    for (let key in languages.options.resources) {
```
tstirrat, line 38: Same thing here. This is copied code from Onboarding1.js

A service or a hook would be great.

```js
const [locale, setLocale, availableLocales] = useLocaleSetByUser();
```

app/views/Settings.js
```diff
@@ -40,12 +62,36 @@ class SettingsScreen extends Component {
...
+      })
+      .catch(error => console.log(error));
   }
 
   componentWillUnmount() {
     BackHandler.removeEventListener('hardwareBackPress', this.handleBackPress);
   }
 
+  locationToggleButtonPressed() {
+    this.state.isLogging ? LocationServices.stop() : LocationServices.start();
```
tstirrat, line 86: If this failed, then what happens to the state stored in`PARTICIPATE`? Not the responsibility of services, hence why I recommend this is a service or hook.

app/views/Settings.js
```diff
@@ -114,35 +160,42 @@ class SettingsScreen extends Component {
...
+              maxHeight: 24,
+              marginRight: 12,
+              marginTop: 2.5,
+            })}
+            {subtitle ? (
+              <Text
+                style={[
+                  styles.settingRowText,
+                  {
+                    color: color || Colors.VIOLET_TEXT,
```
tstirrat, line 183: There is a theme aware `<Typography use="body1">` for this. Wrap the entire screen in `<Theme use="default">` and it all contained `<Typography>` will give you the violet text that you need.

If you need bold, you can `use="headline2"` or similar.

app/views/LocationTracking.js
```diff
@@ -106,23 +110,31 @@ class LocationTracking extends Component {
...
-          case RESULTS.BLOCKED:
-            console.log('NO LOCATION');
-            LocationServices.stop();
-            this.setState({ currentState: StateEnum.UNKNOWN });
-        }
-      })
-      .catch(error => {
-        console.log('error checking location: ' + error);
-      });
+    // If user has location enabled & permissions, start logging
```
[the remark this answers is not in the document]
tstirrat: Steve recommends we tackle service extraction in a different PR. SGTM

tstirrat:
Checked on iOS and Android across many languages and on small screens.

PR #550, opened 2020-04-17
Feature/update italian

matteocrippa:
## Description

- Fix typo in italian localization
- Make some sentences more human readable and common use
- Translates new section history into italian
- Fix typo in eng localization

## How to test

Think there's no UI test automation for localization atm.

PR #553, opened 2020-04-17
Add space into carrythe to make into two words - smallest PR ever!

summetj:
## Description
Adds a space between carrythe in the en.json locales file to make it "carry the" instead.

## How to test
Verify that only one space was changed in the string, make sure that it reads better with the space.

Note: The numbers of characters I had to type on the command line to implement this fix is insane compared to the size of the fix ;>


alpita-masurkar:
https://github.com/tripleblindmarket/covid-safe-paths/pull/550
this PR also mentions and takes care of it.

summetj:
> #550
> this PR also mentions and takes care of it.

Didn't know others had already jumped on it, feel free to delete/merge this one.

PR #554, opened 2020-04-17
FIX: add covid-safe-paths to the end of the git clone command example to make it correct

summetj:
## Description

Updates an example line in the CONTRIBUTING.md file to include the name of the repository
covid-safe-paths so that the command works correctly.

Added in an explicit change directory to the newly created directory 
cd covid-safe paths

so that the next command (adding upstream)  would work properly.

## How to test
Follow the directions to set up a new fork on your local machine and verify that the new commands work.   [I tested on Linux Ubuntu 18.04 ]


tstirrat:
Thanks!!

PR #556, opened 2020-04-17
Allow VS Code to type hint `jest` global

tstirrat:
## Description

This allows autocomplete for jest global in test files!

## How to test

1. `yarn install`
1. open vscode, go to any test and type `jest.` and you should see jest functions :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@babel/runtime": "^7.8.4",
     "@jumpn/react-native-jetifier": "^0.1.4",
     "@react-native-community/eslint-config": "^0.0.7",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -83,6 +83,7 @@
     "@babel/runtime": "^7.8.4",
     "@jumpn/react-native-jetifier": "^0.1.4",
     "@react-native-community/eslint-config": "^0.0.7",
+    "@types/jest": "^25.2.1",
```
tstirrat, line 86: yes this is required, even thought its a Typescript thing :)

PR #558, opened 2020-04-18
Add a couple of yarn shortcut scripts

tstirrat:
## Description

```
yarn start # runs yarn && react-native start
yarn android # runs react-native run-android
yarn ios # ditto with ios
```


tstirrat:
> Probably more clear as `run-android` and `run-ios`. 

Yeah, good point

>`react-start` would be better as well, but that might just be a personal preference.

`yarn start` (and `npm start`) is a convention in a lot of js repos, it's always the thing that gets your dev server up.

https://reactnative.dev/docs/environment-setup

```sh
expo init AwesomeProject

cd AwesomeProject
npm start # you can also use: expo start
```

I think we didnt get it because we upgraded from 0.5x to 0.6x along the way

> n.b. We also have the `1_start_react` and `2_start_android` script that I know @penrods likes. Probably ought to talk with him before we remove them, but would be good to not have too many ways to do the same things.

I was thinking of *not* removing them, given that some have their workflows established.

`./2*.sh` it android only. I could look at making a second ./2_ to be consistent.. but then you'd have collision when tab completing `./2<tab>`

Another point re js repo convention is `yarn run` will list out all your commands, so there's a nice lookup table of contents

I think we could have both flows side by side.

lfalkner:
I'd say it's confusing to have multiple different scripts with the same intent. This is potentially problematic if we need to add a flag to any of them, like  --jetify  or  --ignore-engines, etc. 

`react-native run-android` and `npx react-native run-android` should  do the same thing,  and  we  may want  to  force the  jetification flag onto  the yarn install command, rather than expect users to  use two different tools to interact with react-native  (npx and yarn).

Moreover, the shell script appears to do  nothing more than actually run `npx react-native  run-android` so I'm curious  as to the benefit  of it.

tstirrat:
`npx react-native` is awful to type tbh, I'm not sure why, and it could just be me. 

`yarn` and `npm` seem way less annoying.

tstirrat:
the benefit of the shell scripts as I see it is `./2<tab>` is was easier to type than the expanded forms

though I didn't write them. I used them a lot until I needed `npx react-native run-ios` all the time

tstirrat:
Also, `jetification` already happens on postinstall now. I don't think we manually run it anymore

lfalkner:
why not wrap  the scripts in  a  yarn  command  vs. a  shell  script? That way  it works  and  is called consistently cross-platform.  Also, starting metro should happen automatically via `react-native  run`.

tstirrat:
> why not wrap the scripts in a yarn command vs. a shell script?

Isn't that what I'm doing here with `yarn run-ios` ?

I think there are some who work better and are more familiar with the shell scripts. @kenpugsley and @penrods both use them AFAIK, I'm proposing similar but additional scripts for the npm/yarn fashionistas like me

lfalkner:
only concern I see  with it is having to update the commands  in multiple  places if  we  need to change anything

PR #565, opened 2020-04-18
COVID-19 name consistent on all screens

### LINKED ISSUE #548: Is it COVID19 or COVID-19?
Seen on v0.9.4, with Android 9 on Moto G7.

WIthin the space of a single screen, we refer to the disease by 2 different names:

COVID19 and COVID-19.

Please can we be consistent on all screens throughout the app?  (COVID-19, with the dash, is what I'd expect, but happy with either as long as we are consistent).

-->
image

NethravathiPuttraju:
Resolves #548 ,
To make COVID-19 name consistent on all screens.


matteocrippa:
@NethravathiPuttraju worth check if this happens also in other localizations

PR #566, opened 2020-04-18
Make settings test independent of locales

tstirrat:
`Settings.js` test lists out all locales, so any addition to locales requires an update to the snapshot... this makes the test independent of real locales.


tstirrat:
Superseded by #571

PR #567, opened 2020-04-18
Get detox in workflows

efalkner:
WIP


.github/workflows/test.yml
```diff
@@ -34,3 +34,65 @@ jobs:
       - run: yarn test
         env:
           CI: true
+
+  build-ios:
```
tstirrat, line 38: we already have build-ios, and similar job names can confuse GH, make this e2e-ios

e2e/pages/EnableLocation.po.js
```diff
@@ -0,0 +1,19 @@
+const buttonlabel = 'Enable Location';
+const screenshotText = 'Enable Location Page';
+const screenShotWithMenuText = 'Location Permissions Dialog';
+
+class EnableLocation {
+  tapButton = async () => {
```
tstirrat, line 6: these don't need `this` binding so normal form is fine

```js
async tapButton() {}
```

ios/Podfile
```diff
@@ -3,6 +3,8 @@ require_relative '../node_modules/@react-native-community/cli-platform-ios/nativ
 
 target 'COVIDSafePaths' do
   # Pods for COVIDSafePaths
+  project 'COVIDSafePaths.xcodeproj'
```
tstirrat, line 6: only needed if you have old PrivateKit artifacts under `ios/`

But probably work committing.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
tstirrat, line 1: remove this file

app/components/Button.js
```diff
@@ -36,7 +36,10 @@ class Button extends React.Component {
             buttonStyle ? buttonStyle : styles.container,
             { height: buttonHeight },
           ]}
-          onPress={onPress}>
+          onPress={onPress}
+          accessible
```
tstirrat, line 40: is `accessible` needed if the label is provided?

.eslintrc.js
```diff
@@ -0,0 +1,68 @@
+module.exports = {
```
tstirrat, line 1: if you can get this to be marked as a file move that would be helpful to preserve blame/history :)

ios/Podfile
efalkner: This is from a merge, not intentional commit on my part

app/components/Button.js
efalkner: from a merge

.eslintrc.js
efalkner: will look into how

e2e/pages/EnableLocation.po.js
efalkner: Yup, cruft from a previous approach. will remove

.eslintrc.js
efalkner: Spoke with Tim about this offline, we determined it wasn't a blocker

ios/Podfile
tstirrat: its fine to commit. it helps when people have the old PrivateKit files left in their workdir.

ios/Podfile.lock
```diff
@@ -23,9 +23,9 @@ PODS:
     - DoubleConversion
     - glog
   - glog (0.3.5)
-  - Permission-LocationAlways (2.0.10):
+  - Permission-LocationAlways (2.1.2):
```
tstirrat, line 26: Are these pod updates needed?

package.json
```diff
@@ -43,6 +47,7 @@
     "@testing-library/react-native": "^5.0.3",
     "dayjs": "^1.8.24",
     "emotion-theming": "^10.0.27",
+    "eslint-plugin-detox": "^1.0.0",
```
tstirrat, line 50: duplicated in dev and deps

ios/Podfile.lock
efalkner: Just a result of the merge.

package.json
efalkner: Should only be in dev. Will update.

.vscode/settings.json
tstirrat: Revert this file and then I'm happy and ready to merge!

e2e/pages/Onboarding3.po.js
```diff
@@ -0,0 +1,20 @@
+const buttonlabel = 'Next';
+const screenText =
+  'If you test positive, you can choose to donate your data anonymously';
+const screenshotText = 'Third Onboarding Page';
```
tstirrat, line 4: nit: these come out unordered in filesystem, perhaps `Onboarding - Page 3` and so on

Issue #564, opened 2020-04-18
Confusing notification text on "stop logging" action


summetj:
## Version 0.5.19 private kit from play store, verified exists in development code as well. 

- Type of phone: Android 
- OS Version: 9
- Model: Moto X4

## Steps to reproduce

Tap settings icon in top left to get to the dashboard
Tap the "stop logging" button.

Notification text reads: "Location tracking was Disabled" (good) followed by "PrivateKit requires location services" (technically true, but confusing, as it implies the app does not have access to location services.)

## Expected Behavior

Suggest we change the second line to: "Stopped at user request" in the case tracking was disabled by the user as opposed to the loss of location services. [Or just "Stopped" if the stop action can be triggered by things outside of user requests...]

![Screenshot_20200418-063118](https://user-images.githubusercontent.com/11669847/79635596-d4b86b00-813f-11ea-84d5-8b1d8c840055.png)

I suspect this is caused by a lazy reuse of the label.location_disabled_message in the app/services/LocationService.js file for both the stop action and the checkStatus action.

Suggested fix: Add a new label "location_stopped_message", get it translated for all languages, and differentiate between a stop action and an "if (!locationServicesEnabled)" check failure.

aartiverma1:
I can start looking this issue.

tremblerz:
yes please

PR #574, opened 2020-04-19
Issue 564 - change message ( fixed english only )

aitkeni:
## Description

resolves issue 564. the message was confusing so was changed (in the english resourse only ) to the suggested stopped when location services are disabled or stopped at user request
## How to test
1 go to settings,
click on stop logging text or icon, 
see message appears.  it should now read
Location Tracking was disabled
Stopped at user request.
2 go to settings
turn off location services from OS.
click on stop logging text or icon
see message appears.  it should now read
Location Tracking was disabled
Stopped.

Still todo other languages (please advise!)
(In another locale the english text will appear until the translation is present I think)

### DIFF
--- a/app/locales/en.json
+++ b/app/locales/en.json
@@ -64,6 +64,8 @@
     "location_disabled_title": "Location Tracking Was Disabled",
     "location_enabled_message": "COVID Safe Paths is securely storing your GPS coordinates once every five minutes on this device.",
     "location_enabled_title": "COVID Safe Paths Enabled",
+    "location_stopped_message": "Stopped",
+    "location_stopped_at_user_request_message": "Stopped at user request",
     "maps_import_button_text": "Import past locations",
     "maps_import_disclaimer": "Safe Paths has no affiliation with Google and never shares your data.",
     "maps_import_text": "To see if you encountered someone with COVID-19 prior to downloading this app, you can import your personal location history.",
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -366,9 +366,22 @@ export default class LocationServices {
 
   static stop() {
     // unregister all event listeners
-    PushNotification.localNotification({
-      title: languages.t('label.location_disabled_title'),
-      message: languages.t('label.location_disabled_message'),
+    BackgroundGeolocation.checkStatus(({ locationServicesEnabled }) => {
+      if (!locationServicesEnabled) {
+        PushNotification.localNotification({
+          id: LOCATION_DISABLED_NOTIFICATION,
+          title: languages.t('label.location_disabled_title'),
+          message: languages.t('label.location_stopped_message'),
+        });
+      } else {
+        PushNotification.localNotification({
+          id: LOCATION_DISABLED_NOTIFICATION,
+          title: languages.t('label.location_disabled_title'),
+          message: languages.t(
+            'label.location_stopped_at_user_request_message',
+          ),
+        });
+      }
     });
     BackgroundGeolocation.removeAllListeners();
     BackgroundGeolocation.stop();


tremblerz:
Can you add more verbose message keys and preferable title also?
Something like - "Device location has been disabled. Please enable it in the settings" and "Location access to Safe Paths has been removed. Please enable it in the settings for effective logging" (Or instead we can do the same on-boarding kind of process also from UX perspective)
Ex. - Uber says: "For a better location experience, turn on the device location"

Given that the PR is dedicated towards only changing the message which is same for the both case right now, I would suggest only change text and get it in, and make a ticket for much in depth UI/UX for the events of location being disabled/enabled etc.

aitkeni:
Makes sense.  Leave it with me!

aitkeni:
Actually One clarification @tremblerz , the proposed change will distinguish between if the location services have been switched off externally or if its the user clicking the stop tracking icon.  So what you are suggesting is to change the message when location services have been disabled externally
to
   "location_disabled_title": "Device location has been disabled. Please enable it in the settings",
   "location_disabled_message": "Location access to Safe Paths has been removed. Please enable it in the settings for effective logging.",

I cant help thinking that the title and message duplicate information, as they are both displayed together at all times.  the message can show more info as the font is smaller.
would it not make more sense to have the title convey the current state, and the message to convey the reason for the state and the suggested resolution. 
"location_disabled_title": "Device location has been disabled.",
   "location_disabled_message": "Location access to Safe Paths has been removed. Please enable it in the settings for effective logging.",

How does that sound?

alpita-masurkar:
> Actually One clarification @tremblerz , the proposed change will distinguish between if the location services have been switched off externally or if its the user clicking the stop tracking icon. So what you are suggesting is to change the message when location services have been disabled externally
> to
> "location_disabled_title": "Device location has been disabled. Please enable it in the settings",
> "location_disabled_message": "Location access to Safe Paths has been removed. Please enable it in the settings for effective logging.",
> 
> I cant help thinking that the title and message duplicate information, as they are both displayed together at all times. the message can show more info as the font is smaller.
> would it not make more sense to have the title convey the current state, and the message to convey the reason for the state and the suggested resolution.
> "location_disabled_title": "Device location has been disabled.",
> "location_disabled_message": "Location access to Safe Paths has been removed. Please enable it in the settings for effective logging.",
> 
> How does that sound?

On that note, you also don't need 'has been' in these two strings. The meaning is still conveyed and is shorter and might look better on screen.

"location_disabled_title": "Device location disabled.",
"location_disabled_message": "Location access to Safe Paths removed. Please enable it in the settings for effective logging.",

aitkeni:
agreed

alpita-masurkar:
#564

summetj:
Sorry, missed this pull request because it was not linked to the issue https://github.com/tripleblindmarket/covid-safe-paths/issues/564.   Should be merged with https://github.com/tripleblindmarket/covid-safe-paths/pull/647

tstirrat:
I did a test on this (Pixel 2 emulator), and it gets stuck showing "stopped at user request" even though I actually have location permissions disabled in the OS.

I would expect it to say "Location access to Safe Paths removed. Please enable it in the settings for effective logging." right?

I suspect we need more thorough checks..

Can you show a video of how it should be working?

aitkeni:
Thanks for helping me on this @tstirrat . 
heres a vid of the way the new messages are showing.
[2020-04-25 00-53-20.mkv.zip](https://github.com/tripleblindmarket/covid-safe-paths/files/4532044/2020-04-25.00-53-20.mkv.zip)
Unfortunately I just noticed that the new suggested text for when location services are disabled was too wordy and spills over uncomfortably onto two lines and the message is truncated.  
image
I think I should be shortened again.
Also I could reproduce the issue you are seeing when the loc services are disabled then the ui shows active when bringing the app back into the foreground
[2020-04-25 01-10-59.mkv.zip](https://github.com/tripleblindmarket/covid-safe-paths/files/4532059/2020-04-25.01-10-59.mkv.zip)
i dont think this is an issue i have introduced with my change (but i could be wrong) 
If I confirm its an existing defect then I will raise a new issue and I can take a look this weekend, but will shorten that message first.
many thx.

aitkeni:
just confirmed that this defect exists without my changes.  in fact when location services disabled it gets a bit messed up.  the app can think location tracking is enabled when its not as well as getting stuck at Location tracking disabled ( which is the original message before my changes )  Ill raise an issue for this.

aitkeni:
issue #685 raised

aitkeni:
apologies for the delay in updating.  I have shortened the messages so they fit into the android notification message area.  I have also significantly changed the word 'securely' to 'anonymously' in the Enabled message.  I think it is more accurate and better for PR (it seems all the press is talking about how contact tracing is bad for your privacy rights) so I thought this would be a better message and would make our app rise above...  regarding the issues you saw @tstirrat i have a separate PR coming to fix that issue (issue-685)

tstirrat:
Superseded by #725

PR #577, opened 2020-04-19
Tweaks to better support small screens

penrods:
This fixes issues on phone screens as small as 320 pixels wide in various
parts of the application.
* Navigation bar text was too long (even in English) under the Choose Health Authority.
  When the screen width is < 400 pixels the font-size begins to scale.  It also has enforced
  horizontal padding to prevent overlap with the back arrow -- text will now wrap instead of
  overlap.  This change also helps with some translations which make it even longer.
* The Exposure History legend was too long.  I truncated the text "No exposure" to
  "No known" and the "Possible Exposure" to just "Possible"
* The "Choose health authority" screen description text was being truncated.  I changed
  the flex size of the list below it form 3 to 1 and tweaked the font size.  This
  comfortably shows two entries even on a 320x533 screen in the emulator before that list
  requires scrolling, which will probably account for 90%+ of the use cases.
* Slightly shrank the text in the News title for each webview and centered it so it looks
  better if it wraps.

 ## Tech Notes ##
It seems like modifying Typography.js to scale might be a good idea, but I didn't want to
go that big with this PR.

 ## Translation Notes ##
The text for the two entries "Possible exposure" and "No exposure" have changed.  It is
important that those along with "No data" are kept very short, as they appear side-by-side.
For example, the current Kreole translation for "No data" is "Pa gen okenn done" and
gets cut off even with the short English labels for the other two.

 ### Testing Notes ###
Run on a small or large screen.


app/views/ChooseProvider.js
```diff
@@ -68,8 +68,11 @@ class ChooseProviderScreen extends Component {
     // Update user settings state from async storage
     GetStoreData(AUTHORITY_SOURCE_SETTINGS, false).then(result => {
       if (result !== null) {
-        console.log('Retrieving settings from async storage:');
+        console.log('Steve Retrieving settings from async storage:');
```
kenpugsley, line 71: Probably need to undo the 'Steve' here!  ;-)

app/locales/en.json
```diff
@@ -102,13 +102,13 @@
     "logging_inactive": "Location Inactive"
   },
   "history": {
-    "no_exposure": "No exposure",
+    "no_exposure": "No known",
     "possible_exposure_para": "It is possible you were in contact with or close to someone who tested positive for COVID-19",
-    "possible_exposure": "Possible exposure",
+    "possible_exposure": "Possible",
```
alpita-masurkar, line 107: Given the context of what else is now on this/these screen(s), is there a way to make sure that the words Possible and No known are enough to convey the meaning to an average user?
Maybe a screenshot of these screens and show it to a few multilingual people or the ones not immediately familiar with the screens?

app/views/ChooseProvider.js
alpita-masurkar: Remove the name from the log? :)

alpita-masurkar:
I think this PR was needed and it's great.
Just a heads up about that- it is likely that in other languages, the translations may not be as clean or short as in English, unfortunately (it's definitely not short in the other languages that I speak). So we might still get some longer translations even for shorter words/ phrases and I hope they'll not look too bad in the space that they have on screen.

PR's approved after the log cleaning.

app/views/ChooseProvider.js
```diff
@@ -70,6 +70,9 @@ class ChooseProviderScreen extends Component {
       if (result !== null) {
         console.log('Retrieving settings from async storage:');
         console.log(result);
+        result.push({ key: 'test', url: 'http://gpll.org' });
```
tstirrat, line 73: remove test stuff.

kenpugsley:
@penrods ... sorry ... my mistake for clicking the wrong buttons.  Both @tstirrat and I approve once you have removed the gpll.org reference that Tim pointed out.

penrods:
Here is a shot of the Exposure History on a small screen.
image

app/locales/en.json
penrods: I attached a picture of the small screen.

PR #572, opened 2020-04-19
use react-i18next useTranslation hook

tstirrat:

PR #571, opened 2020-04-19
Improve device locale + user override separation

tstirrat:
Separates the device detection + initial set, locale setting code, and user override flags in a way that makes intent very clear.

PR #557, opened 2020-04-18
Refactor NativePicker, detect device language better

tstirrat:
## Description

This takes out the hard coded UI in `<NativePicker>` and allows anyone to specify how it draws it's resting state label UI. It can now easily be used for anything. Settings and Onboarding1 give examples.

Also detects device language against known language list, rather than forcing everyone to english by default.

## How to test

Test onboard and settings dropdowns to make sure they still work!

Screen Shot 2020-04-17 at 5 15 10 PM
Screen Shot 2020-04-17 at 5 15 17 PM
Screen Shot 2020-04-17 at 5 15 27 PM
Screen Shot 2020-04-17 at 5 15 30 PM
Screen Shot 2020-04-17 at 5 16 16 PM
Screen Shot 2020-04-17 at 5 16 40 PM

PR #581, opened 2020-04-19
Feature-Rtl : Add Arabic Language + handling Right to left languages UI and App behavior

Ahmednagyyy:
## Description

As its a global Application we need it to support many languages as we can, and some of these languages are RTL (Right to left) languages.

**This PR contains the following** 
- Add **ar.json** for Arabic language locales 
- Add RTL language files ISO codes to check the selected language is included within the file or not (12 languages)
- Add React-native-restart package to enable app restarting **because** Mobile Apps language direction change require app restart to apply the new language direction to the UI
- Add **DynamicText** as a reusable component, It inherits the react native **Text** but change text direction based on the selected language  **( Should be used for any new text )**
- Add **DynamicTextInput** as a reusable component, It inherits the react native **TextInput** but change text direction based on the selected language  
- Changed All React native **Text** to **DynamicText** and **TextInput** to **DynamicTextInput**
- Add **languageDirectionHandler** function to languages file as it detects and change current language direction
- Add **languageDirectionHandler** to NativePicker to check if any change to language direction to restart the application to apply the new direction
- Add a new Prop to **NativePicker** named **languagePicker** to make the component know when to use the new **languageDirectionHandler** function , **WHY** to reuse the Picker as we want for all types of pickers
- Add RTL or LTR direction styles to each of **Typography** and **NavigationBarWrapper** component

## How to test

1- Now the only RTL language is the Arabic one
2- If you selected RTL language App will ask you to restart if your current direction is LTR, and vice versa
3- All Views should be shifted to the right when you select RTL language


app/locales/languages.js
```diff
@@ -68,6 +72,41 @@ export function findUserLang(callback) {
   });
 }
 
+// This function is for checking/change current app direction
+export function languageDirectionHandler(language, navigation) {
+  // check if is not rtl and user choose rtl language
+  if (!I18nManager.isRTL && rtlLanguages.includes(language)) {
```
tstirrat, line 78: I've have changes in #571 to separate device locale and user local overrides. I hope to use `i18next` as the source of truth (because active locale vs device local can be different)

In that case, can you use `i18next.dir()`

Can you base it off the ideas in that branch?

Then with Typography, we can do `const {i18n} = useTranslation()` and check `i18n.dir(currentLocale)`

app/locales/languages.js
```diff
@@ -68,6 +72,41 @@ export function findUserLang(callback) {
...
+    );
+  }
+
+  // check if is rtl and user choose ltr language
+  if (I18nManager.isRTL && !rtlLanguages.includes(language)) {
+    showAlert(
+      null,
+      'App need to restart in order to use this language',
+      () => [I18nManager.forceRTL(false), ReactNativeRestart.Restart()],
+      'Restart',
```
tstirrat, line 95: These will need translation.

app/locales/rtl-languages.js
```diff
@@ -0,0 +1,19 @@
+// This file contains all RTL languages ISO code just in case someone added one of this languages
+// Used in OnBoarding1.js file to set app to rtl if user pick one of these languages
+
+const rtlLanguages = [
```
tstirrat, line 4: this should be detectable from `i18n.dir('ar')`

app/views/About.js
```diff
@@ -84,12 +85,12 @@ class AboutScreen extends Component {
           <View style={styles.spacer} />
 
           <SvgXml style={styles.aboutSectionIconLock} xml={lock} />
-          <Text style={styles.aboutSectionTitles}>
+          <DynamicText style={styles.aboutSectionTitles}>
```
tstirrat, line 88: Please converge on `Typography`, all text will eventually use `<Typography>`, we are gradually converging on fixed headline/body styles from [Figma](https://www.figma.com/file/n8fjoYkgJybfYgrOvLRMBf/Safe-Paths---Draft-designs?node-id=1358%3A11099)

app/Util.js
```diff
@@ -13,8 +13,49 @@ export function nowStr() {
   return dayjs().format('H:mm');
 }
 
+export function isEmpty(obj) {
+
+  // null and undefined are "empty"
+  if (obj == null) return true;
+  if (obj == 'indefined') return true;
```
tstirrat, line 20: This seems incorrect? do you mean `typeof obj === 'undefined'`

app/Util.js
```diff
@@ -13,8 +13,49 @@ export function nowStr() {
   return dayjs().format('H:mm');
 }
 
+export function isEmpty(obj) {
```
tstirrat, line 16: function is not used, can we remove?

app/components/DynamicText.js
```diff
@@ -0,0 +1,28 @@
+// Custome Text handling RTL and LTR text direction
+
+import React, { Component } from 'react';
+import { Text , StyleSheet, I18nManager} from 'react-native';
+
+
+class DynamicText extends Component {
```
tstirrat, line 7: let's converge on Typography and make it RTL aware.

app/components/DynamicTextInput.js
```diff
@@ -0,0 +1,28 @@
...
+     
+}
+const styles = StyleSheet.create({
+    defaultTextInputStyle: {
+        writingDirection: I18nManager.isRTL ? 'rtl' : 'ltr'
+    }
+  });
+  
+
+export default DynamicTextInput;
```
tstirrat, line 28: Prefer not to use default exports. they make the intellisense less intelligent. With named exports, it will find things easier.

app/components/NavigationBarWrapper.js
```diff
@@ -83,6 +85,7 @@ const BackArrow = styled.TouchableOpacity`
   align-items: center;
   height: 55px;
   justify-content: center;
+  transform: ${I18nManager.isRTL ? 'rotateY(180deg)' : 'rotateY(0deg)'};
```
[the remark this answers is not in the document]
Ahmednagyyy: Writing-direction is only for Texts
So for views people have 2 options
1- transform the view ( already implemented )
2- use 2 icons left and right but I don’t like to have a lot of assets

app/components/Typography.js
```diff
@@ -93,6 +94,7 @@ const getFontFamily = ({ use, monospace, bold }) =>
     : 'IBMPlexSans';
 
 const ThemedText = styled.Text`
+  writingDirection: ${I18nManager.isRTL ? 'rtl' : 'ltr'};
```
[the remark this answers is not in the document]
Ahmednagyyy: Will check the syntax, But it works as expected

app/Util.js
Ahmednagyyy: Function removed

app/Util.js
Ahmednagyyy: Function removed

app/components/DynamicTextInput.js
Ahmednagyyy: Done, and converted as Typography, and need to add themes later on

app/components/DynamicText.js
Ahmednagyyy: Removed and now using Typography as the default app text

app/views/About.js
Ahmednagyyy: Done, and need to choose the "bodyX|headlineX" (Put I prefer making this with new branch after merging this for styling task only)

app/locales/rtl-languages.js
Ahmednagyyy: Removed and reading the direction directly from i18next

app/locales/languages.js
Ahmednagyyy: Added for english and arabic needed to be updated for the other languages too as most of them missing alot of translations

app/locales/languages.js
Ahmednagyyy: Now we are using i18next as only source of truth about current language and directions
And Typography is now updated to be more consistent 
`const getCurrentDirection = () => languages.dir();`
`const ThemedText = styled.Text`
`writing-direction: ${getCurrentDirection};`

app/components/Typography.js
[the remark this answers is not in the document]
Ahmednagyyy: Done

app/components/NavigationBarWrapper.js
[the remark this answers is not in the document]
Ahmednagyyy: I've added a new method to be more consistent 
`const getCurrentDirection = () => languages.dir() === 'rtl' ? 'rotateY(180deg)' : 'rotateY(0deg)';
`
`const BackArrow = styled.TouchableOpacity`
 `transform: ${getCurrentDirection};`

app/views/Settings.js
```diff
@@ -63,12 +64,15 @@ export const SettingsScreen = ({ navigation }) => {
   const localeChanged = async locale => {
     setUserLocale(locale);
 
-    // If user picks manual lang, update and store setting
-    try {
-      await languages.changeLanguage(locale);
-      await SetStoreData(LANG_OVERRIDE, locale);
-    } catch (e) {
-      console.log('something went wrong in lang change', e);
+    if (languages.dir() === languages.dir(locale)) {
```
tstirrat, line 67: only change language if it's the same dir?

Is that because elsewhere it refreshes the app if dir is changed?

The side effect is not obvious.

app/views/onboarding/Onboarding1.js
```diff
@@ -58,35 +58,38 @@ class Onboarding extends Component {
                 zIndex: 10,
               }}>
               <NativePicker
+                languagePicker // pass picker type to handle RTL checking because we can reuse this picker for any type not language picker only
```
tstirrat, line 61: Ideally, language specific stuff should be in the onValueChanged handlers.

Check #557 with upcoming changes. If you keep your language stuff in onValueChanged then both changes can coexist.

android/app/build.gradle
```diff
@@ -197,6 +197,7 @@ android {
 }
 
 dependencies {
+    implementation project(':react-native-restart')
```
tstirrat, line 200: is this needed because some of the i18n stuff is set globally and is somewhat static?

Do you think once we move to react-i18next hooks everywhere, this will go away?

Ahmednagyyy: The default mobile application for the RTL ( Native behavior ) is restarting the application or change the language from settings ( you can try with facebook for example )
So for example if we don't want to restart we make our views detects the direction and shift every time it changes and there are many approach to do that
**For example In native android I was creating 2 design xml files **rtl** and **ltr** to make android framework handle this by default**  
But here we will do the same thing we did for **typography** to all views

app/components/NavigationBarWrapper.js
```diff
@@ -33,7 +34,9 @@ const NavigationBarWrapper = ({ children, title, onBackPress }) => {
       <BottomContainer>
         <Header>
           <BackArrow onPress={() => onBackPress()}>
-            <BackArrowIcon xml={backArrow} />
+            <BackArrowIcon 
+            
+            xml={backArrow} />
```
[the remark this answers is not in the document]
Ahmednagyyy: done

app/components/DynamicTextInput.js
```diff
@@ -0,0 +1,11 @@
+// Custome Text handling RTL and LTR text direction
+import styled from '@emotion/native';
+import languages from '../locales/languages';
+
+// Themes to be added
+
+const getCurrentDirection = () => languages.dir();
+
+export const DynamicTextInput = styled.TextInput`
+writingDirection: ${getCurrentDirection};
```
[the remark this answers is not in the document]
Ahmednagyyy: done

app/components/NativePicker.js
```diff
@@ -82,7 +83,12 @@ export default class NativePicker extends Component {
             <View style={{ flex: 2 }}>
               <TouchableWithoutFeedback
                 style={{ flex: 1, backgroundColor: '#000000', opacity: 0.4 }}
-                onPress={() => this.setState({ modalVisible: false })}>
+                onPress={() => {
+                  // check if its a language picker we check for RTL to update the view
+                  if (this.props.languagePicker) languageDirectionHandler(this.props.value, () => this.setState({ modalVisible: false }))
```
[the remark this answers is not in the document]
Ahmednagyyy: Prettier format fixed it

android/app/build.gradle
tstirrat: I see, if we have to encode the rtl logic into the views anyways, it would be nice if they re-flowed without a restart. 

(later, in a different PR) We could probably make a simple hook:

```js
const dir = useWritingDirection();
```

though, this is not that much different than `const {i18n} = useTranslation()`

app/locales/languages.js
```diff
@@ -90,6 +132,7 @@ i18next.init({
     sk: { label: 'Slovak', translation: sk },
     vi: { label: 'Vietnamese', translation: vi },
     zh_Hant: { label: '繁體中文', translation: zh_Hant },
+    ar: { label: 'العربية', translation: ar },
```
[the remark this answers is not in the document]
Ahmednagyyy: Now sorted by language ISO code

app/locales/languages.js
```diff
@@ -34,7 +37,7 @@ import zh_Hant from './zh-Hant.json';
 const deviceLocale =
   Platform.OS === 'ios'
     ? NativeModules.SettingsManager.settings.AppleLocale || // iOS < 13
-      NativeModules.SettingsManager.settings.AppleLanguages[0] // iOS 13
+    NativeModules.SettingsManager.settings.AppleLanguages[0] // iOS 13
```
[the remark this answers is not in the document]
Ahmednagyyy: I can't understand it? does it affects the PR context ?

app/locales/languages.js
```diff
@@ -70,6 +73,40 @@ export function findUserLang(callback) {
...
+export function languageDirectionHandler(language, navigation) {
+  // check if is not rtl and user choose rtl language
+  if (i18next.dir() !== i18next.dir(language)) {
+    showAlert(
+      null,
+      i18next.t('label.language_change_alert_title'),
+      () => [
+        localeChanged(language),
+        I18nManager.forceRTL(i18next.dir(language) === 'rtl' ? true : false),
+        ReactNativeRestart.Restart(),
```
tstirrat, line 96: This restart stuff is not sitting well with me. It's like a page refresh on the web. 

* It's not going to play well with tests because the whole app restarts, losing state/context.
* There are reports of it being a **memory leak** and can end up with **crash loops**.

> Warning about using react-native-restart for this (and in general): if any problem occurs during your restart process, the app can fall into crash-restart loop that may see your app get rated as poor quality on app stores. It often works, but it isn't a good solution to this LTR<>RTL problem.

https://github.com/facebook/react-native/issues/17867#issuecomment-423858132

Can you investigate making the RTL aware components more reactive? i'm pulling in `react-i18next` in #572 which will allow better reactivity.

Ahmednagyyy: Added and started the change with Typography and settingsItem, should update other components too in the furture

app/views/onboarding/Onboarding1.js
Ahmednagyyy: Translation logic removed from the Picker

app/views/Settings.js
Ahmednagyyy: Removed as now components aware of the direction

app/locales/languages.js
```diff
@@ -73,43 +73,14 @@ export function findUserLang(callback) {
...
-      null,
-      i18next.t('label.language_change_alert_cancel'),
-    );
-  } else {
-    localeChanged(language);
-    if (navigation) {
-      navigation();
-    }
-  }
+export function useDirHook() {
```
tstirrat, line 76: I like the idea here, if this works, use it throughout.

on naming convention, starts with `use`, remove `Hook` so `useLanguageDirection` or something like that.

tstirrat:
#557 makes a change to detect device locale better too

ios/Podfile.lock
```diff
@@ -206,6 +206,8 @@ PODS:
     - React
   - react-native-viewpager (3.3.0):
     - React
+  - react-native-webview (8.2.1):
```
efalkner, line 209: This should no longer be in your podfile. Please update your podfile and remove Podfile.lock / run `bundle exec pod install`.

ios/Podfile.lock
```diff
@@ -306,6 +306,7 @@ DEPENDENCIES:
   - react-native-splash-screen (from `../node_modules/react-native-splash-screen`)
   - react-native-uuid-generator (from `../node_modules/react-native-uuid-generator`)
   - "react-native-viewpager (from `../node_modules/@react-native-community/viewpager`)"
+  - react-native-webview (from `../node_modules/react-native-webview`)
```
efalkner, line 309: Same as above

package.json
```diff
@@ -17,6 +17,7 @@
     "test:e2e:iphone8": "detox test -c iphone8.sim --loglevel=warn",
     "i18n:extract": "i18next",
     "test": "jest test",
+    "update-snapshots": "jest --updateSnapshot",
```
efalkner, line 20: I'm not sure if this is something we should encourage regularly. I'll defer to @tstirrat on that, but it seems like this encourages us to disregard changes in the snapshots, which negates their purpose.

app/Util.js
```diff
@@ -13,8 +13,25 @@ export function nowStr() {
   return dayjs().format('H:mm');
 }
 
+export function showAlert(title, description, onOkPress, okText, onCancelPress, cancelText) {
```
efalkner, line 16: I'm not sure I understand the purpose of this wrapper. It seems to do the same thing as the Alert native module.

ios/Podfile.lock
Ahmednagyyy: @efalkner This is out of scope of my PR here, Also its on the develop branch and used in LicensesScreen and News screen

ios/Podfile.lock
Ahmednagyyy: @efalkner I replied In the previous comment

package.json
Ahmednagyyy: @efalkner This was necessary as we refactored all text and new components were added 
@tstirrat Can confirm

app/Util.js
Ahmednagyyy: @efalkner To reduce code redundancy, every time someone write the full alert code to show one

ios/Podfile.lock
efalkner: :) I just added a PR to remove it. It's automatically linked, and therefore doesn't need to be added to the Podfile.

It looks like your PR is putting the link back in, perhaps because it's out of date.

Ahmednagyyy: @efalkner Its on the develop branch https://github.com/tripleblindmarket/covid-safe-paths/blob/develop/package.json 

Also as i said its been used in LicensesScreen and News screen, it will require refactor this part and its out of my PR scope

package.json
tstirrat: I think snapshot testing isn't great, and would like to do better/different testing.

Specifically this script is a helpful script while we're in this state. So I'd say, ok, cool.

We should discuss elsewhere about doing better testing, and when we have a better strategy, we can purge snapshot testing then  🔥

ios/Podfile.lock
efalkner: Sorry for the confusion. I'm not asking you to remove the library from the project, just from your Podfile.lock. If you look at the diff, your PR is reintroducing that dependency to it.

app/Util.js
efalkner: I like the thought of abstraction in general, but I'm not sure if this particular one is providing much value. It looks as though this hasn't really abstracted much away from the consumer, and could potentially introduce bugs, for example, if they don't pass in arguments in the right order.

ios/Podfile.lock
Ahmednagyyy: @efalkner Removed

efalkner: React-Native 60+ uses [auto linking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md). That means that for the majority of modules (react-native-webview included), it's not required to be linked manually. For iOS, that typically means adding an entry to the `Podfile`, which generates a `Podfile.lock` when you install cocoapods.

Given that this entry was not in the develop `Podfile.lock`, I can assume that the automatic linking was working properly in develop before your PR. That's why I'm thinking this got introduced unintentionally.

efalkner: ^ we spoke offline, closing the loop :)

app/components/NativePicker.js
```diff
@@ -3,16 +3,31 @@ import {
...
   View,
 } from 'react-native';
 
-import { DynamicTextInput } from './DynamicTextInput';
-import { Typography } from './Typography';
+import { Typography } from '../components/Typography';
+import languages from '../locales/languages';
 
-// Code for the language select dropdown, for nice native handling on both iOS and Android.
+/**
```
efalkner, line 13: 🥳

Issue #425, opened 2020-04-10
FEATURE REQ: Auto-subscribe to Healthcare Authorities


penrods:
Currently the user must manually subscribe to a Healthcare Authority.  This will grow to a point where it is a difficult UI to work with.  I propose the following:

1) Extend the Healthcare Authority yaml which lives on the safe-places repo to include a GPS bounding box.
2) If the user has no existing Healthcare Authority, the app checks that file once every 24 hours.  If the user's current GPS coordinate is within any bounding box, they should be given a popup notification that a Healthcare Authority exists in their are, "Would you like to subscribe?"
** This should only happen until they have the first authority to avoid overloading the server with millions of requests **
3) A similar process can be used when the user manually enters the Subscription page.  I'm not worried about server overload on a manual process.  This can limit the list to authorities which contain any of the points in their full 28-day GPS history, then have a checkbox that will show All Authorities.

The list which they are checked against should be a "production" URL from http://raw.githack.com/.  This will provide decent buffering of the Github server infrastructure, but it is VERY IMPORTANT we purge the cache anytime a new Healthcare Authority is added to the YAML.

This is the production (cached for a year) CDN URL:
https://rawcdn.githack.com/tripleblindmarket/safe-places/940edc081b8a4771deca806ed570eda907b5a261/healthcare-authorities.yaml

tstirrat:
@Patrick-Erichsen was looking into this, I think

Patrick-Erichsen:
I believe #375  is the same body of work. But this explanation is much more in-depth, so we could close #375 in favor of this.

Patrick-Erichsen:
@penrods  Is there any proposed schema for what the HCA YAML file might look like? The current one does not include the GPS bounding box required to complete this issue.

Patrick-Erichsen:
@penrods @tstirrat any updates on a proposed schema for the GPS bounding box of a Healthcare Authority?

 I'm planning on implementing an `isPointInRegion()` function that can be used to check if a user's GPS history has any points that are within the bounding box of an HCA, but without the schema for that bounding box, I can't get started.

DenisPitcher:
How would someone actually update the healthcare-authorities.yaml file to add an authority?

tremblerz:
Do you think we should just generate Geohashes on both sides?

penrods:
@tremblerz, this is really a bounding box so keeping it in distinct lat/lon is useful.  Pseudo-code is:
```
[...]

PR #587, opened 2020-04-20
Auto add authorities

Patrick-Erichsen:
## Description

This PR creates a new notification and filtering feature for Healthcare Authorities in a user's location.

### Issue
#425 

### Unit
`npx jest HCAService.spec.js`

### Integration
- Visit the `Choose health authority` screen and press the `Add Trusted Source` button. Press the GPS filter switch and confirm that all health authorities are removed (the current [live yaml file](https://rawcdn.githack.com/tripleblindmarket/safe-places/940edc081b8a4771deca806ed570eda907b5a261/healthcare-authorities.yaml) does not include a `bounds` field so by default there will be no authorities in your area)

## Comments / Concerns

- I [added a line to the `BackgroundTaskService`](https://github.com/tripleblindmarket/covid-safe-paths/pull/587/files#diff-5ba51018a0ebf079bfa63ee37b622f45R10) that will send a push notification to the user every 12 hours if there is an authority in their region that they haven't subscribed to.
  - Is 12 hours too frequent? I believe we can only have a single `BackgroundTaskService` (ours runs every 12 hours to check for intersections), so I'm not sure how else we could check at a pre-defined interval.
  - What if a user intentionally decides not to subscribe to a particular authority? Currently, they will continue to get a notification every 12 hours to subscribe.
- It would be nice if upon receiving a push notification, we could direct users to the `Choose health authority` screen. However, the only way to do that on iOS as far as I can tell is through the deprecated [`alertAction`](https://developer.apple.com/documentation/uikit/uilocalnotification/1616648-alertaction)  param.  I believe that we can do this for Android through the [`action` field for local notifications](https://github.com/zo0r/react-native-push-notification#local-notifications), but I'm not sure if we want that inconsistency between platforms.
- I'm not sure of a good way to test the push notification piece of this PR without changing some code around to force the notification and then revert the changes. I was thinking of doing a similar approach [that we use for intersections](https://github.com/tripleblindmarket/covid-safe-paths/pull/437/files#diff-5f57202af634a1db561d1c9ec795b37bR57-R70) but would like some feedback before implementing. @tstirrat - feedback here would be appreciated.
- Currently, the authorities list _does not_ contain bounds info. @penrods - how should we go about updating that file on [`raw.githack`](https://rawcdn.githack.com/tripleblindmarket/safe-places/940edc081b8a4771deca806ed570eda907b5a261/healthcare-authorities.yaml), and in general, is there a way that we're going to force authorities to include bounds? And in the case where a user uploads an authority with the `Add authority via URL` option, will we just allow them to not include bounds?

 ## Translation Notes ##
There are several new strings for translation.


penrods:
I've looked at the ChooseProvider.js screen and it works well!  Awesome work, @Patrick-Erichsen.  I tweaked the UI and behavior a little bit on this PR branch.

I haven't had time to verify the auto-subscription notification yet.

penrods:
Answering some of your questions:
* We can require bounding boxes for authorities and validate it when a new authority requests inclusion on that list via a PR.  I have already added tooling to the publisher and will added the bounding box for Haiti under the https://raw.githubusercontent.com/tripleblindmarket/safe-places/develop/healthcare-authorities.yaml.
* The manual addition of a healthcare authority capability does not require a GPS boundry, correct.  It is to allow a user to hook to a safe-paths.json source that for some reason is not in the YAML Global Registry yet.

NOTE: We need to switch to the CDN version of the YAML before we push this code to the app stores.

app/locales/en.json
```diff
@@ -6,6 +6,8 @@
     "authorities_desc": "Choose trusted healthcare authorities in your area to obtain exposure data. Either select a name from the global registry, or enter the web address provided by an authority which has implemented Safe Paths.",
     "authorities_input_placeholder": "Paste your URL here",
     "authorities_no_sources": "No data source yet",
+    "authorities_new_in_area": "New Healthcare Authority In Your Region",
+    "authorities_num_in_area": "Subscribe to {{numAuthories}} new trusted Healthcare Authorities in your location",
```
alpita-masurkar, line 10: typo in numAuthories

Patrick-Erichsen:
> please fix this: typo in numAuthories in en.json

Fixed - that same typo was copy/pasted in another file as well. Thanks for catching.

Patrick-Erichsen:
The linter is [failing](https://github.com/tripleblindmarket/covid-safe-paths/pull/587/checks?check_run_id=602017362) due to an error on [`no-mocks-import`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-mocks-import.md) from these two lines:

```
import * as mockHCA from '../__mocks__/mockHCA';
import {
  mockNullUserLocHistory,
  mockUserLocHistory,
} from '../__mocks__/mockUserLocHistory';
```

Anyone mind if I just ignore that rule for those lines? I think it's meant to prevent actual module mocks - this is just mock data.

kenpugsley:
> The linter is [failing](https://github.com/tripleblindmarket/covid-safe-paths/pull/587/checks?check_run_id=602017362) due to an error on [`no-mocks-import`](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-mocks-import.md) from these two lines:
> 
> ```
> import * as mockHCA from '../__mocks__/mockHCA';
> import {
>   mockNullUserLocHistory,
>   mockUserLocHistory,
> } from '../__mocks__/mockUserLocHistory';
> ```
> 
> Anyone mind if I just ignore that rule for those lines? I think it's meant to prevent actual module mocks - this is just mock data.

@tstirrat - thoughts on this?

kenpugsley:
Added `Do Not Merge (yet) as we need to hold off on merging this until the branch for v1 is complete.

tstirrat:
a rule override is totally fine given that you're not double importing a module mock

kenpugsley:
I think there's still some UI work to do on this.  Feel free to remove the `Do Not Merge (yet)` label when the UI is solid and we think we're in a solid place.

Patrick-Erichsen:
Based on the standup call today, I'm going to do a bit of refactoring on this:

- Create an "Auto-Subscribe" functionality and add it to the onboarding screens
- If a user is auto subscribed, the push notification will notify them of their new authorities rather than just notifying them that there are authorities in their area.

app/views/ChooseProvider.js
```diff
@@ -220,6 +194,18 @@ class ChooseProviderScreen extends Component {
     );
   }
 
+  async toggleFilterAuthoritesByGPSHistory() {
+    console.log('toggle!' + this.state.isAuthorityFilterActive);
```
tstirrat, line 198: remove test code

app/views/ChooseProvider.js
```diff
@@ -19,27 +17,22 @@ import {
...
-import { AUTHORITIES_LIST_URL } from '../constants/authorities';
-import colors from '../constants/colors';
 import Colors from '../constants/colors';
 import fontFamily from '../constants/fonts';
 import { AUTHORITY_SOURCE_SETTINGS } from '../constants/storage';
-import { GetStoreData, SetStoreData } from '../helpers/General';
+import { SetStoreData } from '../helpers/General';
 import { checkIntersect } from '../helpers/Intersect';
 import languages from '../locales/languages';
+import HCAService from '../services/HCAService';
```
tstirrat, line 32: prefer named exports because they work better with VSCode and other IDEs (they're more discoverable)

app/views/ChooseProvider.js
```diff
@@ -335,6 +321,19 @@ class ChooseProviderScreen extends Component {
             </TouchableOpacity>
           </MenuTrigger>
           <MenuOptions>
+            <TouchableOpacity
+              style={styles.authorityFilter}
+              onPress={() => this.toggleFilterAuthoritesByGPSHistory()}>
+              <Typography style={styles.authorityFilterText}>
```
tstirrat, line 327: Try to use an appropriate use="body1" or 2 here if it fits.  And remove inline styles if so.

Patrick-Erichsen: Updated it for this element and the other 7 `<Typography>` elements in the file.

app/views/ChooseProvider.js
Patrick-Erichsen: Done.

app/views/ChooseProvider.js
Patrick-Erichsen: Done.

Patrick-Erichsen:
I can't figure out for the life of me how to get these e2e tests to pass. The `notifications` permission doesn't seem to be working. Every step gets hung up on the notifications permissions, but no issues with location.

tstirrat:
> It would be nice if upon receiving a push notification, we could direct users to the Choose health authority screen. However, the only way to do that on iOS as far as I can tell is ...

Some contextual linking on a platform is better than none. So I'd be in favor of doing as much as possible here.

app/constants/storage.js
```diff
@@ -8,3 +8,4 @@ export const DEBUG_MODE = 'DEBUG_MODE';
 export const AUTHORITY_NEWS = 'AUTHORITY_NEWS';
 export const LAST_CHECKED = 'LAST_CHECKED';
 export const AUTHORITY_SOURCE_SETTINGS = 'AUTHORITY_SOURCE_SETTINGS';
+export const HCA_AUTO_SUBSCRIPTION = 'HCA_AUTO_SUBSCRIPTION';
```
tstirrat, line 11: If this is a boolean flag, make it `_ENABLED`

app/helpers/Intersect.js
```diff
@@ -275,10 +275,13 @@ export function checkIntersect() {
  * Returns the array of day bins (mostly for debugging purposes)
  */
 async function asyncCheckIntersect() {
-
   // first things first ... is it time to actually try the intersection?
-  let last_checked_ms =Number(await GetStoreData(LAST_CHECKED));
-  if (last_checked_ms + MIN_CHECK_INTERSECT_INTERVAL * 60 * 1000 > dayjs().valueOf()) return null;
+  let last_checked_ms = Number(await GetStoreData(LAST_CHECKED));
```
tstirrat, line 279: Since you're touching (even tho its a prettier format), lets standardize on camelCase vars

app/locales/en.json
```diff
@@ -6,11 +6,19 @@
     "authorities_desc": "Choose trusted healthcare authorities in your area to obtain exposure data. Either select a name from the global registry, or enter the web address provided by an authority which has implemented Safe Paths.",
     "authorities_input_placeholder": "Paste your URL here",
     "authorities_no_sources": "No data source yet",
+    "authorities_new_in_area_title": "New Healthcare Authority In Your Region",
```
tstirrat, line 9: Take screenhots of these new strings so we can upload to lokalise

app/locales/en.json
```diff
@@ -6,11 +6,19 @@
...
+    "authorities_new_in_area_msg": "Subscribe to {{numAuthorities}} new trusted Healthcare Authority in your location",
+    "authorities_new_in_area_msg_plural": "Subscribe to {{numAuthorities}} new trusted Healthcare Authorities in your location",
+    "authorities_new_subcription_title": "{{numAuthories}} New Subscription",
```
tstirrat, line 12: move these out of `label.x` and into a logical group. `<whatever you think>.x` 

label has become useless and a dumping ground

lokalise handles this nicely.

run `yarn i18n:extract` to diff en.json changes. make sure there are no extraneous keys here. but don't commit any empty strings.

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+   */
+  async getAuthoritiesList() {
+    let authorities = [];
+
+    try {
+      const result = await this.fetchAuthoritiesYaml();
+      const list = await RNFetchBlob.fs.readFile(result.path(), 'utf8');
+      authorities = Yaml.safeLoad(list).Authorities;
+    } catch (err) {
+      console.log(err);
```
tstirrat, line 42: maybe console.error

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+ */
+class HCAService {
+  /**
+   * Fetches the raw YAML file containing a list of all
+   * of the registered Health Care Authorities.
+   * Saves the response as a cached file for performance.
+   * @returns {void}
+   */
+  async fetchAuthoritiesYaml() {
+    return await RNFetchBlob.config({
```
tstirrat, line 25: omg if this can access local files, it's probably what I needed for eula.html reading!

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  /**
+   * Alerts a user that there are new Healthcare Authorities in their region.
+   * Includes information on the number of Authorities in their current location.
+   *
+   * @param {numAuthorities} Number number of new authorities
+   * @returns {void}
+   */
+  async pushAlertNewAuthoritesFromLoc(numAuthorities) {
```
tstirrat, line 87: async needed? if so await the localNotification

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+  /**
+   * Alerts a user that they have been subscribed to new Healthcare Authorities
+   * in their region. Includes information on the number of Authorities in
+   * their current location.
+   *
+   * @param {numAuthorities} Number number of new authorities
+   * @returns {void}
+   */
+  async pushAlertNewSubscribedAuthorities(numAuthorities) {
+    PushNotification.localNotification({
```
tstirrat, line 105: same here, no await.. async needed?

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+  getAuthorityUrl(authority) {
+    const authorityName = Object.keys(authority)[0];
+    const urlKey = authority[authorityName][0];
+    return urlKey && urlKey['url'];
+  }
+
+  /**
+   * Returns the `bounds` value for an authority
+   * @param {authority} Authority Healthcare Authority object
+   * @returns {Object} contains a `bounds` key.
```
tstirrat, line 129: `@returns {{bounds: {... structure}}}`

until we get TypeScript, this actually does work with the IDE

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+
+    return bounds && locHelper.isPointInBoundingBox(point, bounds);
+  }
+
+  /**
+   * Iterates over the full list of authorities and checks
+   * if there is any GPS point in the user's full 28-day location history
+   * that is within the bounds of the authority.
+   *
+   * @returns {Array} List of health care authorities
```
tstirrat, line 157: `@returns {Type[]}`

app/services/HCAService.js
```diff
@@ -0,0 +1,264 @@
...
+      locData.some(point => this.isPointInAuthorityBounds(point, authority)),
+    );
+  }
+
+  /**
+   * Gets the most recent location of the user and returns a list of
+   * all Healthcare Authorities whose bounds contain the user's current location,
+   * filtering out any Authorities the user has already subscribed to.
+   *
+   * @returns {Array} list of Healthcare Authorities
```
tstirrat, line 174: type the array here too

app/views/ChooseProvider.js
```diff
@@ -60,22 +56,11 @@ class ChooseProviderScreen extends Component {
...
-        this.setState({
-          selectedAuthorities: result,
-        });
-      } else {
-        console.log('No stored authority settings.');
-      }
-    });
+    await this.fetchAuthoritiesList(this.state.isAuthorityFilterActive);
+    await this.fetchUserAuthorities();
+    await this.fetchAutoSubcribeStatus();
```
tstirrat, line 63: this reads super well, thanks!

+100 for not lumping 4 layers of abstraction into this method.

app/views/ChooseProvider.js
```diff
@@ -215,33 +199,59 @@ class ChooseProviderScreen extends Component {
...
+    this.setState({ isAuthorityFilterActive: isAuthorityFilterActive.val });
+  }
+
+  async toggleAutoSubscribe() {
+    this.setState(
+      prevState => ({
+        isAutoSubscribed: !prevState.isAutoSubscribed,
+      }),
+      async () => {
+        this.state.isAutoSubscribed
```
tstirrat, line 219: not sure on this setState form, but is this.state guaranteed to be up to date? or does this function have a parameter with the latest state since update?

app/views/ChooseProvider.js
```diff
@@ -215,33 +199,59 @@ class ChooseProviderScreen extends Component {
...
+  }
+
   render() {
     return (
       <NavigationBarWrapper
         title={languages.t('label.choose_provider_title')}
         onBackPress={this.backToMain.bind(this)}>
         <View style={styles.main}>
-          <Typography style={styles.headerTitle}>
+          <Typography style={styles.headerTitle} use={Type.Headline2}>
```
tstirrat, line 232: This is cool but in future just `use="headline2"` is probably fine, in Typescript it would allow it.

app/views/onboarding/Onboarding5.js
```diff
@@ -1,72 +1,403 @@
...
-    this.props.navigation.replace('Onboarding6');
-  };
+      this.setState({
+        currentStep: nextStep,
+        authSubscriptionStatus: permission,
+      });
+    }
+  }
+
+  getLocationPermissionSetting() {
```
tstirrat, line 185: this feels like it should be in a service. the view doesn't care if iOS has a different permission string.

app/views/onboarding/Onboarding5.js
```diff
@@ -1,72 +1,403 @@
...
+      })
+      .catch(error => {
+        console.log('error checking location: ' + error);
+      });
+  }
+
+  checkNotificationStatus() {
+    const nextStep = this.getNextStep(StepEnum.NOTIFICATIONS);
+
+    checkNotifications().then(({ status }) => {
```
tstirrat, line 148: async/await

app/views/onboarding/Onboarding5.js
```diff
@@ -1,72 +1,403 @@
...
+        return StepEnum.DONE;
+    }
+  }
+
+  checkLocationStatus() {
+    // NEED TO TEST ON ANDROID
+    const nextStep = this.getNextStep(StepEnum.LOCATION);
+    const locationPermission = this.getLocationPermissionSetting();
+
+    check(locationPermission)
```
tstirrat, line 122: async/await

app/views/onboarding/Onboarding5.js
```diff
@@ -1,72 +1,403 @@
...
+      ? PERMISSIONS.IOS.LOCATION_ALWAYS
+      : PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION;
+  }
+
+  requestLocation() {
+    // NEED TO TEST ON ANDROID
+    const nextStep = this.getNextStep(StepEnum.LOCATION);
+    const locationPermission = this.getLocationPermissionSetting();
+
+    request(locationPermission).then(result => {
```
tstirrat, line 196: async await

e2e/InsufficientLocationPermissions.spec.js
```diff
@@ -1,8 +1,11 @@
+/* eslint-disable jest/expect-expect */
 import { navigateThroughOnboarding } from './helpers/onboarding';
 import FinishSetup from './pages/FinishSetup.po.js';
 
 describe('Location set to `inuse` and notifications `true` set', () => {
   beforeAll(async () => {
+    // const permissions = { location: 'inuse', notifications: 'YES' };
```
tstirrat, line 7: dead code, remove

e2e/NoAuthSubscriptionPermission.spec.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable jest/expect-expect */
+/* eslint-disable jest/expect-expect */
```
tstirrat, line 2: duplicate line

e2e/helpers/onboarding.js
```diff
@@ -1,3 +1,32 @@
+// import EnableAuthoritySubscription from '../pages/EnableAuthoritySubscription.po.js';
```
tstirrat, line 1: dead code, remove, add a `// TODO(#1234): Something something` if there is an outstanding test, raise an issue.

app/views/ChooseProvider.js
```diff
@@ -215,33 +199,59 @@ class ChooseProviderScreen extends Component {
...
             {languages.t('label.authorities_title')}
           </Typography>
-          <Typography style={styles.sectionDescription}>
+          <Typography style={styles.sectionDescription} use={Type.Body1}>
             {languages.t('label.authorities_desc')}
           </Typography>
+          <TouchableOpacity style={styles.autoSubcribe}>
+            <CheckBox
```
tstirrat, line 239: Sam introduced a basic Checkbox in common/... we should converge. if we need custom UX/styles, let's use the one in common and make that better...

Would accept this for now to get this over the line. But would like to know the plan for reuse.

app/services/HCAService.js
Patrick-Erichsen: So this was code I copy/pasted - but I believe it's taking the response and caching it as a file, not accessing that cached file.

https://github.com/joltup/rn-fetch-blob#download-to-storage-directly

app/services/HCAService.js
Patrick-Erichsen: Since these tasks are only triggered by BackgroundServices I don't think it makes sense to block on them. I'll remove the `async` - good catch.

app/views/ChooseProvider.js
```diff
@@ -356,7 +386,7 @@ class ChooseProviderScreen extends Component {
                   urlEntryInProgress: true,
                 });
               }}>
-              <Typography style={styles.menuOptionText}>
+              <Typography style={styles.menuOptionText} use={Type.Body2}>
```
tstirrat, line 389: This wasn't you, but please wrap this lower slide UI into a `SafeAreaView` so that it does not show too low down on the phone:

Screen Shot 2020-04-24 at 4 30 16 PM

tstirrat: Actually, out of scope. can fix after.

app/views/ChooseProvider.js
Patrick-Erichsen: Yeah so this is a little hacky but `setState` has an optional callback param that you can pass which will be executed once state is updated. 

https://stackoverflow.com/questions/43370176/using-async-setstate

I think the alternative would be to wrap the state update in a promise and then await that, e.g.

```js
asyncToggleSubscribe() {
  return new Promise((resolve) => this.setState(prevState => ({
        isAutoSubscribed: !prevState.isAutoSubscribed,
      })
}
```

app/locales/en.json
```diff
@@ -51,6 +61,9 @@
     "launch_notif_header": "Notifications will let you know if you cross paths with an infected person.",
     "launch_notif_subheader": "We won't bother you except to share updates on your potential exposure risks.",
     "launch_notification_access": "Allow notifications",
+    "launch_authority_header": "Healthcare Authorities will give us the local data to know if you cross paths with an infected person.",
+    "launch_authority_subheader": "Automatically subscribe to receive the latest updates from Healthcare Authorities in your area.",
+    "launch_authority_access": "Health authority subscription",
```
tstirrat, line 66: This is my main issue, it's ambiguous on the final screen:

Screen Shot 2020-04-24 at 4 16 06 PM

app/locales/en.json
Patrick-Erichsen: If it's alright with you, I'd like to just leave these as is and come back with another PR to re-organize this entire file. I think we should try to settle on a convention on what to break strings down by, e.g. file, functionality, etc.

app/locales/en.json
Patrick-Erichsen: Where should we be placing these screenshots? This is used in a push notification - so we want a screenshot of that notification to upload to Lokalise?

app/views/ChooseProvider.js
Patrick-Erichsen: Nice - didn't know we had that. I removed the package I pulled in and am using our custom Checkbox.

app/views/onboarding/Onboarding5.js
Patrick-Erichsen: Cool if we hold off on this for the larger body of work you mentioned in Slack?

app/services/HCAService.js
Patrick-Erichsen: Could use your feedback on the updates I made here.

app/locales/en.json
Patrick-Erichsen: I like you're proposed update: `Subscribe to nearby authorities`. I think that along with the longer subtext in that screen _should_ be enough context for a user. Maybe this is something, in particular, we can ask the testers to give us thoughts on?

app/views/ChooseProvider.js
tstirrat: All good no change needed. Although if you like making it Async await would be clearer

app/locales/en.json
tstirrat: Normally I'd say "yep sure" but I'm 50% sure this will create double work for 12 translators when we rename keys later.

Granted... There's a lot of key renaming warranted.... So we need a solution that does not interrupt translators... 

Ok.. seems fine to leave like this!

app/locales/en.json
tstirrat: Diarmid could probably weigh in ahead of time too

Patrick-Erichsen:
@tstirrat  I feel pretty solid at this point. Think it's ready to get in front of some testers ASAP.

kenpugsley:
I'll be doing a technical and functional review tonight.  I think we need to open the discussion on whether or not we've covered this well enough for inclusion in v1.0.1.

Patrick-Erichsen:
e2e tests are failing right now due to the missing Haitiain language strings. Once this is merged I'll open a new PR to only run the full suite of language tests on a merge to fix this issue.

tstirrat:
@kenpugsley we checked with Diarmid and there were no specific blockers getting this in. 

One desired enhancement after this, is to immediately auto sub to any nearby HCAs when choosing the auto option, but this requires a refactor of the location tracking stuff as we've mentioned elsewhere (a non trivial job).

tstirrat:
@Patrick-Erichsen I think the label is too long on the onboard 5 screen:

I fixed it so it wraps if too long, but was not quire sure what to do with the vertical height of the Onboarding5 screen.

if we can keep the checkbox line to a single line, it's not quite ideal but it would probably prevent much of this wrapping:

(galaxy nexus emulator)

Screen Shot 2020-04-26 at 11 00 43 PM

tstirrat:
Still a problem on nexus4:

Screen Shot 2020-04-26 at 11 00 43 PM

Even if I make it single line :(

Screen Shot 2020-04-26 at 11 08 15 PM

I suspect we could merge and fast follow with some kind of text hack that uses less text on small devices....  as long as you pretty pretty pinky swear to look into it straight after I'd still love to get this PR in to prevent as many merge conflicts etc.

Patrick-Erichsen:
Pushed up a fix to instead make the subheader text width `.8` instead of `.55`. Given how long the text was it felt like `.55` was too narrow.

Patrick-Erichsen:
@diar

tstirrat:
> Pushed up a fix to instead make the subheader text width `.8` instead of `.55`. Given how long the text was it felt like `.55` was too narrow.

If you make a visual change provide a screenshot so that I don't have to check out everything and rebuild just to see the text tweak you did :)

Patrick-Erichsen:
Screen Shot 2020-04-27 at 3 57 28 PM

@tstirrat

PR #592, opened 2020-04-20
[i18n] Upload strings to Lokalise, when merged into develop

tstirrat:
This should automate one part of the i18n flow to [lokalise.com](https://lokalise.com)

## How to test

I or @alpita-masurkar can test that new strings are uploaded.

See example of it working [here](https://github.com/tripleblindmarket/covid-safe-paths/pull/592/checks?check_run_id=603009861)


tstirrat:
Verified that these all upload ok. Changed it to upload merges into `develop` only.

PR #593, opened 2020-04-20
Adding translation to Brazilian Portuguese

MatheusAlvesAlmeida:
I had problems accessing the branch I created earlier, so I made another one and submitted the file with the requested changes. 

If you need to change anything else, you can comment and I will do it as soon as possible.


i18next-parser.config.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
     'id',
     'nl',
     'pl',
+    'pt-br',
```
tstirrat, line 51: pt_BR

app/locales/pt-br.json
```diff
@@ -0,0 +1,117 @@
+{
```
tstirrat, line 1: make this pt_BR.json

MatheusAlvesAlmeida: Done

PR #597, opened 2020-04-20
Set returnEmptyString to false

darkerthanblackoff:
## Description

PR sets i18next's returnEmptyString flag to false for avoid app crashes on passing empty strings. Resolves #593.

## How to test

Choose Nederalnds language on onboarding or in settings, open settings.


tstirrat:
Thanks! Just noticed this.

PR #601, opened 2020-04-21
issue#528 fixed 

Krish023:


app/views/LocationTracking.js
```diff
@@ -41,6 +42,7 @@ import languages from '../locales/languages';
...
...yP/g6w3BUW5ordEjPaiiFuRHdpZDM+jN5C9Cu39aENkd87vjFJMLz3XYuAS+DQ+ZSIQjYASVgvUJK0fsfjLLYrUddFPSVZH17Xxanc6AkpYLdiTsD7A4seR7dnIFhJ4k48Pk7CeFhARiE1AVwlbEGdQ2rSwp7coUsdFZyhZ1bHbfbRZe1ht+oG9rGUoYo+2aKbOHj3ssQ17zqxqD6a32Qq02AsB7WG16QkGp51gPgK5mROoTcihFlv7j1SyCoVXfrMQUMLKQIlBehfF/itD0SoX+W843FHlBqpt/gnokDBjH3Fo2IuidqvB8IyrVKnYnTRmBAnLxbvpqgRWbclHQAkrBy+S1loUvxfV6YHf6bR3GMlKVwVzbCsqGoaADglzcG0M2k+zis3IWAd7lUbuoWRVh64uRxuVsHL2E4P3IVbZF9kE/VU2uxK4L+2t6vsiq9x3lW2bElY3upZBfBOr7YVmdWP1MqxiyXhv2nljGYJVjPUhoHNYHfQ157R2YfVr0fIduPG2qt3JbsnqBm+BKR4RUMLqcBsgadnrsyxp2Qn5stsMGmBvS7mv7A1R/NUkoIRVQL+StNbGzRhU6GR8BYSWx8X9FB5FsrKrgjIRcElA57AK6BYG+VO42Radh+xVSWUyi9fmtvq4klWZuk2xikABBNjb2g1NR2Uwi3P3ApotFyIgAmUlQBLoh76H5iCPZnGdhTTFcVk3MsUtAkUTICFshK5EoV8bRhWZzOL4Lco1DXPRXORPBETAMQESxMbo1yjVHpfVa/WnmE7Zcc8oNBEQgS4JkDAGoJPQVBTDrB6rb0CXQWmBCJSIgG5rSNRZJJFNqXpkQ/ZCid4FhDIPH39BY9EYrvrpsZoCoMqFHwJKWA76guRlc8bbPVzDkM1sOrChNflcDi1p9uiM3eT5REM2E6jNIjGBJGVv15GJQCUJKGE571aSmc3D1Rf1Qfbs4uskJc1LBQiZCIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIiACIhACQn8P8rOOhTDbYOjAAAAAElFTkSuQmCC"
```
kenpugsley, line 45: Doesn't really belong here  ... at a minimum should be over in the images directory.  Is it not possible to do the svg render on iOS 12.0?

app/views/LocationTracking.js
```diff
@@ -291,6 +293,14 @@ class LocationTracking extends Component {
     }
     return BackgroundImage;
   }
```
kenpugsley, line 295: My understanding was that this was only an issue on iOS12.0.  I know in my testing on iOS 11 it wasn't a problem.  Please verify the specific iOS version needed.

package.json
```diff
@@ -43,6 +43,7 @@
     "@react-native-community/async-storage": "^1.8.1",
     "@react-native-community/geolocation": "^2.0.2",
     "@react-native-community/masked-view": "0.1.5",
+    "@react-native-community/netinfo": "^5.7.0",
```
kenpugsley, line 46: what's the reason for adding netinfo?

yarn.lock
```diff
@@ -1246,6 +1246,11 @@
   resolved "https://registry.yarnpkg.com/@react-native-community/masked-view/-/masked-view-0.1.5.tgz#25421be6cd943a4b1660b62cfbcd45be8891462c"
   integrity sha512-Lj1DzfCmW0f4HnmHtEuX8Yy2f7cnUA8r5KGGUuDDGtQt1so6QJkKeUmsnLo2zYDtsF8due6hvIL06Vdo5xxuLQ==
 
+"@react-native-community/netinfo@^5.7.0":
```
kenpugsley, line 1249: same comment as above

app/views/LocationTracking.js
Krish023: Yeah. As I have commented on the issue 528. It is issue with svg library which we are using. So I have used data instead of svg xml.

app/views/LocationTracking.js
Krish023: Okay, I will try on ios 11. I haven't tried on ios 11. If it is an issue with ios12.0 only then I will change the condition.

package.json
Krish023: It is for network check. I have added another issue which was 256 which was not merged. I will remove this also.

Krish023: Removed

yarn.lock
Krish023: Removed.Please verify it

app/views/LocationTracking.js
Krish023: I have checked the condition for 12.0 and 12.0.1 only. Please verify it.

app/views/LocationTracking.js
Krish023: I have moved image data to separate files inside assets. Please verify it.

Krish023:
I have changed based on the above comment. Could you please verify and merge it?

app/views/LocationTracking.js
```diff
@@ -291,6 +293,15 @@ class LocationTracking extends Component {
...
+  getiOSVersionEqualto12() {
```
tstirrat, line 296: This can be a module scope function. It does not use `this`

.eslintcache
```diff
@@ -0,0 +1 @@
+[{"/Users/krishna.jha1/Desktop/krishnajha/Covid-19/4/covid-safe-paths/app/assets/svgs/settingsGearData.js":"1","/Users/krishna.jha1/Desktop/krishnajha/Covid-19/4/covid-safe-paths/app/views/LocationTracking.js":"2"},{"size":11965,"mtime":1587449693293,"results":"3","hashOfConfig":"4"},{"size":17373,"mtime":1587449693294,"results":"5","hashOfConfig":"4"},{"filePath":"6","messages":"7","errorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},"156yab6",{"filePath":"8","messages":"9","errorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},"/Users/krishna.jha1/Desktop/krishnajha/Covid-19/4/covid-safe-paths/app/assets/svgs/settingsGearData.js",[],"/Users/krishna.jha1/Desktop/krishnajha/Covid-19/4/covid-safe-paths/app/views/LocationTracking.js",[]]
```
tstirrat, line 1: Remove this cache file

Krish023:
Hi, @tstirrat Thanks for your suggestions. It works for the settings button. I have used stoke="none".  But it is not working for the back button. All other files reset based on your comments. could you please verify once again. I have changed only on settingsGear.js.
Desc:
I have observed, It is an issue with only the back button and settings image because of its use of image data. All other svg files not used image data so it is working fine. If I missed any svg file like this then please let me know. I will look at the issue and will create another PR for the along with the back button issue.

tstirrat:
Oh I see. the SVG items use image data inside them..!? wow. Is this what breaks ios 12.0?

wow.

Your diff has too much in it again. I cannot see the change clearly. Please fix

Krish023:
Hi @tstirrat , Yeah Image data giving problem for iOS 12.0. I can't see other file changes. I have changed only on the SettingsGear.js file. I have created a new PR (issue 528-fixed#issue528 #643). Can you please verify it? If still, you see too many file changes, please discard it except SettingsGear.js.

Issue #629, opened 2020-04-22
Missing documentation for adding background-geolocation environment variable 


samikhalildev:
## Version

- Type of phone: Android 

## Steps to reproduce
I realised this was missing in the documentation when I was trying to run the application for android.

### Configure background-geolocation SDK
Create a file named ```local.properties``` in /android directory with an environment variable pointing to the android SDK:
* For windows add: ```sdk.dir = C:\\Users\\USERNAME\\AppData\\Local\\Android\\sdk```
* For macOS add: ```sdk.dir = /Users/USERNAME/Library/Android/sdk```
* For linux add: ```sdk.dir = /home/USERNAME/Android/Sdk```

summetj:
Linked to #630, closed.

PR #630, opened 2020-04-22
Added notes for background-geolocation SDK environment variable

samikhalildev:
## Description
Add documentation for configuring background-geolocation SDK  #629 

## How to test
npx react-native run-android


kenpugsley:
Sorry... not sure I understand.  I do not do these steps.  What exactly do these changes do?

tstirrat:
Hi, on linux/mac the dev_setup.sh script sets `ANDROID_SDK_ROOT=` to the correct value. So this should not be necessary.

Do you have this env var defined? On linux/mac there is a `~/.profile_mobileapp` file which configures it. this should be sourced into your profile.

Which platform are you on?

tstirrat:
Will close for now, since this is already covered in `dev_setup.sh` on Mac/Linux

If you think there needs to be more clarity around `source ~/.profile_mobileapp`, then please reopen with some updates. Thanks!

Issue #603, opened 2020-04-21
Unknown status (due to location permission disabled) disappears on clicking on Settings


tremblerz:
## Version

- Model: Pixel 3 API 29

## Steps to reproduce

Install the app, onboard like a nice person and then go to settings, disable the location permissions to the App. Go back to the App, you will see "Unknown status" and then tap on settings, dashboard appears which shows Location active (**Shouldn't be**) and you go back it shows standard "No known contact"

## Expected Behavior

Tapping on settings shouldn't have changed any state and settings drawer should have shown "Location inactive"

shilpashendre:
I can start working on this issue.

summetj:
https://github.com/tripleblindmarket/covid-safe-paths/pull/631

PR #631, opened 2020-04-22
Set proper location status indicator in settings page

shilpashendre:
## Description

location status was showing incorrect in settings page, if user makes location permission off from app setting , which is fixed now.


summetj:
https://github.com/tripleblindmarket/covid-safe-paths/issues/603

app/views/Settings.js
```diff
@@ -53,11 +57,49 @@ export const SettingsScreen = ({ navigation }) => {
...
+        LocationServices.stop();
+        setBGLocationLogging(false);
+        setIsLogging(false);
+        BackgroundTaskServices.stop();
+      }
+    });
+  };
+
+  // this is called when you click on device permission dialog "DENY" & "ALLOW"{ /*tested properly on android device*/}
+  const checkBackgroundStatus = async () => {
```
tstirrat, line 85: This isn't actually **checking** the background status, its responding to the event, which may fire at any time. Perhaps this should go in the initial `useEffect()`

app/views/Settings.js
```diff
@@ -53,11 +57,49 @@ export const SettingsScreen = ({ navigation }) => {
...
+    });
+  };
+
   const locationToggleButtonPressed = async () => {
     try {
       isLogging ? LocationServices.stop() : LocationServices.start();
       await SetStoreData(PARTICIPATE, !isLogging);
       setIsLogging(!isLogging);
+      setBGLocationLogging(true);
+      checkBackgroundStatus();
```
tstirrat, line 102: This check is not really a check, so it's not worth doing this each time the button is pressed. Keep it once in the initial `useEffect()`

app/views/Settings.js
```diff
@@ -81,11 +122,11 @@ export const SettingsScreen = ({ navigation }) => {
         <Section>
           <Item
             label={
-              isLogging
+              isLogging && isBGLocationLogging
```
tstirrat, line 125: these two variables seem to track the same thing. So they could do with different names.

perhaps: `isUserParticipating && isBgLocationLogging`

app/views/Settings.js
```diff
@@ -53,11 +57,49 @@ export const SettingsScreen = ({ navigation }) => {
...
+      //BroadcastingServices.start();
+    });
+    // Check and see if they actually authorized in the system dialog.
+    // If not, stop services and set the state to !isLogging
+    // Fixes tripleblindmarket/private-kit#129
+    BackgroundGeolocation.checkStatus(({ authorization }) => {
+      if (authorization === BackgroundGeolocation.AUTHORIZED) {
+        LocationServices.start();
+        setBGLocationLogging(true);
+        setIsLogging(true);
```
tstirrat, line 74: It looks like you are trying to track these as two separate booleans:

- `isLogging` = user PARTICIPATE state
- `isBgLogging` = actual bg service state

if that's the case, then don't set isLogging to true here, because now you've muddled the state between the two variables.

shilpashendre:
Thank you for your response. I have made changes in code with explainable comments. Please check it.

app/views/Settings.js
```diff
@@ -52,13 +54,42 @@ export const SettingsScreen = ({ navigation }) => {
     return () => {
       BackHandler.removeEventListener('hardwareBackPress', handleBackPress);
     };
-  }, [navigation]);
+  }, []);
+
+  //Initialyy when settings page is loaded , this method will check whether we have remove location permission from app settings.
+  // It will set status "Location Active" only if permission is granted.
+  const checkIsLocationEnable = () => {
```
tstirrat, line 61: All these checks are now encoded in `LocationService.checkStatus()`: 

https://github.com/Path-Check/covid-safe-paths/blob/develop/app/services/LocationService.js#L347-L382

It was added in #749 

Please use that, it has tests. And the views do not need to access BackgroundGeolocation directly anymore.

app/views/Settings.js
```diff
@@ -1,44 +1,46 @@
...
 import xmarkIcon from '../assets/svgs/xmarkIcon';
 import { Divider } from '../components/Divider';
-import { Feature } from '../components/Feature';
 import NativePicker from '../components/NativePicker';
 import NavigationBarWrapper from '../components/NavigationBarWrapper';
 import Colors from '../constants/colors';
 import { PARTICIPATE } from '../constants/storage';
 import { GetStoreData, SetStoreData } from '../helpers/General';
-import {
+import languages, {
```
tstirrat, line 15: Don't use languages.t anymore, use `const {t} = useTranslation()`

app/views/Settings.js
```diff
@@ -52,13 +54,42 @@ export const SettingsScreen = ({ navigation }) => {
     return () => {
       BackHandler.removeEventListener('hardwareBackPress', handleBackPress);
     };
-  }, [navigation]);
+  }, []);
```
tstirrat, line 57: this revert could cause state issues.. please keep the previous [navigation]

Install the [vs code eslint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) so you can see warnings about this in your editor.

app/views/Settings.js
```diff
@@ -1,44 +1,46 @@
...
   supportedDeviceLanguageOrEnglish,
 } from '../locales/languages';
+import BackgroundTaskServices from '../services/BackgroundTaskService';
 import LocationServices from '../services/LocationService';
 import { GoogleMapsImport } from './Settings/GoogleMapsImport';
 import { SettingsItem as Item } from './Settings/SettingsItem';
 
 export const SettingsScreen = ({ navigation }) => {
-  const { t } = useTranslation();
+  const backToMain = () => {
```
tstirrat, line 27: You've reverted some other changes.. please sync and undo any changes that are not part of this change.

app/views/Settings.js
```diff
@@ -52,13 +54,42 @@ export const SettingsScreen = ({ navigation }) => {
...
+      } else if (authorization === BackgroundGeolocation.NOT_AUTHORIZED) {
+        LocationServices.stop();
+        setIsLogging(false);
+        BackgroundTaskServices.stop();
+      }
+    });
+  };
+
+  //  This method is used to track whether user has "Allow" or "Deny" location permission from system Dialog.
+  const SystemDialogPermission = async () => {
```
tstirrat, line 75: methods should be `camelCase()`

It's unclear whether this is supposed to request permissions, or check the status of permissions?

app/views/Settings.js
tstirrat: Once you use the new checkStatus, this code should become more clear.

app/views/Settings.js
```diff
@@ -73,18 +104,17 @@ export const SettingsScreen = ({ navigation }) => {
...
+      title={languages.t('label.settings_title')}
       onBackPress={backToMain}>
       <ScrollView>
         <Section>
           <Item
             label={
               isLogging
-                ? t('label.logging_active')
-                : t('label.logging_inactive')
+                ? languages.t('label.logging_active')
```
tstirrat, line 116: For now it's probably ok to show "Location Inactive" for any of the "canTrack: false" states:

```js
const {canTrack, reason} = await LocationServices.checkStatus();

const isLogging = canTrack;
```

Here are the states btw:

- user toggled off inside the app
  ```js
  reason == Reason.USER_OFF
  ```
- permissions are not granted
  ```js
  reason == Reason.NOT_AUTHORIZED
  ```
- location services is turned off
  ```js
   reason == Reason.LOCATION_OFF
   ```

Eventually we would like to differentiate between these states.. but it's **ok to stick with `setIsLogging(canTrack)`** until the UX team gives guidance.

Feel free to reach out to them on slack if you want to incorporate more specific messaging here.

PR #642, opened 2020-04-23
[i18n] Don't translate terms_of_use_url

tstirrat:
Remove the `terms_of_use_url` from i18n, and move to functional + hooks + Typography

## How to test

Check screenshot, looks the same:

Screen Shot 2020-04-22 at 7 32 46 PM


tstirrat:
> LGTM - the only comment is that I feel like it would be nice to have that URL as an env variable or something so that we can change it easily if need be. Especially after all of the back-and-forth we've had during EULA things.

Yeah, it sounds like we'll need language settings payloads somewhere re mayo clinic too. But configuring the app in lokalise seems weird too.

Issue #432, opened 2020-04-11
Repeated notification for same intersection and redundant intersection checks


tremblerz:
The way we are checking for intersection is based on storing data and checking it periodically and sending a notification if an intersection is found. Now the issue here is that the intersection which happened 5 days ago will get flagged again when it is found on the next day as an intersection which happened 6 days ago.

We need to find a way to get rid of the duplicate notifications in the time domain. This, in general, also brings the point that right now we check for whole intersection every 12 hours which might not be a good idea, especially when the public json files would be larger.

harshvitra:
One possible solution is that when we check for intersection **we store the timestamp in the Local Storage**

If there is no timestamp - check entire public json
If there is timestamp - we only check values higher than the timestamp

PR #644, opened 2020-04-23
Record where user has crossed paths to avoid repeated notifications

samikhalildev:
## Description
This fixes the issue where notification is fired on every intersection. I have created a seperate store to add the user's intersections and help avoid redundant checks and notifications. #432 

The intersections store is indexed via the user’s location time where the contact happened. The inner object represents the infected cases.
```
1588132102236: {
      1588132101236: {
        "latitude": -33.849211, "longitude": 150.906321
      },
      1588156321426: {
        "latitude": 37.33182281, "longitude": -122.03038342
      }
},
```
When ```asyncCheckIntersect()``` runs, we keep track of the exposure period of all new intersections, only when that counter is within a set threshold, the notification is fired.

## How to test
1. Record locations with exposure period for 10 minutes or longer
2. Monitor notification pop up
3. Reload app


summetj:
@tremblerz  - samikhalildev is working to clear the auto-check and then this will be ready for your review.

app/helpers/Intersect.js
```diff
@@ -118,6 +119,24 @@ export function intersectSetIntoBins(
...
         dayBins[daysAgo] += exposurePeriod;
 
+        // Identify if this is a new intersection by checking the crossedPaths property of the current location against the infected location j
+        if (
+          !('crossedPaths' in currentLocation) ||
+          !currentLocation['crossedPaths'].some(
+            crossedPaths =>
+              crossedPaths.time == concernArray[j].time &&
+              crossedPaths.latitude == concernArray[j].latitude &&
+              crossedPaths.longitude == concernArray[j].longitude,
```
tstirrat: This is a good candidate for extracting a well named method:

```js
if (!hasExistingIntersection(currentLocation, concernArray[j])) {

}
```

app/helpers/Intersect.js
```diff
@@ -118,6 +119,24 @@ export function intersectSetIntoBins(
...
+          )
+        ) {
+          newIntersections = addNewIntersection(
+            localArray,
```
tstirrat: You don't need the `i` inside the method, so this could take `currentLocation` as param 1

app/helpers/Intersect.js
```diff
@@ -402,3 +432,21 @@ export function disableDebugMode() {
...
+ * @param {number} newIntersections - Counter of new intersections found
+ */
+function addNewIntersection(localArray, i, infectedLocation, newIntersections) {
+  console.log('New intersection');
+
+  if ('crossedPaths' in localArray[i])
+    localArray[i]['crossedPaths'].push(infectedLocation);
+  else localArray[i]['crossedPaths'] = [infectedLocation];
+
+  return newIntersections + 1;
```
tstirrat: This method shouldn't know about the count. It always adds 1, so the count is irrelevant. You could leave count in the parent method.

app/helpers/Intersect.js
```diff
@@ -202,11 +224,15 @@ export function normalizeAndSortLocations(arr) {
...
       if ('time' in elem && 'latitude' in elem && 'longitude' in elem) {
-        result.push({
+        let data = {
           time: Number(elem.time),
           latitude: Number(elem.latitude),
           longitude: Number(elem.longitude),
-        });
+        };
+
+        if ('crossedPaths' in elem) data.crossedPaths = elem.crossedPaths;
```
tstirrat: This is scary, copying over crossedPaths which might be mutated elsewhere.
clone the array:

```js
let data = {
  // ..
  crossedPaths: elem.crossedPaths && [...elem.crossedPaths],
}
```

app/helpers/Intersect.js
```diff
@@ -127,7 +146,10 @@ export function intersectSetIntoBins(
     }
   }
 
-  return dayBins;
+  // Updates 'crossedPaths' field within location data if a user has crossed paths with infected case
+  if (newIntersections > 0) SetStoreData(LOCATION_DATA, localArray);
```
tstirrat: Why this side effect here? It seems like we'd write localArray at the very end end when all iterations are done right?

Could this have a problem where it overwrites previous valid crossedPaths with empty crossedPaths because we're only halfway through the process?

app/helpers/Intersect.js
```diff
@@ -402,3 +432,21 @@ export function disableDebugMode() {
...
+ * @param {array} localArray - Array of user locations
+ * @param {number} i - Index of current location in the localArray
+ * @param {array} infectedLocation - Array of infected locations
+ * @param {number} newIntersections - Counter of new intersections found
+ */
+function addNewIntersection(localArray, i, infectedLocation, newIntersections) {
+  console.log('New intersection');
+
+  if ('crossedPaths' in localArray[i])
+    localArray[i]['crossedPaths'].push(infectedLocation);
```
tstirrat: Mutating an object that you don't own (didn't create) is asking for trouble. 

I'd really like to avoid mutating `localArray`, but if that was the intended behaviour by @kenpugsley that localArray would progressively be mutated, I'll allow it. @kenpugsley  can you please weigh-in?

It can often end up in hard to debug state issue later.

app/helpers/__tests__/Intersect.spec.js
```diff
@@ -126,7 +126,7 @@ describe('intersect with empty sets', () => {
     baseLocations = normalizeAndSortLocations(baseLocations);
     concernLocations = normalizeAndSortLocations(concernLocations);
 
-    let resultBins = intersectSetIntoBins(baseLocations, concernLocations);
+    let resultBins = intersectSetIntoBins(baseLocations, concernLocations).bins;
```
tstirrat: definitely need thorough tests on the new behaviour please.

app/helpers/Intersect.js
```diff
@@ -402,3 +432,21 @@ export function disableDebugMode() {
...
+/**
+ * Records where user has crossed paths to avoid repeated notifications or checking
+ *
+ * @param {array} localArray - Array of user locations
+ * @param {number} i - Index of current location in the localArray
+ * @param {array} infectedLocation - Array of infected locations
+ * @param {number} newIntersections - Counter of new intersections found
+ */
+function addNewIntersection(localArray, i, infectedLocation, newIntersections) {
+  console.log('New intersection');
```
tstirrat: remove console logs

app/helpers/Intersect.js
```diff
@@ -301,18 +311,29 @@ async function asyncCheckIntersect() {
...
         });
 
         // intersect the users location with the locations from the authority
-        let tempDayBin = intersectSetIntoBins(
+        let { bins, newIntersections } = intersectSetIntoBins(
           locationArray,
           normalizeAndSortLocations(responseJson.concern_points),
         );
 
+        // Updates crossedPaths field within location data if a user has crossed paths with a new infected case
```
kenpugsley: I think we can't do this ...  at least not this way.  It will cause a race condition with the LocationService, and has a non-zero chance of causing us to overwrite data - specifically if a new gps point came in while the intersection was running.  We'd end up overwriting with data that doesn't included the newly saved GPS point, and while it's probably a small percentage, it's not zero and we can't do it.

I think the answer for you to set up a separate store of the intersection results - basically a parallel store, probably keyed by the time that will store the intersection results.  Since it's a separate store it won't cause the race condition.

app/helpers/Intersect.js
```diff
@@ -332,8 +353,8 @@ async function asyncCheckIntersect() {
   // Store the news arary for the authorities found.
   SetStoreData(AUTHORITY_NEWS, name_news);
 
-  // if any of the bins are > 0, tell the user
-  if (dayBins.some(a => a > 0)) notifyUserOfRisk();
+  // if there are any recent interaction, notify user once, no need for multiple notifications at once (i think)
+  if (newIntersectionsCount > 0) notifyUserOfRisk();
```
kenpugsley: I think it would be better to have a constant for a minimum delta in the intersection data (probably an amount of exposure time) that is required for a new notification.  The idea would be that a really small change we wouldn't notify you about, but if there was a significant increase we would.  I'm thinking on the order of 10 minutes of extra exposure to cause a notification - though that may get debated, so having it set as a constant (in units of time) would be a big help if we need to adjust it.

app/helpers/Intersect.js
```diff
@@ -415,3 +436,32 @@ export function disableDebugMode() {
   // Kick off intersection calculations to restore data
   checkIntersect();
 }
+
```
kenpugsley: See my notes about mutating the current location data, and the need to store intersections in a parallel store.  Will probably affect the next two functions.

app/helpers/Intersect.js
```diff
@@ -30,6 +30,8 @@ import languages from '../locales/languages';
  * Intersects the locationArray with the concernLocationArray, returning the results
  *   as a dayBin array.
  *
+ * NOTE: localArray items can be mutated on each run
```
kenpugsley: Not a good idea to mutate the localArray ... it's asking for future problems with someone comes back in later and tries to refactor or reuse these methods.  Better to make your own copy.  Plus, see my comments on how you'll need to deal with the intersection results storage - I think you'll have to rework it anyway.

app/helpers/Intersect.js
```diff
@@ -119,6 +122,12 @@ export function intersectSetIntoBins(
         // now add the exposure period to the bin
         dayBins[daysAgo] += exposurePeriod;
 
+        // Identify if this is a new intersection
+        if (!hasExistingIntersection(currentLocation, concernArray[j])) {
+          addNewIntersection(currentLocation, concernArray[j]);
+          newIntersections += 1;
```
kenpugsley: Probably better to do this in units of time (e.g. ` += exposurePeriod`).  That way we can notify if only if the extra intersection time is over some threshold.

app/helpers/Intersect.js
tstirrat: Another point here is that someone is working on the geolocation store to make it encrypted. That means writing back will change. So, I think it's a good idea to keep this new structure separate so that those incoming changes don't invalidate this work. Thanks again for adjusting to our needs!

samikhalildev:
@kenpugsley @tstirrat Please review my update PR.

ios/Podfile.lock
```diff
@@ -499,4 +499,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: fe7a0c609d647e21562a9b114f89414b0377d154
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.7.5
```
tstirrat: remove this change

ios/Podfile.lock
```diff
@@ -334,7 +334,7 @@ DEPENDENCIES:
   - Yoga (from `../node_modules/react-native/ReactCommon/yoga`)
 
 SPEC REPOS:
-  trunk:
+  https://github.com/cocoapods/specs.git:
```
tstirrat: What does this do?

app/helpers/Intersect.js
```diff
@@ -415,3 +451,43 @@ export function disableDebugMode() {
...
+ */
+function hasExistingIntersection(
+  intersections,
+  currentLocation,
+  concernLocation,
+) {
+  let { time } = currentLocation;
+
+  return (
+    time in intersections &&
```
tstirrat: Just go with `intersections[time] && `

app/helpers/Intersect.js
```diff
@@ -30,15 +32,19 @@ import languages from '../locales/languages';
...
  * @param {array} concernArray - superset array of all concerning points from health authorities.  Assumed to have been sorted and normalized
+ * @param {object} intersections - The user intersections history
  * @param {int} numDayBins - (optional) number of bins in the array returned
  * @param {int} concernTimeWindowMS - (optional) window of time to use when determining an exposure
  * @param {int} defaultExposurePeriodMS - (optional) the default exposure period to use when necessary when an exposure is found
  */
 export function intersectSetIntoBins(
   localArray,
   concernArray,
+  intersections = {},
```
tstirrat: `previousIntersections` or `knownIntersections`

app/helpers/Intersect.js
```diff
@@ -128,7 +149,7 @@ export function intersectSetIntoBins(
     }
   }
 
-  return dayBins;
+  return { bins: dayBins, intersections, newIntersections };
```
tstirrat: you're returning intersections here but never using the return value.. remove it from the return data.

newIntersections could be confused with an array .. its a count so call it `newIntersectionCount`

app/helpers/Intersect.js
```diff
@@ -203,11 +224,14 @@ export function normalizeAndSortLocations(arr) {
     for (let i = 0; i < arr.length; i++) {
       let elem = arr[i];
       if ('time' in elem && 'latitude' in elem && 'longitude' in elem) {
-        result.push({
+        let data = {
           time: Number(elem.time),
           latitude: Number(elem.latitude),
           longitude: Number(elem.longitude),
-        });
+          crossedPaths: elem.crossedPaths ? [...elem.crossedPaths] : [],
```
tstirrat: This can be reverted now, right?

app/helpers/Intersect.js
```diff
@@ -300,6 +325,7 @@ async function asyncCheckIntersect() {
     for (const authority of authority_list) {
       try {
         let responseJson = await retrieveUrlAsJson(authority.url);
+        let intersections = await GetStoreData(INTERSECTIONS_DATA, false);
```
tstirrat: Add a comment warning that this **might be mutated.** Maybe it even needs a clear name like `updatedIntersections`

app/helpers/Intersect.js
```diff
@@ -308,11 +334,20 @@ async function asyncCheckIntersect() {
...
+        if (newIntersections > 0) {
+          await SetStoreData(INTERSECTIONS_DATA, intersections);
+          newIntersectionsCount += newIntersections;
+        }
+
+        let tempDayBin = bins;
+
         // Update each day's bin with the result from the intersection.  To keep the
         //  difference between no data (==-1) and exposure data (>=0), there
         //  are a total of 3 cases to consider.
```
tstirrat: This wasnt your code but can you extract this sum method below on line 354 into a separate method?

There's a lot to process here and it would read better if its something like `dayBins = sumBins(dayBins, bins)`

app/helpers/Intersect.js
```diff
@@ -415,3 +451,43 @@ export function disableDebugMode() {
...
+function hasExistingIntersection(
+  intersections,
+  currentLocation,
+  concernLocation,
+) {
+  let { time } = currentLocation;
+
+  return (
+    time in intersections &&
+    intersections[time].some(
```
tstirrat: This could become a performance issue if there are lots of prior intersections at the same time value... @kenpugsley can you weigh in here? is that likely?

app/constants/storage.js
```diff
@@ -1,4 +1,5 @@
 export const LOCATION_DATA = 'LOCATION_DATA';
+export const INTERSECTIONS_DATA = 'INTERSECTIONS_DATA';
```
tstirrat: Needs a better name. this is the list of already notified intersections, maybe `KNOWN_INTERSECTIONS`

app/helpers/Intersect.js
```diff
@@ -415,3 +451,43 @@ export function disableDebugMode() {
...
+/**
+ * Records where user has crossed paths to avoid repeated notifications or checking
+ * @param {object} intersections - The user intersections history
+ * @param {object} currentLocation - The user's location
+ * @param {object} concernLocation - The infected location
+ */
+function addNewIntersection(intersections, currentLocation, infectedLocation) {
+  let { time } = currentLocation;
+
+  if (time in intersections) {
```
tstirrat: `intersections[time]` is fine here.

ios/Podfile.lock
samikhalildev: Apologies, I should of not committed this file

samikhalildev:
@tstirrat  Thanks for the feedback! :)

app/helpers/Intersect.js
samikhalildev: I was thinking of implementing a binary search similar to the one being used to find the minimum time for concern locations. The known intersections will need to be sorted however as the time can be unordered. Let me know what you think. @tstirrat @kenpugsley @tremblerz

samikhalildev: Code has been updated. Instead of storing a list of intersections at a given time, we could store the infected location time as the key to look up the location data (longitude and latitude). To check if an intersection has been identified, it would be as simple as this:
```
return (
    knownIntersections[currentLocationTime] &&
    knownIntersections[currentLocationTime][concernLocationTime]
  )
```

summetj:
@samikhalildev I think this is waiting on you to resolve the conflicts.

PR #648, opened 2020-04-23
Fix incorrect "plural" keys, add key checks for new PRs

tstirrat:
- Cleans up some incorrect plural keys.
- Moves some keys to `share.*` and `import.google.*`
- Add a check for missing keys. [Example failure](https://github.com/tripleblindmarket/covid-safe-paths/pull/648/checks?check_run_id=612872599)

## How to test

Check that the strings show in the app. Check that the renamed keys are matched with old values in Lokalise


app/views/Export.js
```diff
@@ -129,27 +130,39 @@ export const ExportScreen = ({ navigation }) => {
...
-              <Typography style={styles.exportSectionTitles}>
-                {t('label.tested_positive_title')}
-              </Typography>
-              <Typography style={styles.exportSectionPara}>
-                {t('label.export_para_1')}
-              </Typography>
-              <Typography style={styles.exportSectionPara}>
-                {t('label.export_para_2')}
+              <Typography use='headline2' style={styles.exportSectionTitles}>
+                {t('share.title')}
```
Patrick-Erichsen, line 134: Should this be 

```jsx
<Trans i18nKey='share.title>Share location history</Trans>
```

app/views/Export.js
```diff
@@ -129,27 +130,39 @@ export const ExportScreen = ({ navigation }) => {
...
+                >
+                  Location is shared as a simple list of times and places, no
+                  additional information.
+                </Typography>
+              </Trans>
 
               <TouchableOpacity style={styles.exportButton} onPress={onShare}>
                 <Typography style={styles.exportButtonText}>
-                  {t('label.share_location_data')}
+                  {t('share.button_text')}
```
Patrick-Erichsen, line 157: Should this be 

```jsx
<Trans i18nKey='share.title>Share location data</Trans>
```

tstirrat: I back tracked on `<Trans>` because we can end up in a situation where the react code can says `<Trans>Hey cool here's some text</Trans>`, but through some legal or other reason we have to change the english text, and this will read from the `en.json` file, and now there's a weird disconnect in the code and UI, especially for running unit tests.

What do you think?

Patrick-Erichsen: Yeah, that's a good point. Personally, when poking around the codebase it's a bit of a pain to have to keep `en.json` open and reference back to that all the time too though. 

Either way you are probably right about pulling back on the use of `<Trans>` though - I can live with a bit of a nuisance in development over a bug in prod

app/components/NativePicker.js
```diff
@@ -104,7 +104,7 @@ export default class NativePicker extends Component {
                         marginRight: 22,
                       }}
                       onPress={() => this.setState({ modalVisible: false })}>
-                      {languages.t('Done')}
+                      <Trans i18nKey='common.done'>Done</Trans>
```
Patrick-Erichsen, line 107: Based on our convo lets this to just 

```js
{t('common.done')}
```

app/views/Export.js
```diff
@@ -129,27 +130,39 @@ export const ExportScreen = ({ navigation }) => {
...
-              </Typography>
-              <Typography style={styles.exportSectionPara}>
-                {t('label.export_para_1')}
-              </Typography>
-              <Typography style={styles.exportSectionPara}>
-                {t('label.export_para_2')}
+              <Typography use='headline2' style={styles.exportSectionTitles}>
+                {t('share.title')}
               </Typography>
+              <Trans i18nKey='share.paragraphs'>
```
Patrick-Erichsen, line 136: Based on our convo lets remove this `<Trans>` and use a `t()` function in the Typography elements below.

app/views/__tests__/__snapshots__/Import.spec.js.snap
```diff
@@ -161,14 +161,15 @@ Array [
             }
             use="body1"
           >
-            Visit Google Takeout and export your Location History using following settings: 
+            Visit Google Takeout and export your Location History using the following settings: 
 1. Delivery method: "Add to Drive" 
 2. Frequency: "Export once" 
-3. File type & size: ".zip" and "1GB" 
+3. File type &amp; size: ".zip" and "1GB"
```
Patrick-Erichsen, line 167: I was able to confirm that this is showing up as `&amp;` instead of `&` in the app. Looks like the bug is on [this key](https://github.com/tripleblindmarket/covid-safe-paths/pull/648/files#diff-95aa5b48e9899a303ae08d4353789356R18) in en.json

.github/workflows/i18n.yml
```diff
@@ -1,23 +1,23 @@
...
+    branches: [develop]
+    paths:
+      - app/locales/en.json
+      - ios/en.lproj/*.strings
+      - android/app/src/main/res/values/strings.xml
 
 jobs:
+  # Only run on a final merge into develop, uploads all changed keys to Lokalised
   lokalise-upload:
+    if: github.event_name == 'push' && github.ref == 'develop'
```
lfalkner, line 20: 😍

app/locales/en.json
```diff
@@ -1,4 +1,29 @@
 {
+  "common": {
```
lfalkner, line 7: so much more readable!

app/views/Import.js
```diff
@@ -86,15 +90,15 @@ const ImportScreen = props => {
             }
             style={styles.buttonTouchable}>
             <Typography style={styles.buttonText}>
-              {languages.t('label.import_takeout').toUpperCase()}
+              {languages.t('import.google.visit_button_text')}
```
lfalkner, line 93: Was this intentionally made not capitalized

app/views/Import.js
```diff
@@ -86,15 +90,15 @@ const ImportScreen = props => {
...
             </Typography>
           </TouchableOpacity>
           <TouchableOpacity
             testID='google-takeout-import-btn'
             onPress={importPickFile}
             style={styles.buttonTouchable}>
             <Typography style={styles.buttonText}>
-              {languages.t('label.import_title').toUpperCase()}
+              {languages.t('import.title')}
```
lfalkner, line 101: Same as above

app/views/Import.js
tstirrat: Yep, because our buttons styles should be determined at the button level not at the consumer level.

app/views/Import.js
tstirrat: See above, it's not in the spec to have caps buttons

app/views/Import.js
tstirrat: Screen Shot 2020-04-27 at 4 00 58 PM

Issue #652, opened 2020-04-23
Enable app force upgrade in v1


elviraux:
on IOS and Android

Issue #40, opened 2020-03-14
FEATURE: Verify application freshness


penrods:
Add a mechanism to remind the user to update if they have turned off auto-updates on their phone.

The first version can be a simple notification when the app is outdated.

tushar-dadlani:
A library that can be used. I'm new to react native, I'll try this out

Do we have CI to test iOS and android? 

https://github.com/kimxogus/react-native-version-check

harshvitra:
@tushar-dadlani  

Hey did you get to try the library ? 
The functions are capable of achieving what we need

snmishra:
I've used [codepush](https://github.com/Microsoft/react-native-code-push) in one of my previous apps. It's a heavier change requiring changes to the release process. But it does allow updating the javascript bundle without updating the entire app. It turned out to be less useful in the initial releases where the native code was also changing fast. Still, I was able to fix some crashes quickly without having to wait for the user to update the app.

PR #653, opened 2020-04-23
Add version check service

Fatkhi:
## Description

Adds upgrade notification functionality.

Fixes https://github.com/tripleblindmarket/covid-safe-paths/issues/652, https://github.com/tripleblindmarket/covid-safe-paths/issues/40

Currently, version checker relies on [static version file](https://github.com/Fatkhi/covid-safe-paths/blob/releases/versions.json)

It makes fetch requests as soon as a user opens the app and then every hour in the background. If the app's current version (from package.json) is less than `version` specified in json file, a simple alert will appear with an ability to delay the upgrade:

Simulator Screen Shot - iPhone 11 - 2020-04-23 at 16 14 33

If the app's current version (from package.json) is less than `mandatory_version` specified in json file, alert without the ability to delay + push notification will appear:

Simulator Screen Shot - iPhone 11 - 2020-04-23 at 16 16 16

The above is useful for critical updates.

### TODO:

- [ ] Add translations for new texts: Push title, message, alert label, sublabel, later and update buttons
- [x] Add correct links to AppStore and Playstore 
- [x] Change location of `versions.json` file, or switch to other solution (need to discuss)
- [x] Fix typo in `Application is outdate` :)

## How to test

Change your `package.json` to 
- 1.0.0 to see no notifications
- 0.9.4 to see only gentle alert
- 0.9.1 for mandatory update

### DIFF
--- a/app/constants/urls.js
+++ b/app/constants/urls.js
@@ -1,0 +1,6 @@
+// need to get actual links to stores
+export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1501903733';
+export const PLAYSTORE_URL = 'market://details?id=edu.mit.privatekit';
+// placeholder json for testing, to be changed
+export const VERSION_URL =
+  'https://raw.githubusercontent.com/Fatkhi/covid-safe-paths/releases/versions.json';
--- a/app/constants/urls.js
+++ b/app/constants/urls.js
@@ -1,6 +1,6 @@
 // need to get actual links to stores
-export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1501903733';
-export const PLAYSTORE_URL = 'market://details?id=edu.mit.privatekit';
+export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1508266966';
+export const PLAYSTORE_URL = 'market://details?id=org.pathcheck.covidsafepaths';
 // placeholder json for testing, to be changed
 export const VERSION_URL =
   'https://raw.githubusercontent.com/Fatkhi/covid-safe-paths/releases/versions.json';
--- a/app/constants/urls.js
+++ b/app/constants/urls.js
@@ -1,6 +1,0 @@
-// need to get actual links to stores
-export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1508266966';
-export const PLAYSTORE_URL = 'market://details?id=org.pathcheck.covidsafepaths';
-// placeholder json for testing, to be changed
-export const VERSION_URL =
-  'https://raw.githubusercontent.com/Fatkhi/covid-safe-paths/releases/versions.json';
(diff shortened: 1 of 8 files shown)


app/constants/urls.js
```diff
@@ -0,0 +1,6 @@
+// need to get actual links to stores
+export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1501903733';
```
tremblerz, line 2: The id for safepaths app is 1508266966

app/constants/urls.js
```diff
@@ -0,0 +1,6 @@
...
+export const PLAYSTORE_URL = 'market://details?id=edu.mit.privatekit';
```
tremblerz, line 3: for playstore id is gonna be `org.pathcheck.covidsafepaths`

app/locales/en.json
```diff
@@ -109,5 +109,13 @@
     "what_does_this_mean": "What does this mean?",
     "what_if_no_symptoms_para": "If you have no symptoms but still would like to be tested you can go to your nearest testing site.\n\nIndividuals who don't exhibit symptoms can sometimes still carry the infection and infect others. Being careful about social distancing and coming in contact with large groups or at risk individuals (the elderly, those with significant other medical issues) is important to manage both your risk and the risk to others.",
     "what_if_no_symptoms": "What if I’m not showing symptoms?"
+  },
+  "version_update": {
+    "push_notification_title": "Application is outdate",
```
Patrick-Erichsen, line 114:
```suggestion
    "push_notification_title": "Application is out of date",
```

app/constants/urls.js
```diff
@@ -0,0 +1,6 @@
...
+// placeholder json for testing, to be changed
+export const VERSION_URL =
+  'https://raw.githubusercontent.com/Fatkhi/covid-safe-paths/releases/versions.json';
```
tremblerz, line 6: Can you add a separate PR for this JSON file to be included in the repository? so that we test here directly

kenpugsley: @penrods suggested that we should use the CDN via https://raw.githack.com/.  He's apparently talked with them about this use case.  Related ... might need to do the same for the healthcare-authority.yaml file.  Need him to verify.

Fatkhi:
Will update link to yaml file after https://github.com/tripleblindmarket/covid-safe-paths/pull/658 is merged

app/constants/versionCheck.js
```diff
@@ -0,0 +1,6 @@
+export const APPSTORE_URL = 'itms-apps://itunes.apple.com/us/app/id1508266966';
+export const PLAYSTORE_URL = 'market://details?id=org.pathcheck.covidsafepaths';
+export const VERSION_URL =
+  'https://rawcdn.githack.com/tripleblindmarket/covid-safe-paths/0008e269c355e14bb216bb6e69cf4b89e25012b2/versions.yaml';
```
kenpugsley, line 4: Please change to `https://rawcdn.githack.com/tripleblindmarket/covid-safe-paths/master/versions.yaml`.  This will still use the CDN with longer caching, but won't include a tag or commit hash

PR #654, opened 2020-04-23
Temporarily filter out extra languages in release mode

tstirrat:
Filters out extra languages for the first release, until we have EULAs in place.

## How to test

Check language list in dev mode

Check language list in non-dev mode with CMD+m -> settings -> disable Dev mode -> restart react native

Screen Shot 2020-04-23 at 4 16 53 PM

### DIFF
--- a/app/locales/languages.js
+++ b/app/locales/languages.js
@@ -72,6 +72,25 @@ export async function setUserLocaleOverride(locale) {
   await SetStoreData(LANG_OVERRIDE, locale);
 }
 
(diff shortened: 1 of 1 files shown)


tstirrat:
> Looks like this will change the language order (arabic will be #3). I think that's just fine TBH. Will do a final verify when we build for release.

I can sort the list. one sec

tstirrat:
The list is sorted now

Screen Shot 2020-04-23 at 4 27 04 PM

PR #666, opened 2020-04-24
Polish the EULA

penrods:
Several changes to make the EULA palatable:
* Reduced the indents in the HTML with inline CSS
* Reduced text size so it is small like legal text should be ;-)
* Reducing some padding around elements to fit small Android 6 screen
* Removed the scroll to botton -- injected javascript was not running,
  or at least not generating any calls to handleWebViewMessage()

 ### Testing Notes ###
Install, look at the glorious EULA!

 ### Tech Notes ###
Under the Android 6 emulator I never got any messages when I placed
a console.log("here!") in the handleWebViewMessage().  I'm assuming
the webview didn't support injecting JS back then.

### DIFF
--- a/app/components/EulaModal.js
+++ b/app/components/EulaModal.js
@@ -39,7 +39,7 @@ const webViewScrollDetection = `
   };
 
   function scrollHandler() {
-    if (window.innerHeight + window.scrollY >= document.body.offsetHeight) {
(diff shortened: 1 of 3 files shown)


penrods:
New look:
image

penrods:
Previously it looked like this in an Android 6 emulator:
image

Issue #618, opened 2020-04-21
[Choose health authority] - View title not fully visible 


ilonadynska:
## Version - 0.9.5

- Type of phone: Android 
- OS Version: 9.0, 10.0
- Model: Samsung A50, Samsung Galaxy S10

## Steps to reproduce
1. From home tap the right top setting button 
2. Tap on Choose health authority 

Applies to following languages:
- Spanish
- French
- Italian
- Dutch
- Polish
- Russian

## Expected Behavior

Title should be fully visible. 

![Screenshot_20200422-002247_COVID Safe Paths](https://user-images.githubusercontent.com/59801345/79924518-3f1b2500-8438-11ea-8447-65c5525d6bbe.jpg)

ilonadynska:
@tstirrat can we make the font smaller or fit the space (or truncate) or we should involve each translator to modify the text and (if possible) make it shorter?

NethravathiPuttraju:
I can look into this. @tstirrat  can you give your suggestions .

NethravathiPuttraju:
Can anyone give suggestion which solution will be better , reducing the font or truncate the text..

tstirrat:
@alpita-masurkar and @penrods will probably have thoughts here. Thanks for checking

Patrick-Erichsen:
Closing as we have UI/UX work in flight to refactor this screen.

PR #670, opened 2020-04-24
Fit_text_InsideView

NethravathiPuttraju:
Resolves issue #618 ,
Fit Title text of all languages inside Header View . 
Screenshot_title_InsideView_COVID Safe Paths


tstirrat:
Hello, this increases the height of the nav bar everywhere. I don't think we want to do that.

If you still want to help solve this you could come on slack and talk to the UX folks about how to handle these long strings here.

driedgrape:
We're in the process of a redesign of the "select a health authority" process, so let's punt on this issue for the time-being, but I'll make sure it's considered and included in the redesign.

Issue #675, opened 2020-04-24
Navigation issues in V1.0 updated setup sequence.


diarmidmackenzie:
## Version

V1.0.0
- Android 9
- Moto G7

## Steps to reproduce

Multiple navigation issues in the startup flow.

By fast the worst is when you get to the HA screen.  Having entered your HA, you then have to go "back" to go forward.  And the back button is in the top left hand corner, which doesn't match any of the navigation patterns the user has seen up to that point.

Vedha didn;t spot that back button, so tried the Android back button, and at this stage, that just closes the app.

I think most users will fail to get past this stage without closing the app themselves or hitting the back button.

There are a number of other navigation issues around this initial flow.  None so critical, but in combination they do make things even worse.
- You can't go back on any of the opening screens.  That was OK before as they were mostly just messages, with no config, but now e.g. the location screen is directly after tha HAs.  Totally legitimate that you might want to go back & tweak your HA settings, but there is no back button.  Android back button closes the app.
- If you do close the app (whether using Android back button, or by other means), there are some odd things in the flow:
- You have to accept the Icense agreement again, which feels odd.
- You *don't* have to enter an HA again, if you entered it already.  But the prompt tells you that you have to.  There is no adjustment to the text to reflect the fact that you have done thisstep already - I guess the design didn't anticipate the user having to come bak a 2nd time.  So the text says: "... you will need to subscribe to a health authority" even tough you already have.
- When you have not yet entered a Health Authority, I would suggest the "Next" button be labelled "Later" (or "Skip").
- It is inconsistent that you can skip HA setup, but not skip location setup (even though you can turn location off again in-app later!).

## Expected Behavior

Expectations are:
- A seamless setup experience, where I keep moving forward and don't have ot move back
- If I do need to move back, the back controls should be clerly visible and the ANdroid Back button should also work.
- HA & Location screens should be consistent with regard to navigation (can I go back?) and whether I can skip.
- If I end up repeating after closing the app, we should have a consistent paradigm on whether config /choices need to be re-entered.  My preference should be that choices doon't need to be re-entered, but I value consistency most of all.

diarmidmackenzie:
Some screenshots.

Here is the HA selection screen.  The back button is in the top left.  This made sense previously when you had navigated to the screen via the dashboard link in the top right of the app, but in this new context it makes no sense at all.

[...]

PR #679, opened 2020-04-24
Remove authority selection from onboarding

Patrick-Erichsen:

## Description 

Removes the Healthcare Authority selection from onboarding - requested by our testers after finding the current flow to be confusing.

This is breaking the e2e tests but with the impending authority auto subscription work I think it's okay for now.

#### Linked issues:

#675 

#### How to test

Go through onboarding in both iOS and Android and confirm there is not Healthcare step and that the process in general still works correctly.


tstirrat:
Tested on emulators: pixel2 and nexus 4

PR #681, opened 2020-04-24
yarn run-ios now runs pod install

tstirrat:
tested locallly, worked fine

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "yarn && react-native start",
     "run-android": "react-native run-android",
-    "run-ios": "react-native run-ios",
+    "run-ios": "cd ios && pod install && cd .. && react-native run-ios",
     "preinstall": "node -e \"if(process.env.npm_execpath.indexOf('yarn') === -1) throw new Error('You must use Yarn to install, not NPM')\"",
     "install:pod": "cd ios && bundle install && bundle exec pod install",
     "postinstall": "patch-package; npx react-native-jetifier",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "yarn && react-native start",
     "run-android": "react-native run-android",
-    "run-ios": "react-native run-ios",
+    "run-ios": "cd ios && pod install && cd .. && react-native run-ios",
```
kenpugsley, line 8: Wouldn't it be slightly better to do the same as the **install:pod**?  
e.g. `cd ios && bundle install && bundle exec pod install && cd .. && react-native run-ios`

tstirrat: Done, switched to using the other script

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "yarn && react-native start",
     "run-android": "react-native run-android",
-    "run-ios": "cd ios && pod install && cd .. && react-native run-ios",
+    "run-ios": "yarn install:pod && react-native run-ios",
```
kenpugsley, line 8: are we assuming everyone will use yarn?

kenpugsley: sorry ... I meant that to be a comment, not request changes.  If realistically everyone must use yarn, then I think this is OK.  I'm just not sure.

tstirrat: Oh I see. yes we enforce yarn for this repo (over npm). so they wont get far without it

PR #688, opened 2020-04-26
Add darkened background on Mayo Clinic link

tstirrat:
- Makes it much more clear where the touchable part is
- Prevents it from being in the "unsafe bezel area" on later model iPhone models.
- Removes the `onPress` for each inner text element, so that the entire "button" area highlights when touched.

#### Linked issues:

None

#### Screenshots:

Screen Shot 2020-04-25 at 9 10 21 PM
Screen Shot 2020-04-25 at 9 11 27 PM
Screen Shot 2020-04-25 at 9 13 43 PM

### DIFF
--- a/app/views/LocationTracking.js
+++ b/app/views/LocationTracking.js
@@ -11,6 +11,7 @@ import {
   StatusBar,
   StyleSheet,
   Text,
+  TouchableHighlight,
   TouchableOpacity,
   View,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


tstirrat:
> On the smallest screens this completely covers the exposure history link, and effectively makes it unusable. This is a Nexus One screenshot ...

Is that a new bug? it seems like this change would not have introduced that bug. I suspect it was always unclickable. Can you confirm?

tstirrat:
> > On the smallest screens this completely covers the exposure history link, and effectively makes it unusable. This is a Nexus One screenshot ...
> 
> Is that a new bug? it seems like this change would not have introduced that bug. I suspect it was always unclickable. Can you confirm?

Yep, existing issue, the mayo link covers the button.

Screen Shot 2020-04-28 at 10 54 27 AM

kenpugsley:
Yes, it's true that it overlapped before, but at least if you did a little sideways dance and touched just right you could hit the button.  With your change there's no way to activate the button on really small screens.

tstirrat:
> Yes, it's true that it overlapped before, but at least if you did a little sideways dance and touched just right you could hit the button. With your change there's no way to activate the button on really small screens.

Oh I see, bummer. Closing for now.

Issue #677, opened 2020-04-24
In Haitian Creole, some of the Exposure Data screen is still in English


diarmidmackenzie:
## Version

V1.0.0
Android 9
Moto G7

## Steps to reproduce

Generate a detected exposure (e.g. https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/21430273/23+April+2019+-+My+first+risk+of+infection+v0.9.5)

Observe the English words.

Specifically the following are all in English:
- Months
- Days od the week
- Time units (e.g. minutes).

## Expected Behavior

I expect these things to be in the local language.

## Upload any files to the Issue useful in helping us to investigate

![image](https://user-images.githubusercontent.com/16045703/80233689-318fb600-8657-11ea-9f83-eed94aac1351.png)

Issue #689, opened 2020-04-26
Add End to End tests for Haitian Creole on iOS


efalkner:
Update Detox test framework to execute our tests against a simulator localized to Haitian Creole.

Issue #522, opened 2020-04-15
Top priority translations (Haitian Kweyol and Vietnamese)


mediajunkie:
@kenpugsley:

(from Khahlil)
> I've attached two files here with the app language translated to local languages - Kweyol and Vietnamese. Can we prioritize these to have them available as soon as the app is ready? 
> 
> This is especially for the Haitian team who have the beta download already and are testing. There is a current Kweyol version available, but updates have been made to the language, which is what's contained in this file. 
> 
> Let me know if you need support on this. Thanks

[fwdpriorityapptranslation.zip](https://github.com/tripleblindmarket/covid-safe-paths/files/4481938/fwdpriorityapptranslation.zip)

"done by EOW if possible" per Sam Z.

PR #690, opened 2020-04-26
Feature/creole tests

efalkner:
#### Linked issues:

https://github.com/tripleblindmarket/covid-safe-paths/issues/689
https://github.com/tripleblindmarket/covid-safe-paths/issues/677
https://github.com/tripleblindmarket/covid-safe-paths/issues/522

#### Screenshots:
These are generated by CI via the tests running.

#### How to test
Run end to end tests locally or in CI. Expect each test to run against English and Haitian Creole localized app.


summetj:
@efalkner   FYI - If you want to "auto-link" the issues to this PR, you must have the keyword (fixes/closes) in front of each issue number separately, so you need to say:
fixes #698
fixes #677
instead of just listing the numbers.
(I went ahead and manually linked those two issues in the sidebar for you).

efalkner:
Thanks @summetj . This really is only related to one of the issues (and encapsulates the other issue). I'm not sure if we have a way to auto link using 'relates to', but that would be useful, I think

summetj:
> Thanks @summetj . This really is only related to one of the issues (and encapsulates the other issue). I'm not sure if we have a way to auto link using 'relates to', but that would be useful, I think

Got it,  I've removed #677 from the linked issues so that only #689 will be auto-closed when this is merged.

PR #692, opened 2020-04-26
Fix Android NavigationBarWrapper display jump bug (seen when exiting Settings screen)

StartupSam:
## Description

As the title says, this PR is fixing a small display bug where exiting the Settings screen caused the display to jump noticeably. It was caused by the StatusBar translucent property being iOS specific for some reason. Fixed by making the status bar consistently take the same amount of space, then added some extra padding to NavigationBarWrapper to make up for the loss on Android.

#### Screenshots:

Android with padding:
Screenshot from 2020-04-26 07-21-54

#### How to test

Please take a look to make sure this acts as expected on high res Android devices and devices with notches.

It shouldn't affect iOS display at all, but a double check would be great.

### DIFF
--- a/app/components/NavigationBarWrapper.js
+++ b/app/components/NavigationBarWrapper.js
@@ -28,8 +28,8 @@ const NavigationBarWrapper = ({ children, title, onBackPress }) => {
     <>
       <StatusBar
         barStyle='light-content'
-        backgroundColor={barColor}
-        translucent={isPlatformiOS()}
+        backgroundColor={isPlatformiOS() ? barColor : 'transparent'}
+        translucent
       />
       <TopContainer />
       <BottomContainer>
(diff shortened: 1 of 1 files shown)


app/components/NavigationBarWrapper.js
```diff
@@ -63,12 +63,21 @@ const BottomContainer = styled.SafeAreaView`
...
+      background-color: ${themeNavBar};
+      border-bottom-color: ${themeNavBarBorder};
+      border-bottom-width: 1px;
+      flex-direction: row;
+    `
+  : styled.View`
+      background-color: ${themeNavBar};
+      border-bottom-color: ${themeNavBarBorder};
+      border-bottom-width: 1px;
+      padding-top: 10px;
```
tstirrat, line 78: With styled components you inline the function which takes props as the first arg:

`padding-top: ${props => isPlatformiOS() ? 0 : 10}`

tstirrat: Also this needs to be 24px to match the previous spacing:

Screen Shot 2020-04-26 at 9 24 37 AM


I suggest inserting a spacing View, rather than top padding for Android, otherwise it makes the title text center into the Title bar div too high:

Screen Shot 2020-04-26 at 9 22 25 AM


And when you do this fix it might push the settings icon down on the main screen.. cant quite tell.

tstirrat:
pixel 3 xl (notch) doesn't work :( (this is with 24px padding)

I wonder if it's the SafeAreaView not working in the emulator.

Screen Shot 2020-04-26 at 9 32 40 AM

tstirrat:
Adding an extra SaveAreaView helped:

```js
<TopContainer />
      {!isPlatformiOS() && (
        <SafeAreaView style={{ backgroundColor: barColor }} />
      )}
```

Screen Shot 2020-04-26 at 10 23 32 AM


But then of course, we cant have nice things:

Screen Shot 2020-04-26 at 10 32 08 AM

StartupSam:
Does absolute positioning still respect safeAreaView? Asking since the header text uses absolute positioning

tstirrat:
> Does absolute positioning still respect safeAreaView? 

it will probably ignore padding because it positions into the parent container's max bounds, so I imagine it will ignore safeareaview stuff.

tstirrat:
Closing this due to no movement, reopen if you make any progress.

PR #696, opened 2020-04-27
webview issue on news

### LINKED ISSUE #615: News does not load up on the first time sometimes
## Version

- Model: iPhone 6s iOS 13.4

## Steps to reproduce

Reproducing is a bit inconsistent but it has happened sufficient number of times to assume the bug does exist.
Install the app, go to the latest news
It just remains stuck on loading
Press back and go to the latest news again
This time it loads up right away

## Expected Behavior

The News should get loaded in the first pass and shouldn't require going back and coming into it again.

saranyaharidas:
go to the latest news
It just remains stuck on loading
Press back and go to the latest news again
This time it loads up right away

News will be loading properly
#### Linked issues:

Fixes #615 
#### How to test
It just remains stuck on loading
Press back and go to the latest news again
This time it loads up right away

### DIFF
--- a/app/views/News.js
+++ b/app/views/News.js
@@ -55,7 +55,8 @@ class NewsScreen extends Component {
 
   renderItem = item => {
     return (
-      <View style={styles.singleNews}>
+      // <View style={styles.singleNews}>
+      <>
         <View style={styles.singleNewsHead}>
           <Typography style={styles.singleNewsHeadText}>
             {item.item.name}
@@ -69,6 +70,7 @@ class NewsScreen extends Component {
             borderBottomLeftRadius: 12,
             borderBottomRightRadius: 12,
           }}
+          style={{ flex: 1, width: '100%' }}
           cacheEnabled
           onLoad={() =>
             this.setState({
@@ -76,7 +78,8 @@ class NewsScreen extends Component {
             })
           }
         />
-      </View>
+      </>
+      //  </View>
     );
   };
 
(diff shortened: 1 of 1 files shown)


tstirrat:
Hello, thanks is this the same as #704 ?

saranyaharidas:
Looks like these are different issues

app/views/News.js
```diff
@@ -55,7 +55,8 @@ class NewsScreen extends Component {
 
   renderItem = item => {
     return (
-      <View style={styles.singleNews}>
+      // <View style={styles.singleNews}>
```
tstirrat, line 58: remove dead code

remove the unused style block below

However, I think you want `alignItems: 'stretch'` here.

app/views/News.js
```diff
@@ -69,14 +70,16 @@ class NewsScreen extends Component {
             borderBottomLeftRadius: 12,
             borderBottomRightRadius: 12,
           }}
+          style={{ flex: 1, width: '100%' }}
```
tstirrat, line 73: You're placing `width: 100%` on all children. It looks like you want `align-items: stretch` on the parent, rather than `width: 100%` on all children.

app/views/News.js
```diff
@@ -168,10 +171,14 @@ const styles = StyleSheet.create({
   },
   singleNewsHead: {
     height: 48,
+    backgroundColor: 'rgba(255,255,255,0.6)',
+    borderTopLeftRadius: 12,
```
tstirrat, line 175: Since you are changing visuals, can you provide screenshots?

saranyaharidas: Simulator Screen Shot - iPhone 8 - 2020-04-30 at 14 00 07

app/views/News.js
saranyaharidas: these are changed now

saranyaharidas:
After doing some more research on this, I think `   <Carousel
               data={this.state.newsUrls}
                renderItem={this.renderItem}
                sliderWidth={width}
                itemWidth={width * 0.85}
                layout={'default'}
                scrollEnabled
              />` component is causing this issue. When we render webview with flex:1 directly issue is not happening . Do not merge my above commits.

PR #697, opened 2020-04-27
Fallback to English translations in e2e tests

tstirrat:
Fixes e2e when there are test on new strings. It will fall back to English for any missing keys, which matches the i18n lib's functionality

#### How to test

e2e should pass

### DIFF
--- a/e2e/helpers/language.js
+++ b/e2e/helpers/language.js
@@ -1,4 +1,4 @@
-import assign from 'lodash/assign';
+import merge from 'lodash/merge';
 
 import * as english from '../../app/locales/en.json';
 import * as haitian from '../../app/locales/ht.json';
@@ -9,7 +9,7 @@ import * as haitian from '../../app/locales/ht.json';
  */
 const languageStrings = {
   'en-US': english,
-  'ht-HT': assign({}, english, haitian),
+  'ht-HT': merge({}, english, haitian),
 };
 
 export const languages = Object.keys(languageStrings).map(locale => [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,6 +57,7 @@
     "emotion-theming": "^10.0.27",
     "i18next": "^19.3.3",
     "js-yaml": "^3.13.1",
+    "lodash": "^4.17.15",
```
efalkner, line 60: wow, I can't believe we didn't have this yet

e2e/helpers/language.js
```diff
@@ -1,9 +1,15 @@
...
 import * as haitian from '../../app/locales/ht.json';
 
+/**
+ * Language strings for each locale. Strings are merged with english as a base
+ * to emulate i18next's english fallback in the case of missing locale strings.
+ */
 const languageStrings = {
   'en-US': english,
-  'ht-HT': haitian,
+  'ht-HT': merge({}, english, haitian),
```
Patrick-Erichsen, line 12: How would you feel about moving this `merge` logic down into the map function? e.g.

```js
export const languages = Object.keys(languageStrings).map(locale => [
  locale,
  merge({}, english, languageStrings[locale]),
]);
```

tstirrat: Way better, will do

package.json
tstirrat: Yeah, it was only a transitive dep before.

Issue #685, opened 2020-04-25
when use location is turned off in settings safe paths does not behave correctly.


aitkeni:
## Version  built off v 1.0.0 (24) commit 2b95fd40ec5c421c361b8eb1419ace8743926619

Android Pixel 2 API 28

## Steps to reproduce
open app
background app , 
open settings and disable use location
go back to app. 

Issue 1 notice location tracking still green/ enabled
expected - red / disabled

Issue 2 
Click green icon to disable message is displayed.
click on that message , the message is displayed again
keep clicking

expected - click on message would not trigger same message again

aitkeni:
Noticed that there was some commented out code in LocationService.js#L227, L315 When I uncommented these blocks the situation improved, specifically when app detects that use location is disabled it will present an Alert box allowing the user to open the settings and turn it on.  This then resolves issue 1.
When I went thru the history it seems this code was commented out during the merge to the new UI design.  What I am thinking is it was accidentally left commented out, or there may have been another reason which would need worked on. Can you confirm if you remember @kenpugsley .  If accidental, I can un-comment and get it all working again.
also can i ask what is the best way to refresh the settings page - when i change the state to active, sometimes need to force a refresh.

ShreyaaSahu:
iPhone XR 
ios version 13.3 
App version 1.0.1

This issue is happening in ios App also .

Niniya173:
Same issue is seen in
 iPhone 5S
ios version 11.2.6
App version 1.0.1

aitkeni:
thanks for confirming guys!  I have some code which tries to fix this issue and reenables the feature to open the location settings so user can turn it on.  just ironing out some scenarios where it gets confused about its state.  will make a PR today hopefully.

tstirrat:
Issue 1 fixed in #749 

Issue 2 tracked in #656

PR #725, opened 2020-04-29
Improve status checks of background location service

aitkeni:
#### Linked issues:

fixes issue 1 in #685 
when for the check for backgound service is still running, it was getting it wrong.  
and the fact that the variable isBackgroundGeolocationConfigured doesnt actually reflect the correct state of the BackgroundGeolocation in certain cases
also reinstates the code to open an Alert box allowing the user to open the settings and turn on location services.
also includes fix for issues #564 which was merged in

#### Screenshots:
image

#### How to test
for issue #685 part1
open app
background app ,
open settings and disable use location
go back to app.

Used to be - location tracking icon still green/ enabled
With fix will correctly show - red / disabled
App will detect this and pop up an Alert to allow the user to open the settings and turn on location services.

issue 564 see PR #574 for how to test

##### TODO 
localize resourse changes / additions
i think the settings app Alert opens twice sometimes.  Its doesnt break anything, its just a bit of a glitch
tests ... please advise!


app/locales/en.json
```diff
@@ -100,23 +100,34 @@
     "launch_set_up_phone": "Set up my phone",
     "legal_page_title": "Legal",
     "loading_public_data": "loading data...",
-    "location_disabled_message": "COVID Safe Paths requires location services.",
-    "location_disabled_title": "Location Tracking Was Disabled",
-    "location_enabled_message": "COVID Safe Paths is securely storing your GPS coordinates once every five minutes on this device.",
-    "location_enabled_title": "COVID Safe Paths Enabled",
+    "location_disabled_message": "Please enable location access for Safe Paths.",
+    "location_disabled_title": "Device location disabled",
+    "location_enabled_message": "Safe Paths is anonymously recording your location.",
```
Patrick-Erichsen, line 105: @tstirrat  thoughts on this language string change?

app/services/LocationService.js
```diff
@@ -212,10 +213,15 @@ export default class LocationServices {
     const locationData = new LocationData();
     // handles edge cases around Android where start might get called again even though
     // the service is already created.  Make sure the listeners are still bound and exit
-    if (isBackgroundGeolocationConfigured) {
-      BackgroundGeolocation.start();
-      return;
-    }
+    let result = BackgroundGeolocation.checkStatus(({ isRunning }) => {
```
Patrick-Erichsen, line 216: This piece of logic (L216-L224) feels like it's getting complicated enough to pull into its own method.

I think it would be a bit cleaner to just do the boolean status check in the method and start BackgroundGeolocation afterwards. E.g.


```js
if (this.isBackgroundGeolocationRunning()) {
  return 
} else {
  BackgroundGeolocation.start()
}
```

Patrick-Erichsen: Also - I believe we can get rid of [isBackgroundGeolocationConfigured](https://github.com/tripleblindmarket/covid-safe-paths/pull/725/files#diff-3a190ea6886abecabc4984cc71181843R12) and all references to it in the file now?

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
+                },
+              ],
+            ),
+          1000,
+        );
       }
     });
   }
 
+  static isheadlessAlreadyRunning() {
```
Patrick-Erichsen, line 484:
```suggestion
  static isHeadlessTaskRunning() {
```

Also - please add a small comment to the method.

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
+    return BackgroundGeolocation.getConfig(config => {
+      if (!config.stopOnTerminate && config.notificationsEnabled) return true;
```
Patrick-Erichsen, line 486: this could be simplified to 

```js
return BackgroundGeolocation.getConfig(config => {
  return !config.stopOnTerminate && config.notificationsEnabled
});
```

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
+          id: LOCATION_DISABLED_NOTIFICATION,
+          title: languages.t('label.location_disabled_title'),
+          message: languages.t('label.location_stopped_message'),
+        });
+      }
+    });
     isBackgroundGeolocationConfigured = false;
     SetStoreData(PARTICIPATE, 'false').then(() => {
-      // nav.navigate('LocationTrackingScreen', {});
+      //nav.navigate('LocationTrackingScreen', {});
```
Patrick-Erichsen, line 512: I think we can just remove this block of stale code altogether as part of the PR.

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
   static stop() {
     // unregister all event listeners
-    PushNotification.localNotification({
-      title: languages.t('label.location_disabled_title'),
-      message: languages.t('label.location_disabled_message'),
-    });
     BackgroundGeolocation.removeAllListeners();
     BackgroundGeolocation.stop();
 
+    BackgroundGeolocation.checkStatus(({ locationServicesEnabled }) => {
```
Patrick-Erichsen, line 495: Could we pull the callback into a standalone method here? Something along the lines of `notifyDisabledLocation()`?

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
+      if (!locationServicesEnabled) {
```
Patrick-Erichsen, line 496: Since the only thing different between these two statements is the message, i think it would read a bit cleaner like this 

```js
const message = locationServicesEnabled
  ? languages.t('label.location_disabled_message')
  : languages.t('label.location_stopped_message');
        
PushNotification.localNotification({
  id: LOCATION_DISABLED_NOTIFICATION,
  title: languages.t('label.location_disabled_title'),
  message,
});
```

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
-        //           text: languages.t('label.no'),
-        //           onPress: () => console.log('No Pressed'),
-        //           style: 'cancel',
-        //         },
-        //       ],
-        //     ),
-        //   1000,
-        // );
+        setTimeout(
+          () =>
```
Patrick-Erichsen, line 459: same thing - looks to be the same alert from above

app/services/LocationService.js
```diff
@@ -383,70 +423,93 @@ export default class LocationServices {
...
-        //           text: languages.t('label.no'),
-        //           onPress: () => console.log('No Pressed'),
-        //           style: 'cancel',
-        //         },
-        //       ],
-        //     ),
-        //   1000,
-        // );
+        //we need to set delay or otherwise alert may not be shown
+        setTimeout(
```
Patrick-Erichsen, line 427: looks like this is the exact code from above? if so - we can just call the same method here after extracting the block above into a standalone method.

app/services/LocationService.js
```diff
@@ -296,40 +303,77 @@ export default class LocationServices {
...
+                      },
+                      style: 'cancel',
+                    },
+                  ],
+                ),
+              1000,
+            );
           }
+
+          //          if (!locationServicesEnabled) {
```
Patrick-Erichsen, line 366: can remove this comment block

app/services/LocationService.js
```diff
@@ -296,40 +303,77 @@ export default class LocationServices {
...
-              id: LOCATION_DISABLED_NOTIFICATION,
-              title: languages.t('label.location_disabled_title'),
-              message: languages.t('label.location_disabled_message'),
-            });
-          } else {
-            PushNotification.cancelLocalNotifications({
-              id: LOCATION_DISABLED_NOTIFICATION,
-            });
+            //we need to set delay or otherwise alert may not be shown
+            setTimeout(
```
Patrick-Erichsen, line 334: could we pull this big alert body into its own method here? something along the lines of `alertRequiredLocationServices()`?

app/services/LocationService.js
```diff
@@ -296,40 +303,77 @@ export default class LocationServices {
...
-        //           onPress: () => console.log('No Pressed'),
-        //           style: 'cancel',
-        //         },
-        //       ],
-        //     ),
-        //   1000,
-        // );
+        setTimeout(
+          () =>
+            Alert.alert(
```
Patrick-Erichsen, line 308: could we pull this big alert body into its own method here? something along the lines of `alertRequiredLocationInfo()`?

app/locales/en.json
```diff
@@ -100,23 +100,34 @@
...
+    "location_stopped_message": "Stopped at user request",
     "logging_active": "Location Active",
     "logging_inactive": "Location Inactive",
+    "maps_import_button_text": "Import past locations",
+    "maps_import_disclaimer": "Safe Paths has no affiliation with Google and never shares your data.",
+    "maps_import_text": "To see if you encountered someone with COVID-19 prior to downloading this app, you can import your personal location history.",
+    "maps_import_title": "Google Maps",
     "news_subtitle": "Read about the latest COVID updates from your health authority and in general.",
     "news_title": "Latest news",
+    "no": "No",
```
Patrick-Erichsen, line 116: Please move this string, and `"Yes"`, to the `"common"` section.

app/locales/en.json
tstirrat: We iterated on the language stuff in #574, though the diff there makes it hard to see. I think that this is a newer version of that change, is that right @aitkeni ?

The updated language seems fine to me.

app/locales/en.json
```diff
@@ -100,23 +100,34 @@
...
+    "location_disabled_title": "Device location disabled",
+    "location_enabled_message": "Safe Paths is anonymously recording your location.",
+    "location_enabled_title": "Device location Enabled",
+    "location_stopped_message": "Stopped at user request",
     "logging_active": "Location Active",
     "logging_inactive": "Location Inactive",
+    "maps_import_button_text": "Import past locations",
+    "maps_import_disclaimer": "Safe Paths has no affiliation with Google and never shares your data.",
+    "maps_import_text": "To see if you encountered someone with COVID-19 prior to downloading this app, you can import your personal location history.",
+    "maps_import_title": "Google Maps",
```
tstirrat, line 113: These maps strings look like they came from a bad merge.

Try running `yarn i18n:extract` to remove any unused keys.

app/services/LocationService.js
```diff
@@ -212,10 +213,15 @@ export default class LocationServices {
...
-    }
+    let result = BackgroundGeolocation.checkStatus(({ isRunning }) => {
+      if (isRunning) {
+        BackgroundGeolocation.start();
+        return true;
+      } else {
+        return false;
+      }
+    });
+    if (result) return;
```
tstirrat, line 224: From what I can tell there is no return type of [checkStatus()](https://github.com/mauron85/react-native-background-geolocation#checkstatussuccess-fail). 

I think this will always be false, remove.

app/services/LocationService.js
tstirrat: The return type of this method is probably always `undefined` because [it's not a promise](https://github.com/mauron85/react-native-background-geolocation/blob/c13b3ce45a16a1b538be6ddb1bb02f48c23354fb/index.js#L130-L134), can you confirm?

I think what you want is:

```js
return new Promise((resolve, reject) => {
  BackgroundGeolocation.getConfig((config, error) => {
    if (error) { 
      reject(error);
    } else {
      resolve(!config.stopOnTerminate && config.notificationsEnabled);
    }
});
```

app/locales/en.json
aitkeni: yes that is correct.  There is a new change which is the word "securely" is changed to "anonymously" in the string "Safe Paths is anonymously recording your location."  This I felt was more accurate and better PR for the app.

app/services/LocationService.js
aitkeni: will do

app/services/LocationService.js
aitkeni: will do

app/locales/en.json
tstirrat: OK, will close the other one

aitkeni:
One thing I need to do is add some unit tests... Also would be good for an Iphone user to try it to see if the iphone settings opens up ok before i merge.

app/services/LocationService.js
```diff
@@ -380,73 +363,86 @@ export default class LocationServices {
...
+        if (error) {
+          reject(error);
+        } else {
+          resolve(isRunning);
+        }
+      });
+    });
+  }
+
+  static async isHeadlessTaskRunning() {
```
Patrick-Erichsen, line 412:
```suggestion
  static isHeadlessTaskRunning() {
```

Since we aren't using the `await` keyword this method and `checkIfAlreadyRunning` do not need `async`. Check out the Return value of the [MDN async docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)

app/services/LocationService.js
```diff
@@ -212,11 +212,17 @@ export default class LocationServices {
     const locationData = new LocationData();
     // handles edge cases around Android where start might get called again even though
     // the service is already created.  Make sure the listeners are still bound and exit
-    if (isBackgroundGeolocationConfigured) {
+    if (LocationServices.checkIfAlreadyRunning() === true) {
```
Patrick-Erichsen, line 215:
```suggestion
    if (LocationServices.checkIfAlreadyRunning()) {
```

app/services/LocationService.js
```diff
@@ -270,7 +276,10 @@ export default class LocationServices {
       });
     });
 
-    if (isPlatformAndroid()) {
+    if (
+      LocationServices.isHeadlessTaskRunning() === true &&
```
Patrick-Erichsen, line 280:
```suggestion
      LocationServices.isHeadlessTaskRunning() &&
```

Patrick-Erichsen:
Thanks for incorporating all of that feedback 👍 

However - I still am getting a `Location Active` after following the test instructions. Can you confirm that you are seeing `Location Inactive` after manually disabling location services in device settings?

tstirrat:
Closing because there has been no activity in 8 days and we made changes to LocationService.js and refactored location checks in #749

Issue #527, opened 2020-04-15
Android location saving after restart part 2


troach-sf:
Safe Paths is not saving locations after device restart until the app is opened. Part 1 of the issue was here: https://github.com/tripleblindmarket/covid-safe-paths/issues/523. There was a react library bug that prevented the geolocation service from restarting upon a phone reboot.

This issue is for part 2, which is that the Safe Paths geolocation listener no longer exists after a device restart.

## Version

- Type of phone: Android (Solution will impact iPhone)
- OS Version: All
- Model: All

## Steps to reproduce

Take note of the locations inside RKStorage dataset (async-storage). Restart the application and do not open Safe Paths. No additional locations are logged in RKStorage until Safe Paths opens.

## Expected Behavior

When I restart my device, I expect new locations logged in RKStorage before opening Safe Paths application.

## Solution

react-native-background-geolocation has its own sqlite database of lat/long/timestamp with additional information. We can set a max row count of 8,064 which is 1 row / every 5 minutes / for 28 days.

We can then get list of locations by calling getLocations (https://github.com/mauron85/react-native-background-geolocation#getlocationssuccess-fail). We can then further filter on timestamp to ensure we only keep 28 days of data in the list.

There is no need to store locations in async-storage. At this point, it is only duplicating data that already exists in the react-native-background-geolocation database. This will be far more efficient, as the current method of adding a location in async-storage involves pulling the entire list of locations, appending a new location in an array, and then re-saving a json blob to sqlite.

The listener will no longer be required to store locations, and can be repurposed just for logging.

troach-salesforce:
Async-Storage database. App installed, device restarted shortly after opening. 1 location saved.

```
[
  {
    "latitude": 34.24109059012831,
    "longitude": -87.95186424460518,
    "time": 1586979689990
  }
]
```

react-native-background-geolocation database. Same scenario as above with 2 additional locations tracked after restart.

```
[
  {
    "_id": 1,
    "time": 1586979689990,
    "accuracy": 1.27143120765686,
    "speed": 0.296305745840073,
    "bearing": 0,
    "altitude": 42.7420883178711,
    "latitude": 34.2410905901283,
    "longitude": -87.9518642446052,
    "radius": 0,
    "has_accuracy": 1,
    "has_speed": 1,
    "has_bearing": 0,
    "has_altitude": 1,
    "has_radius": 0,
    "provider": "network",
    "service_provider": 0,
    "valid": 2,
[...]

PR #747, opened 2020-05-01
Securely Store Location Data + Other improvements

troach-sf:
Securely store locations in encrypted Realm db

This PR securely stores all captured Geolocation coordinates in an encrypted Realm database.

* Secures Locations imported via Google
* Migrates old locations to new secure storage -> deletes old on success
* Still allows for json export
* Fixes bug where locations stopped being captured after device restart

#### Linked issues:

#527
Fixes https://github.com/safe-paths-contrib/covid-safe-paths/issues/29

#### How to test

For debug builds, you can download the Realm database inside the apps storage. You can open the database using Realm Studio: https://realm.io/products/realm-studio/. Because the database is encrypted, you will need to log a hex-encoded encryption key.


troach-sf:
@tstirrat @kenpugsley Can you assign yourselves as reviewers?

troach-sf:
Encryption review is pending. Lets hold off on merging until we get an approval.

android/app/build.gradle
```diff
@@ -210,7 +216,9 @@ dependencies {
     } else {
         implementation jscFlavor
     }
-    androidTestImplementation('com.wix:detox:+')  // detox: e2e testing
+    androidTestImplementation('com.wix:detox:+')
+    implementation "androidx.core:core-ktx:+"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version" // detox: e2e testing
```
tstirrat, line 221: the comment here is for the detox line

android/app/src/main/java/org/pathcheck/covidsafepaths/storage/Location.kt
```diff
@@ -0,0 +1,93 @@
...
+          speed = if (backgroundLocation.hasSpeed()) backgroundLocation.speed else null,
+          accuracy = if (backgroundLocation.hasAccuracy()) backgroundLocation.accuracy else null,
+          bearing = if (backgroundLocation.hasBearing()) backgroundLocation.bearing else null,
+          provider = backgroundLocation.provider,
+          mockFlags = backgroundLocation.mockFlags,
+          source = SOURCE_DEVICE
+      )
+    }
+
+    fun fromImportLocation(
```
tstirrat, line 62: is this from "google" ImportLocation?

troach-sf: Depends. Could be google or migration. Source is passed in that method.

android/app/build.gradle
troach-sf: Odd, I think that got moved by enabling Kotlin automatically. I’ll correct.

android/app/src/main/java/org/pathcheck/covidsafepaths/storage/Location.kt
tstirrat: Ah, got it.

tremblerz:
Great work!

Ferrumofomega:
This is wonderful stuff @troach-sf .

.github/workflows/test.yml
```diff
@@ -85,7 +87,7 @@ jobs:
           restore-keys: |
             ${{ runner.OS }}-pods-
 
-      - run: pod install --repo-update
+      - run: pod install --repo-update --verbose
```
efalkner, line 90: Not sure we want to have verbose logging on every build

tstirrat: There was a terrible bug with no actionable output except "-e script.sh".. so this was helpful for that... I wonder if there's a way to turn it on via a special commit message maybe?

troach-sf: I went ahead and removed.

troach-sf:
@tstirrat, I've completed testing on both sides. Looks like there is 1 conflicting file that needs to be resolved now. I'll merge that in tomorrow morning.

tstirrat:
Thanks! How does one run the native tests?

It would be good to encode that into 2 more jobs in the `test.yml` workflow.. /cc @efalkner

troach-sf:
I just wrote a document on how to manually capture encryption key for develop builds to allow for opening the database. This should help for manual tests. https://gist.github.com/troach-sf/f257bb7b80e6dddd4f3bade81b7b1410

For running the automatic tests.

Android:
```
cd android
../gradlew clean
../gradlew connectedAndroidTest
```

iOS:
Haven't tried from the command line but open the project, open the CovidSafePathsTests scheme, and click on the test icon in the sidebar and run all tests in RealmSecureStorageTest

troach-sf:
Right now Android requires a device/emulator. I don't think Realm is compatible with Robolectric for unit testing without a device. That may make it tricky to get working on GH Actions.

kenpugsley:
Note to others ... I'm currently working on doing some functional testing on this, and will try to run new native tests locally.

app/services/LocationService.js
```diff
@@ -24,114 +23,7 @@ export class LocationData {
...
-
-      // Curate the list of points, only keep the last 28 days
-      let curated = [];
-      for (let i = 0; i < locationArray.length; i++) {
-        if (locationArray[i]['time'] > unixtimeUTC_28daysAgo) {
-          curated.push(locationArray[i]);
-        }
-      }
-
-      // Backfill the stationary points, if available
```
kenpugsley, line 83: Sorry I'm coming late to the party here ... it looks like we no longer have this backfill capability anymore.  Is that intended?

I think without this backfill we lose the capability to intersect in cases where the phone was essentially still for a long period but an intersecting time-location is reported that matches.  It also fills in large gaps of non-movement, which is helpful when the data is loaded by a health authority.

kenpugsley:
> cd android
> ../gradlew clean
> ../gradlew connectedAndroidTest

This doesn't work for me.  It's failing in the geolocation tests ... any ideas?
`> Task :@mauron85_react-native-background-geolocation:connectedDebugAndroidTest
12:47:08 V/ddms: execute: running am get-config
12:47:09 V/ddms: execute 'am get-config' on 'emulator-5554' : EOF hit. Read: -1
12:47:09 V/ddms: execute: returning
12:47:09 D/@mauron85_react-native-background-geolocation-debug-androidTest.apk: Uploading @mauron85_react-native-background-geolocation-debug-androidTest.apk onto device 'emulator-5554'
12:47:09 D/Device: Uploading file onto device 'emulator-5554'
12:47:09 D/ddms: Reading file permision of /Users/ken/code/covid-safe-paths/node_modules/@mauron85/react-native-background-geolocation/android/lib/build/outputs/apk/androidTest/debug/@mauron85_react-native-background-geolocation-debug-androidTest.apk as: rw-r--r--
12:47:09 V/ddms: execute: running pm install -r -t "/data/local/tmp/@mauron85_react-native-background-geolocation-debug-androidTest.apk"
12:47:12 V/ddms: execute 'pm install -r -t "/data/local/tmp/@mauron85_react-native-background-geolocation-debug-androidTest.apk"' on 'emulator-5554' : EOF hit. Read: -1
12:47:12 V/ddms: execute: returning
12:47:12 V/ddms: execute: running rm "/data/local/tmp/@mauron85_react-native-background-geolocation-debug-androidTest.apk"
12:47:13 V/ddms: execute 'rm "/data/local/tmp/@mauron85_react-native-background-geolocation-debug-androidTest.apk"' on 'emulator-5554' : EOF hit. Read: -1
12:47:13 V/ddms: execute: returning
12:47:13 I/RemoteAndroidTest: Running am instrument -w -r   com.marianhello.bgloc.react.test/android.support.test.runner.AndroidJUnitRunner on Nexus_4_API_27(AVD) - 8.1.0
12:47:13 V/ddms: execute: running am instrument -w -r   com.marianhello.bgloc.react.test/android.support.test.runner.AndroidJUnitRunner
12:47:13 V/InstrumentationResultParser: INSTRUMENTATION_RESULT: shortMsg=Process crashed.
12:47:13 I/InstrumentationResultParser: test run failed: 'Instrumentation run failed due to 'Process crashed.''`

troach-sf:
@kenpugsley ah sorry, run :app:androidConnectedTest to limit to just our own tests. Good eye on the backfill. I discussed this with @Ferrumofomega and @tremblerz that we would only store real locations in the db. It would be easy enough to add backfilling in on the js layer when pulling locations. After discussing that a bit, I believe we came to an agreement that if there is a delay we shouldn't make assumptions someone stayed in one spot if the previous location had barely changed. I'll leave that up to you all to decide, but if we add it back in, rather than throwing assumed locations in the db, we should only add during processing.

app/services/LocationService.js
troach-sf: See below comment.

kenpugsley:
> we should only add during processing.

I see the point on the intersection, but this also have effects on the flow of data to safe paths.  I see the point about not wanting location data that is inferred stored, but this will be a breaking change to the "contract" (not that there is one explicitly) with safe places.  If we don't store the points in the local db, then in both the intersection as well as the data export we have to remember to backfill the gaps.  Seems like a service layer thing to me.

tstirrat:
Closing this in favor of #788 where we are automating tests and merging conflicts.

AdamLeonSmith:
> I just wrote a document on how to manually capture encryption key for develop builds to allow for opening the database. This should help for manual tests. https://gist.github.com/troach-sf/f257bb7b80e6dddd4f3bade81b7b1410

I'm getting compilation errors (Android) when I add this immediately after SecureRandom().nextBytes(newKey) in getEncryptionKey()

troach-sf:
@AdamLeonSmith, I have updated the gist. There was a missing ")".

PR #749, opened 2020-05-01
Refactor location status checks

dahliasalem:


app/Entry.js
```diff
@@ -55,7 +56,7 @@ class Entry extends Component {
           {this.state.initialRouteName === 'true' ? (
             <Stack.Screen
               name='InitialScreen'
-              component={LocationTracking}
+              component={LocationController}
```
tstirrat, line 59: Since you're making new files, let's call this "Home" or "Main" now

It once was NOT the main page, but now it is.

@kenpugsley any pref on Home or Main?

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
...
+          console.log('Found crossed paths');
+          resolve(true);
+        } else {
+          console.log("Can't find crossed paths");
+          resolve(false);
+        }
+      });
+    });
+  }
+  static getParticpating() {
```
tstirrat, line 471: nit: blank line above method

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
...
+    return new Promise(resolve => {
+      GetStoreData(PARTICIPATE, false).then(isParticipating => {
+        resolve(isParticipating);
```
tstirrat, line 474: nit: this will return string `'true'`, so do a `isParticipating === 'true'`

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
...
+        reason: 'USER_OFF',
+        crossedPaths: crossedPaths,
+      };
+    }
+
+    const status = await this.getBackgroundGeoStatus();
+    if (!status.locationServicesEnabled) {
+      return {
+        canTrack: false,
+        reason: 'LOCATION_OFF',
```
tstirrat, line 503: make a enum-ish const for these so they can be exported and used elsewhere to prevent magic strings.

```js
const Reason = {
  LOCATION_OFF: 'LOCATION_OFF',
};
```

app/views/LocationController.js
```diff
@@ -0,0 +1,114 @@
...
+import { isPlatformAndroid } from './../Util';
+import Colors from '../constants/colors';
+import LocationServices from '../services/LocationService';
+import DefaultPage from './tracking/DefaultPage';
+import ExposurePage from './tracking/ExposurePage';
+import OffPage from './tracking/OffPage';
+import styles from './tracking/style';
+import UnknownPage from './tracking/UnknownPage';
+
+export default class LocationController extends Component {
```
tstirrat, line 23: use a functional component with useState()

app/views/LocationController.js
```diff
@@ -0,0 +1,114 @@
...
+    super(props);
+
+    if (isPlatformAndroid()) {
+      StatusBar.setBackgroundColor(Colors.TRANSPARENT);
+      StatusBar.setBarStyle('light-content');
+      StatusBar.setTranslucent(true);
+    }
+
+    this.state = {
+      location: {
```
tstirrat, line 34: use useState for these.

app/views/LocationController.js
```diff
@@ -0,0 +1,114 @@
...
+
+  getSettings() {
+    return (
+      <TouchableOpacity
+        style={styles.settingsContainer}
+        onPress={() => {
+          this.props.navigation.navigate('SettingsScreen');
+        }}>
+        {/* Is there is a reason there's this imageless image tag here? Can we delete it? */}
+        <Image resizeMode={'contain'} />
```
tstirrat, line 76: delete it. I think it was an attempt tofix a bug in ios 12. which we fixed in another way.

app/views/LocationController.js
```diff
@@ -0,0 +1,114 @@
...
+        page = <OffPage />;
+      } else {
+        page = <UnknownPage />;
+      }
+    }
+
+    return (
+      <View style={styles.backgroundImage}>
+        {page}
+        {this.getSettings()}
```
tstirrat, line 110: Make smaller components instead of `this.getX()`


```jsx
<SettingsIcon onPress={} />
```

app/views/tracking/DefaultPage.js
```diff
@@ -0,0 +1,56 @@
...
+            {languages.t('label.home_no_contact_subtext')}
+          </Typography>
+        </View>
+      </View>
+      {MayoInfo()}
+    </ImageBackground>
+  );
+};
+
+export default DefaultPage;
```
tstirrat, line 56: no default exports, named exports are easier for intellisense/autocomplete to find.

app/views/tracking/ExposurePage.js
```diff
@@ -0,0 +1,74 @@
...
+              onPress={() => {
+                navigation.navigate('ExposureHistoryScreen');
+              }}
+              buttonColor={Colors.BLUE_BUTTON}
+              bgColor={Colors.WHITE}
+            />
+          </View>
+        </View>
+      </View>
+      {MayoInfo()}
```
tstirrat, line 69: `<MayoInfo />`

app/views/tracking/MayoInfo.js
```diff
@@ -0,0 +1,32 @@
...
+import { Image, Linking, TouchableOpacity, View } from 'react-native';
+
+import foreArrow from './../../assets/images/foreArrow.png';
+import { Typography } from './../../components/Typography';
+import languages from '../../locales/languages';
+import styles from './style';
+
+const MAYO_COVID_URL = 'https://www.mayoclinic.org/coronavirus-covid-19';
+
+const Press = () => Linking.openURL(MAYO_COVID_URL);
```
tstirrat, line 11: naming: `onPress` or `handlePress`

app/views/tracking/OffPage.js
```diff
@@ -0,0 +1,69 @@
...
+import Colors from './../../constants/colors';
+import ButtonWrapper from '../../components/ButtonWrapper';
+import languages from '../../locales/languages';
+import MayoInfo from './MayoInfo';
+import styles from './style';
+
+const buttonLabel = languages.t('label.home_enable_location');
+const size = Dimensions.get('window').height;
+
+export default function OffPage() {
```
tstirrat, line 24: no default exports and use the `export const OffPage = () => {}` format to be consistent with other files.

I'll eventually write this up as a wiki style guide!

app/views/tracking/OffPage.js
```diff
@@ -0,0 +1,69 @@
...
+import {
+  Dimensions,
+  ImageBackground,
+  StatusBar,
+  Text,
+  View,
+} from 'react-native';
+import { SvgXml } from 'react-native-svg';
+
+import BackgroundImage from './../../assets/images/launchScreenBackground.png';
```
tstirrat, line 12: use camelCase for non classes

This is why I prefer no default exports :)

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,73 @@
...
+import StateUnknown from './../../assets/svgs/stateUnknown';
+import { Typography } from './../../components/Typography';
+import Colors from './../../constants/colors';
+import ButtonWrapper from '../../components/ButtonWrapper';
+import languages from '../../locales/languages';
+import MayoInfo from './MayoInfo';
+import styles from './style';
+
+const buttonLabel = languages.t('label.home_enable_location');
+const buttonFunction = () => {
```
tstirrat, line 22: very vague, perhaps `handleEnableLocationPress`

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,73 @@
...
+import BackgroundImage from './../../assets/images/launchScreenBackground.png';
+import StateUnknown from './../../assets/svgs/stateUnknown';
+import { Typography } from './../../components/Typography';
+import Colors from './../../constants/colors';
+import ButtonWrapper from '../../components/ButtonWrapper';
+import languages from '../../locales/languages';
+import MayoInfo from './MayoInfo';
+import styles from './style';
+
+const buttonLabel = languages.t('label.home_enable_location');
```
tstirrat, line 21: use `const {t} = useTranslation()` inside the functional component. because its reactive.

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,73 @@
...
+const buttonFunction = () => {
+  openSettings();
+};
+const size = Dimensions.get('window').height;
```
tstirrat, line 25: I think there is a useDimension hook

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,73 @@
...
+              onPress={() => {
+                buttonFunction();
+              }}
+              buttonColor={Colors.BLUE_BUTTON}
+              bgColor={Colors.WHITE}
+            />
+          </View>
+        </View>
+      </View>
+      {MayoInfo()}
```
tstirrat, line 68: `<MayoInfo />`

app/views/tracking/style.js
```diff
@@ -0,0 +1,121 @@
...
+    fontFamily: fontFamily.primaryRegular,
+  },
+  arrowContainer: {
+    alignSelf: 'center',
+    paddingRight: 20,
+    paddingLeft: 20,
+  },
+});
+
+export default styles;
```
tstirrat, line 121: no default exports

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,73 @@
...
+    </ImageBackground>
+  );
+};
+
+export default UnknownPage;
```
tstirrat, line 73: no default exports

app/views/tracking/DefaultPage.js
```diff
@@ -0,0 +1,56 @@
...
+          xml={StateNoContact}
+          width={size ? size : 80}
+          height={size ? size : 80}
+        />
+      </View>
+      <View style={styles.mainContainer}>
+        <View style={styles.contentAbovePulse} />
+        <View style={styles.contentBelowPulse}>
+          <Typography style={styles.mainTextBelow}>
+            {languages.t('label.home_no_contact_header')}
```
tstirrat, line 44: `const {t} = useTranslation()` hook here and everywhere (inside functional components anyways)

app/views/LocationController.js
```diff
@@ -0,0 +1,114 @@
...
+    this.state = {
+      location: {
+        canTrack: true,
+        reason: '',
+        crossedPaths: false,
+      },
+    };
+  }
+
+  componentDidMount() {
```
tstirrat, line 42: this becomes 

```js
useEffect(() => {
  // on mount code

  return () => { // teardown code };
}, []);
```

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
       // nav.navigate('LocationTrackingScreen', {});
     });
   }
+
+  static getCrossedPaths() {
```
tstirrat, line 457: this returns a boolean and not the actual crossed paths, so perhaps `getHasCrossedPaths`

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
...
+        resolve(status);
+      });
+    });
+  }
+
+  static async checkStatus() {
+    const crossedPaths = await this.getCrossedPaths();
+    const particpating = await this.getParticpating();
+
+    if (!particpating) {
```
tstirrat, line 491: love all this. It needs some unit tests. 

The general direction is great.

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
...
+      return {
+        canTrack: false,
+        reason: 'USER_OFF',
+        crossedPaths: crossedPaths,
```
tstirrat, line 495: Just realizing this is a boolean, so give it a has/is prefix

And... maybe just `hasPotentialExposure`

app/services/LocationService.js
```diff
@@ -449,4 +453,70 @@ export default class LocationServices {
       // nav.navigate('LocationTrackingScreen', {});
     });
   }
+
+  static getCrossedPaths() {
+    return new Promise(resolve => {
+      GetStoreData(CROSSED_PATHS).then(dayBin => {
+        dayBin = JSON.parse(dayBin);
+        if (dayBin !== null && dayBin.reduce((a, b) => a + b, 0) > 0) {
```
tstirrat, line 461: the check is more clearly expressed as `dayBin.some(exposure => exposure > 0)`

reduce has way more overhead.

app/services/LocationService.js
dahliasalem: It returns a boolean because we pass false as the second argument

tstirrat: If it's `'true'` in the store, it will resolve as `'true'` and not `true` though.

app/views/tracking/UnknownPage.js
dahliasalem: the default export of the languages file uses the `useTranslation()` hook

app/services/LocationService.js
dahliasalem: I am not sure I am following. The getstore will return JSON.parse('true') which returns the boolean true so the promise will return the boolean as well when I resolve it

app/views/tracking/DefaultPage.js
dahliasalem: the exported languages variable uses the useTranslation hook to instantiate

app/views/tracking/UnknownPage.js
dahliasalem: I tried using this one https://github.com/Swizec/useDimensions but the returned size width/height was much smaller

app/views/LocationController.js
```diff
@@ -13,102 +13,90 @@ import { SvgXml } from 'react-native-svg';
...
-    });
-    // refresh state if settings change
-    this.unsubscribe = this.props.navigation.addListener('focus', () => {
-      this.updateStateInfo();
-    });
-    // handle back press
-    BackHandler.addEventListener('hardwareBackPress', this.handleBackPress);
+export const LocationController = props => {
+  if (isPlatformAndroid()) {
+    StatusBar.setBackgroundColor(Colors.TRANSPARENT);
```
tstirrat: Is this imperative stuff needed? I think you can do it in the render block as JSX: `<StatusBar backgroundColor=x />`

If actually needed here as a side effect, add a comment why, and put all these bits in the main `useEffect` block.

app/views/LocationController.js
```diff
@@ -13,102 +13,90 @@ import { SvgXml } from 'react-native-svg';
...
-    return true;
-  };
+  const [location, setLocation] = useState({
+    canTrack: true,
+    reason: '',
+    hasPotentialExposure: false,
+  });
 
-  getSettings() {
+  const SettingsIcon = () => {
```
tstirrat: Move out into module scope... but I think this whole component be replaced with `<IconButton>` in the common components dir.

app/views/LocationController.js
```diff
@@ -13,102 +13,90 @@ import { SvgXml } from 'react-native-svg';
...
-      <View style={styles.backgroundImage}>
-        {page}
-        {this.getSettings()}
-      </View>
-    );
   }
-}
+
+  return (
+    <View style={styles.backgroundImage}>
```
tstirrat: Here is a good place to switch themes. `<Theme>` handles the text and icon button color for child components.

`<Theme use={getTheme}>`

And `getTheme` can switch between `violet` for normal modes and `charcoal` when there is an exposure etc

app/services/LocationService.js
tstirrat: GetStoreData does not `JSON.parse()` if the [value is a string](https://github.com/tripleblindmarket/covid-safe-paths/blob/8e684228a0d543964df402269335470de1d45927/app/helpers/General.js#L14-L16), it really *should*, but it doesn't at the moment.

tstirrat: oh. I thought that was isString(value).... nevermind. carry on.

app/views/tracking/ExposurePage.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <ImageBackground source={backgroundImage} style={styles.backgroundImage}>
+      <StatusBar
+        barStyle='light-content'
+        backgroundColor='transparent'
+        translucent
+      />
+      <View style={styles.pulseContainer}>
+        <Text>Testdsfafasfsdafasfsadf</Text>
```
tstirrat, line 35: Whatever this is LOL, probably should be removed.

app/views/tracking/OffPage.js
```diff
@@ -0,0 +1,69 @@
...
+  const size = Dimensions.get('window').height;
+  return (
+    <ImageBackground source={backgroundImage} style={styles.backgroundImage}>
+      <StatusBar
+        barStyle='light-content'
+        backgroundColor='transparent'
+        translucent
+      />
+      <View style={styles.pulseContainer}>
+        <Text>Testdsfafasfsdafasfsadf</Text>
```
tstirrat, line 34: Remove.

app/views/tracking/UnknownPage.js
```diff
@@ -0,0 +1,71 @@
...
+  const size = Dimensions.get('window').height;
+  return (
+    <ImageBackground source={backgroundImage} style={styles.backgroundImage}>
+      <StatusBar
+        barStyle='light-content'
+        backgroundColor='transparent'
+        translucent
+      />
+      <View style={styles.pulseContainer}>
+        <Text>Testdsfafasfsdafasfsadf</Text>
```
tstirrat, line 36: Remove.

app/views/tracking/MayoInfo.js
```diff
@@ -0,0 +1,31 @@
...
+import { useTranslation } from 'react-i18next';
+import { Image, Linking, TouchableOpacity, View } from 'react-native';
+
+import foreArrow from './../../assets/images/foreArrow.png';
+import { Typography } from './../../components/Typography';
+import { styles } from './style';
+
+const MAYO_COVID_URL = 'https://www.mayoclinic.org/coronavirus-covid-19';
+
+export const MayoInfo = () => {
```
tstirrat, line 11: This is a button, `MayoButton`

app/services/LocationService.js
```diff
@@ -446,7 +456,70 @@ export default class LocationServices {
 
     isBackgroundGeolocationConfigured = false;
     SetStoreData(PARTICIPATE, 'false').then(() => {
-      // nav.navigate('LocationTrackingScreen', {});
+      // nav.navigate('Main', {});
     });
   }
+  static getHasPotentialExposure() {
```
tstirrat, line 462: Please add some tests for all the new methods you added here.

app/views/Main.js
```diff
@@ -0,0 +1,116 @@
...
+  } else {
+    if (location.reason === Reason.USER_OFF) {
+      page = <OffPage />;
+    } else {
+      page = <UnknownPage />;
+    }
+  }
+
+  return (
+    <View style={styles.backgroundImage}>
```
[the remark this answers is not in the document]
dahliasalem: the information for the correct background color/style is already contained in the individual components (ExposurePage vs NoKnownExposure)

app/views/Main.js
```diff
@@ -0,0 +1,116 @@
...
+    StatusBar.setTranslucent(true);
+  }
+
+  const [location, setLocation] = useState({
+    canTrack: true,
+    reason: '',
+    hasPotentialExposure: false,
+  });
+
+  const SettingsIcon = () => {
```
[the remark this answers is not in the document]
dahliasalem: IconButton does not support onPress handlers yet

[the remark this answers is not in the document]
tstirrat: Yeah, it does via passing through `{...otherProps}`:

https://github.com/Path-Check/covid-safe-paths/blob/develop/app/components/IconButton.js#L22-L25

Though, I could have documented that better.

app/views/Main.js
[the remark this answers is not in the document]
tstirrat: oh, cool. and that makes more sense!

app/services/__tests__/LocationService.spec.js
```diff
@@ -1,5 +1,8 @@
+import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
+
 import * as General from '../../helpers/General';
-import { LocationData } from '../LocationService';
+import { LocationData, Reason } from '../LocationService';
+import LocationServices from '../LocationService';
 
 function mockGetStoreData(data) {
   General.GetStoreData = () => {
```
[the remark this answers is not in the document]
efalkner: To that end, I highly recommend adding a few methods to a  beforeEach hook after your describe block(s), to clear the state of the mocks between tests. This will ensure that the tests are not dependent on each other's data.
```
beforeEach(() => {
    jest.resetAllMocks();
    jest.resetAllModules();
});
```

tstirrat:
> The only issue I might raise is that we should probably have a state for "no health authority", but I don't think we should hold this up for that.

Yeah, that's being tracked here: https://pathcheck.atlassian.net/browse/SAF-85

aitkeni:
Awesome!

PR #762, opened 2020-05-03
Runtime Feature Flags

Patrick-Erichsen:
#### Description:

Implements runtime feature flagging. Default values for the flags are set from the respective build time env file.

Also creates a `<Switch />` component.

#### Screenshots:

Simulator Screen Shot - iPhone 11 - 2020-05-03 at 17 40 53

#### How to test:

- _Toggles hide the feature as expected_
  - Verify that toggling `Exposure mode` and `google_imports` toggles the respective features in the UI. 
  - Toggle `better_location_status_checks` and verify that `<LocationTracking />` is displayed instead (this uses the `fallback` prop)
- _Toggles persist in async storage_
  - Toggle `google_imports` to off (non-default state). Restart the app. The value should have persisted, and the toggle should still be off.
- _`.env` is the source of truth for which toggles are available_
  - Add a new toggle to `.env`, e.g. `flag_toggle_test`. Run `npx react-native run-android` to rebuild with the new flag. Open the app and toggle the new flag so that it persists to storage. Remove the flag from `.env` and then re-run `npx react-native run-android`. The flag should not be displayed on the toggle screen.
- _The feature flags screen only displays in dev mode_
  - Build a release on either platform and confirm that the `Feature Flags` nav button does not display.


app/views/FeatureFlags.js
```diff
@@ -0,0 +1,129 @@
...
+
+  useEffect(() => {
+    const handleBackPress = () => {
+      navigation.goBack();
+      return true;
+    };
+
+    BackHandler.addEventListener('hardwareBackPress', handleBackPress);
+
+    (async () => {
```
tstirrat, line 26: Give it a name so that if something dies the stack trace will have a name.

app/views/FeatureFlags.js
```diff
@@ -0,0 +1,129 @@
...
+          isExposed
+            ? { ...styles.toggleRowText, color: Colors.RED_TEXT }
+            : styles.toggleRowText
+        }>
+        Exposure Mode
+      </Typography>
+    </View>
+  );
+
+  const getFlagToggleRow = ([flagName, isFlagActive]) => (
```
tstirrat, line 70: This is a component, make it a functional component:

`<FlagToggleRow key={} value={} />`

and move it out into the module scope

app/views/FeatureFlags.js
```diff
@@ -0,0 +1,129 @@
...
+    <View key={flagName} style={styles.toggleRow}>
+      <Switch value={isFlagActive} onValueChange={() => toggleFlag(flagName)} />
+      <View style={styles.toggleRowText}>
+        <Typography use={'body1'} bold>{`${flagName}`}</Typography>
```
tstirrat, line 74: `use="body1"` is fine

app/views/FeatureFlags.js
```diff
@@ -0,0 +1,129 @@
...
+          Default:
+          <Typography
+            style={styles.toggleDefaultVal}
+            use={'body3'}>{` ${getBuildtimeFlagVal(flagName)}`}</Typography>
+        </Typography>
+      </View>
+    </View>
+  );
+
+  const getFlagToggleList = flags => (
```
tstirrat, line 85: make a component, move to module scope

App.js
```diff
@@ -14,13 +14,13 @@ const App = () => {
   }, []);
 
   return (
-    <FlagsProvider flags={buildTimeFlags}>
+    <FlagsContext.Provider value={useState(buildTimeFlags)}>
```
tstirrat, line 17: Can `useState` be moved into `flags.js` so that this is just the provider here? Now that we control the API, may as well locate it inside the flags.js because nothing else should care, own, create or touch the direct state value.

Perhaps `const { FlagsProvider } = createFlags()`

app/views/LocationTracking.js
```diff
@@ -280,6 +281,28 @@ class LocationTracking extends Component {
     );
   }
 
+  getFeatureFlagsBtn() {
```
tstirrat, line 284: its a component, move to module scope.

app/views/LocationTracking.js
```diff
@@ -448,7 +471,8 @@ class LocationTracking extends Component {
             </View>
           </TouchableOpacity>
         </View>
-        {this.getSettings()}
+        {this.getSettingsBtn()}
+        {this.getFeatureFlagsBtn()}
```
tstirrat, line 475: Should this show only in `__DEV__` ?

app/components/__tests__/Feature.spec.js
```diff
@@ -2,76 +2,72 @@ import { render } from '@testing-library/react-native';
...
-      <Feature name='feature1'>
-        <Text>feature1</Text>
-      </Feature>
-    </FlagsProvider>,
-  );
+const mockSetFlags = jest.fn();
 
-  expect(asJSON()).toMatchSnapshot();
-});
+it('renders feature if the flag is enabled', () => {
```
tstirrat, line 10: needs a test that setting flags works.

app/views/LocationTracking.js
Patrick-Erichsen: ah yeah - that is important

App.js
Patrick-Erichsen: are you alright with having to use a ```// eslint-disable-next-line react-hooks/rules-of-hooks``` in that declaration?

app/components/__tests__/Feature.spec.js
Patrick-Erichsen: gave up on this :(

Patrick-Erichsen:
@kenpugsley @tstirrat  this is ready for review. after my 2+ day saga of trying to debug my test that is supposed to verify that toggling a switch on `<FeatureFlagsScreen />` sets the switch value to `false`, I have given up.

Aside from that small issue, there should be sufficient test coverage. Please let me know if you have any issues following my testing instructions above.

app/views/Main.js
```diff
@@ -103,18 +104,19 @@ const Main = () => {
   return (
     <View style={styles.backgroundImage}>
       {page}
-      <SettingsIcon />
+      <SettingsNavButton />
+      <FeatureFlagNavButton navigation={navigation} />
```
tstirrat, line 108: What does this look like? below, to the left ? provide a screenshot

app/views/__tests__/FeatureFlagToggles.spec.js
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+
+import * as flagsEnv from '../../constants/flagsEnv';
+import * as Flags from '../../helpers/Flags';
+import { FeatureFlagsScreen } from '../FeatureFlagToggles';
+
+const { FlagsProvider } = Flags;
+const buildTimeFlags = { feature1: true };
+
+flagsEnv.buildTimeFlags = buildTimeFlags;
```
tstirrat, line 11: potential global mutation, + not reset per test = scary

And I wonder if this is related to why the test code does not toggle....

app/components/Switch.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { Switch as RNSwitch } from 'react-native';
+
+import Colors from '../constants/colors';
+
+export const Switch = ({
+  onValueChange,
+  value = false,
+  trackColor = { false: Colors.GRAY_BACKGROUND, true: Colors.SUCCESS },
```
tstirrat, line 9: `// TODO(https://pathcheck.atlassian.net/browse/SAF-230): make theme aware`

prefer to tackle this using styled components and `({theme}) => theme.success` and `theme.danger`, but better for another PR

Something like: 
- `<Switch>` (green track when true) 
- `<Switch danger>` (red track when true) 
- `<Switch danger invert>` (red when false)

app/helpers/Flags.js
```diff
@@ -0,0 +1,87 @@
...
+import { buildTimeFlags } from '../constants/flagsEnv';
+import { FEATURE_FLAG_VALS } from '../constants/storage';
+import { GetStoreData, SetStoreData } from './General';
+
+/**
+ * Converts the flag name from env into a more human readable form.
+ *
+ * Example:
+ *
+ * `getCleanedFlagName('hello_world')` becomes `Hello World`
```
tstirrat, line 13: nit: semi-shouty case is weird: Flags That Have Long Names Look Strange

prefer "Sentence case for flags"

app/components/Switch.js
Patrick-Erichsen: thanks for making that ticket.

app/views/Main.js
Patrick-Erichsen: Screen Shot 2020-05-09 at 4 14 12 PM

tstirrat: oof. Sorry to push back again on this... This is pretty heavy and could confuse users as we make screenshots of the main page (in dev mode, it's not easy to remove, so would need commenting out of code..)

How do you feel about 

1. long press on the team icon to get to the flags page?

2. long press or 5 presses on the team icon enables flags on this page, and allows to disable?

I'm leaning towards 1.

2. at least allows us to remove this icon for screenshots.

tstirrat: @kenpugsley @efalkner @penrods

Patrick-Erichsen: ah yeah i see how this could be a bottleneck in those scenarios. good point. 

on the flip side though, i do feel like it ends up being a bit of an "insider knowledge" thing with either of those approaches. thinking of a comment that Christian made about "that incantation required to toggle exposure mode".

what do you think of something like this, that would allow us to still take normal screenshots?

Screen Shot 2020-05-09 at 6 44 10 PM

Patrick-Erichsen: However - if the concern is just having any sort of UI element that navigates users to the page, then I like idea #1.

tstirrat: "Insider" "incantation" is actually my preference (like android dev mode, 7 presses on the build number), so that only internal devs and testers find it. Or select people that we inform.

Your option of the last panel in setting is quite good. I like that. 

I'd consider only showing it _after_ the special incantation. But would certainly consider it being there always in dev builds

app/constants/__tests__/flagsEnv.spec.js
```diff
@@ -0,0 +1,28 @@
+import env from 'react-native-config';
+
+import { getBuildtimeFlags, parseFlags } from '../flagsEnv';
+
+describe('getBuildTimeFlags', () => {
+  it('returns the parsed flags', () => {
+    // Default value from __mocks__
```
tstirrat, line 7: Consider setting a default value here, here to keep this test from breaking if others change the defaults.

Then the test is encapsulated to: given, when, then

The global mock is to unblock components that don't care but may use a `<FeatureFlag>` deeper in their tree.

app/views/Main.js
tstirrat: Go with a row in settings for now

app/constants/__tests__/flagsEnv.spec.js
Patrick-Erichsen: Good advice.

Patrick-Erichsen: Neat, didn't know it was this simple to unlock a single method from a file.

```js
jest.mock('../../constants/flagsEnv');
const { parseFlags } = jest.requireActual('../flagsEnv');
```

app/constants/__tests__/flagsEnv.spec.js
```diff
@@ -1,11 +1,14 @@
...
+import { getBuildtimeFlags } from '../flagsEnv';
+
+jest.mock('../../constants/flagsEnv');
+const { parseFlags } = jest.requireActual('../flagsEnv');
 
 describe('getBuildTimeFlags', () => {
   it('returns the parsed flags', () => {
-    // Default value from __mocks__
-    expect(getBuildtimeFlags()).toEqual({ FOO_BAR: true });
+    getBuildtimeFlags.mockReturnValue({ feature1: true });
```
tstirrat, line 10: This test doesn't test anything. You can remove it.

The lower remaining tests could have a mockReturnValue added so they're standalone.

Issue #656, opened 2020-04-24
Repeated notification when location disabled due to redundant checks


tremblerz:
## Version

- Type of phone: Android / iPhone
- OS Version:
- Model: iPhone 11 (simulator)

## Steps to reproduce

Go to Settings, turn off location
1. You get notification
2. now you pull the notification drawer and swipe on notification to get rid of it
3. Drawer closes and the app comes in foreground and another notification gets triggered
4. Go to 2.

## Expected Behavior

Any event should generate notification only once in general, the way we trigger notification in our app is asynchronously checking states at every screen mount/refresh, we need to shift the logic perhaps

eyganeshv:
I will check this issue. We need detail clarification on this issue.

summetj:
@eyganeshv , post questions you have about this issue here and we will do our best to address them.

NaziyaNazir:
This issue is seen for Android also.

ArchanaNair:
@NaziyaNazir, reproducible steps same for android?

NaziyaNazir:
@ArchanaNair Yes, the repro steps are same.

ArchanaNair:
May I know the OS version? I am unable to reproduce it on Android, verifying it on OS version 9

NaziyaNazir:
@ArchanaNair  OS version is Android 9.

ArchanaNair:
@NaziyaNazir , I could see the issue when app goes to background and again comes to foreground, but not in the scenario which is given in reproducible step.

summetj:
This is starting to sound like a low priority issue, anybody disagree?
 And is it strictly a UI/UX issue, or an issue with how the location subsystem works?

ArchanaNair:
> @NaziyaNazir , I could see the issue when app goes to background and again comes to foreground, but not in the scenario which is given in reproducible step

@tremblerz @NaziyaNazir  included a fix for this in android in PR #764

tstirrat:
I could not get this to reproduce on iOS or Android anymore. I believe it was fixed in #749

ArchanaNair:
Yes the issue is fixed in android also.

PR #764, opened 2020-05-04
Fix "too many notifications" when location tracking is disabled

eyganeshv:
#### Linked issues:

#656 

#### How to test:

Go to Settings, turn off location

You get notification
now you pull the notification drawer and swipe on notification to get rid of it
Drawer closes and the app comes in foreground and another notification gets triggered
Go to 2.


tstirrat:
Hello, we just made some changes to location status checks in #749, can you check if this is still an issue in that code?

eyganeshv:
I checked with #749 and this is not issue for iOS.
@ArchanaNair, Could you please check with #749 for Android and confirm?

tstirrat:
I could not get this to reproduce on android anymore. I think it was fixed in #749

PR #766, opened 2020-05-04
Allowing setup on macOS/Darwin for Android only.

martingd:
New option to dev_setup.sh, --skip-ios, to skip configuration related to iOS.
If Xcode is not installed, configuring without this option is not allowed.

 ### Testing Notes ###
Please run ./dev-setup.sh on a Mac without Xcode. Result should be:
  Xcode must be installed to develop for iOS.
  Use option '--skip-ios' to setup development for Android only.
  (exit code 1)
Add option --skip-ios and script should complete setting up
development for Android only.

Please run ./dev-setup on Mac _with_ Xcode and the script should run
like before and setup for both Android and iOS development.

 ### Fixed Issues ###
Setup script dev_setup.sh failed on macOS/Darwin if tools for iOS
development were not installed.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -7,6 +7,7 @@ Usage: dev_setup.sh [options]
    Configures a Linux or MacOS machine for working with the code in this repo.
 Options:
     -ni             Non-interactive mode, auto answers Yes to all questions.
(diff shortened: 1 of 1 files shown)


tstirrat:
Thanks!

Issue #691, opened 2020-04-26
SP location is inaccurate in one instance by 346m


AdamLeonSmith:
## Version

- Type of phone: Android
- OS Version: 9
- Model: Samsung Galaxy S8

The location history recorded by Safe Places is inaccurate on one data point by 346 metres.  The user has also exported their google location history from the same device and provided to me.  The google location history shows the correct location with a high confidence factor.

Additional information can be provided privately/on slack.

diarmidmackenzie:
I have also been investigating GPS data accuracy recoirded by Safe Paths & have similar concerns that our data is
(a) not accurate enough for Safe Paths needs
(b) not as accurate as other GPS services.

This is across 3 phone models (Motorola G7, Samsung J6 and Redmi 6A.

One hypothesis is detailed here:
https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/24412603/26+April+-+GPS+comparison+testing+with+Google+Maps

I wonder whether we are taking the first info we get back on the GPS API as correct, but in reality we need to do more to let the GPS signal "train" before taking it as meaningful?

If we do already have some logic in the app to accommodate GPS training times (or if we don't and we don't know what logic to implement), I wonder whether we could reach out to assistance to organizations who have solved this problem before for some guidance, e.g. Google Maps, Pokemon Go, Geocaching, Strava etc.?

sergesemashko:
Just wanted to share, I've noticed in some GPS hardware (other than smartphones) that quality of GPS signal matters a lot for accurate location. If there is a way to identify signal strength on a smartphone we could add different handling.

diarmidmackenzie:
We are also testing with various GPS Loggers in the Play Store.  Using exactly the same hardware they are delivering much better data than our app.  So this is a software issue.

What they do to get the better data, I don't know - maybe they factor in signal strength.  At least one of them is Open Source, so we can have a look.

troach-sf:
[...]

PR #770, opened 2020-05-06
Add retry logic to get desired location accuracy

ten-sis:
#### Description:

As suggested, retry logic is added in case GPS location returns a low accuracy location. There might be up to 2 retries until the required accuracy of 30 (I guess meters is met). If non of the returned locations has met the threshold the most accurate location is being stored.

If this doesn't improve things much we can try forking [mauron85
/
background-geolocation-android](https://github.com/mauron85/background-geolocation-android/) and updating `ACCURACY_FINE` to `ACCURACY_HIGH` on [this line](https://github.com/mauron85/background-geolocation-android/blob/58464febf6901f1e489817e4425418d48443929b/src/main/java/com/marianhello/bgloc/LocationManager.java#L95).

Also, we can try asking paid geolocation lib providers if they can provide a license for a non-profit open source project :) Lib candidates:
- https://github.com/transistorsoft/react-native-background-geolocation
- https://radar.io/content/alternatives/react-native-background-geolocation-vs-radar

#### Linked issues:

https://pathcheck.atlassian.net/browse/SAF-93https://pathcheck.atlassian.net/browse/SAF-93

#### How to test:

I need help testing this on a real device as I don't have Android phone. Emulator always returns a good accuracy location.

Android:
1. Install an application on a real device
2. Track the location
3. Stored location might potentially have a better quality

### DIFF
--- a/app/constants/location.js
+++ b/app/constants/location.js
@@ -1,0 +1,27 @@
+/**
+ * The max amount of attempts to get desired accuracy.
+ * @type {number}
+ */
+export const MAX_ACCURACY_RETRY = 2;
+
+/**
+ * The desired accuracy (in meters) quality which serves as a threshold.
+ *
+ * @type {number}
+ */
+export const DESIRED_ACCURACY_QUALITY = 30;
+
+/**
+ * The desired location interval, and the minimum acceptable interval.
+ * Time (in milliseconds) between location information polls.
+ * E.g. 60000*5 = 5 minutes
+ * @type {number}
+ */
+export const LOCATION_INTERVAL = 60000 * 5;
+
+/**
+ * Maximum time that we will backfill for missing data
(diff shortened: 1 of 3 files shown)


summetj:
Would it be safe to say this fixes #691 as well?

ten-sis:
@summetj , yes, Jira ticket was created based on #691

troach-sf:
Hey @ten-sis, lets try to sync up in Slack. The fix will need to be applied in on the native library level. Those listeners that are being used to save the location are going to be removed shortly in https://github.com/Path-Check/covid-safe-paths/pull/747. These listeners were problematic because they don't persist device restarts, and were saving unencrypted locations in a sqlite database.

sergesemashko:
closing as per internal conversation with @troach-sf. JS listener doesn’t work after restarting a device until the app is reopened. So, retry/desired accuracy logic should be in native code.

PR #778, opened 2020-05-07
Feature: Logging Location via QR Codes

tyleryasaka:
#### Description:

This feature would allow the integration of QR codes with SafePaths. The idea is that this feature could supplement automatic GPS logging in situations where GPS data might be unavailable or imprecise. Locations which support this feature would use a QR code generator (such as [this demo](https://safepaths-qr-6b479.web.app/) I made) and then display this QR code publicly. People would be able to scan these with their phone. Currently, this supports both deep linking (scanning with an external QR code reader will directly open in the app) and scanning from within the app itself.

Of note, this does not currently support universal linking (a feature which would use a web url to handle the link, which would direct users to the app if they already have it, or to the app store otherwise). I held off on universal linking because there could be privacy concerns there. (I.e. because the QR code has to contain location coordinates as url parameters, technically these would be getting sent to a server somewhere where they could potentially be associated with IP addresses.) Deep linking directly to the app, as I'm doing currently, doesn't have this issue.

- Source code for the QR code generator: https://github.com/tyleryasaka/safe-paths-qr-gen
- QR code generator demo: https://safepaths-qr-6b479.web.app/

#### Linked issues:

N/A

#### Screenshots:

Screenshot_20200509-150014

Screenshot_20200507-141018

##### 1. Testing externally scanned QR codes (deep linking)

- For Android, run this command once you have the app running: `adb shell am start -W -a android.intent.action.VIEW -d "safepaths://qr/34.56/-123.45" org.pathcheck.covidsafepaths`
- For iOS, run this command once you have the app running: `xcrun simctl openurl booted safepaths://qr/34.56/-123.45`
- You should see a screen saying that the scan was successful.
- Tap the "home" button on this screen to return to the main screen. Now tap the settings icon and export your location history. Examine this file. You should see an entry with coordinates corresponding to the latitude 34.56 and longitude -123.45 (from the dummy url above).

##### 2. Testing QR code scanning within the app

- First, go to https://safepaths-qr-6b479.web.app/ and follow the steps to generate a QR code.
- Open the app and tap the QR code icon next to the settings icon in the top-right-hand corner of the screen. Scan the QR code generated above (or alternatively, to see the error screen, scan any other QR code). You should see a screen saying that the scan was successful (or that it failed, if you scanned an invalid code).
- Tap the "home" button on this screen to return to the main screen. Now tap the settings icon and export your location history. Examine this file. You should see an entry with coordinates corresponding to the location you used when generating the QR code.


app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+};
+
+const StateIcon = ({ status, size }) => {
+  let icon;
+  switch (status) {
+    case StateEnum.DEFAULT:
+      icon = null;
+      break;
+    case StateEnum.SCAN_SUCCESS:
+      icon = StateNoContact;
```
tremblerz, line 39: For now it is good but we might want to use different set of icons.

app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+                />
+              </View>
+            </View>
+          </View>
+        )}
+        {this.state.currentState === StateEnum.SCAN_IN_PROGRESS && (
+          <View style={styles.qrScanContainer}>
+            <QRCodeScanner
+              onRead={this.onScanSuccess.bind(this)}
+              flashMode={RNCamera.Constants.FlashMode.torch}
```
tremblerz, line 216: I would suggest having flash off by default. Can think of use-case where flashlight might not be a good idea.

tremblerz: Future ux improvement can add the ability to toggle the flashlight within the screen.

app/views/Main.js
```diff
@@ -48,6 +49,22 @@ const Main = () => {
     );
   };
 
+  const QRIcon = () => {
```
tremblerz, line 52: I would recommend shifting the button to the Settings screen instead. This allows providing some context in the subtitle about the feature as well and also we are keeping all standalone features linkable through Settings screen.

summetj:
@tyleryasaka  - Great work, thanks!  I've invited you to a slack channel to discuss this feature, but having the implementation basically done makes the discussion a lot easier!

tyleryasaka:
@tremblerz @summetj Thanks for the feedback! I pushed changes to disable flash when scanning QR codes, and moved the QR code scanning button to the settings menu.

1. I agree that it would be simplest to merge and then iterate on the UI (I'm not a UI expert), as long as the camera permissions issue can be taken care of (as discussed in Slack, we don't want to request camera permissions in the release version which doesn't have this feature).
2. I like this idea of using abstract identifiers rather than GPS coordinates, and that's essentially what my original project is doing. I went with the GPS coordinates for now because that seemed easiest to integrate with what's already been built. But, adding this feature sounds great to me.

Let me know any further changes I can make to this PR.

android/app/build.gradle
```diff
@@ -202,6 +204,8 @@ android {
 }
 
 dependencies {
+    implementation project(':react-native-permissions')
```
Patrick-Erichsen, line 207: I'm a little confused as to why we need to include this dependency - how are we currently requesting permissions through `react-native-permissions` without having this setup?

app/Entry.js
```diff
@@ -42,8 +43,15 @@ class Entry extends Component {
   }
 
   render() {
+    const linking = {
```
Patrick-Erichsen, line 46: Could we pull this out of the render method?

app/locales/en.json
```diff
@@ -122,7 +122,16 @@
     "skip_this_step": "Skip this step",
     "team": "Team",
     "team_para": "Our team is composed of a consortium of epidemiologists, engineers, data scientists, digital privacy evangelists, professors and researchers from reputable institutions, including: MIT, Harvard, The Mayo Clinic, TripleBlind, EyeNetra, Ernst & Young and Link Ventures.",
-    "terms_of_use": "Terms of use"
+    "terms_of_use": "Terms of use",
+    "qr_exit": "Home",
```
Patrick-Erichsen, line 126: Could we pull all of these out into a `qr: {}` key? And then just have something like 
```
"qr": {
   exit: "Home", 
   ... 
}
```

app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+      break;
+  }
+  return (
+    <SvgXml xml={icon} width={size ? size : 80} height={size ? size : 80} />
+  );
+};
+
+const height = Dimensions.get('window').height;
+
+class QRScan extends Component {
```
Patrick-Erichsen, line 55: Could we convert this to a function component with hooks? The app is kind of split between class and function components at the moment, but I think the long term goal is to move towards function components.

app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+        longitude: Number(longitude),
+        time: Date.now(),
+      });
+      this.setState({ currentState: StateEnum.SCAN_SUCCESS });
+    } else {
+      this.setState({ currentState: StateEnum.SCAN_FAIL });
+    }
+  }
+
+  isValidCoordinates(latitude, longitude) {
```
Patrick-Erichsen, line 134: feels like we can pull this out into a helper method. or maybe just dump it in `LocationService.js`

android/app/build.gradle
tyleryasaka: I am not sure about your current setup. I was just following the installation [instructions for react-native-qrcode-scanner](https://github.com/moaazsidat/react-native-qrcode-scanner#to-install-and-start-using-react-native-qrcode-scanner). This code was added automatically by running `react-native link react-native-permissions`.

app/views/QRScan.js
tyleryasaka: Seems appropriate as a helper method, so I created a new `Location` helper.

app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+    );
+  }
+
+  componentWillUnmount() {
+    BackHandler.removeEventListener('hardwareBackPress', this.handleBackPress);
+    this.unsubscribeFocus();
+    this.unsubscribeState();
+  }
+
+  onNavigate() {
```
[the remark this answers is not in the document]
tyleryasaka: It's called anytime the component is "navigated" to, or probably better terminology is brought into focus. Needs to handle both the case of external deep linking, as well as the case where the user decides to scan through the app. Renamed to `onFocus`.

app/views/QRScan.js
```diff
@@ -0,0 +1,291 @@
...
+    if (this.props.route && this.props.route.params) {
+      const { latitude, longitude } = this.props.route.params;
+      if (typeof latitude !== 'undefined' && typeof longitude !== 'undefined') {
+        this.saveCoordinates(latitude, longitude);
+      } else {
+        this.setState({ currentState: StateEnum.SCAN_IN_PROGRESS });
```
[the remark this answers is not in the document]
tyleryasaka: Yes, this is just checking whether or not latitude and longitude are part of the route params. They are in the route params if and only if this screen has been brought into focus through deep linking. Otherwise, it means we got here because the user is trying to scan through the app, so we set to `SCAN_IN_PROGRESS` which renders the QR code scanner.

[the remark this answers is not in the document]
tyleryasaka: I've now refactored to hopefully make this clearer:

```js
  const onFocus = () => {
    const { latitude, longitude } = getLatLonFromRouteParams();
    const isDeepLink = typeof latitude !== 'undefined' && typeof longitude !== 'undefined'
    if (isDeepLink) {
      const savedSuccessfully = saveCoordinates(latitude, longitude);
      if (savedSuccessfully) {
        setcurrentState(StateEnum.SCAN_SUCCESS);
      } else {
        setcurrentState(StateEnum.SCAN_FAIL);
      }
    } else {
      setcurrentState(StateEnum.SCAN_IN_PROGRESS);
    }
  };
```

tyleryasaka:
Thanks for the review @Patrick-Erichsen . I've addressed your comments and pushed changes.

android/app/build.gradle
Patrick-Erichsen: Interesting. This looks right to me, I'm more just confused by the fact that we already have `react-native-permissions` as an NPM dependency, and we're already getting location permissions from users without this line. 

Moreso just confused as to how this was working for Location permissions without this line 🤷‍♂️

app/helpers/__tests__/Location.spec.js
```diff
@@ -0,0 +1,42 @@
+import { isValidCoordinates, parseQRCodeUrl } from '../Location';
+
+describe('isValidCoordinates', () => {
+  it('Should only validate strings', () => {
+    expect(isValidCoordinates(40.7128, -74.006)).toBe(false);
```
Patrick-Erichsen, line 5: awesome 👍

app/views/QRScan.js
```diff
@@ -0,0 +1,117 @@
...
+import { styles } from './QRScan/style';
+
+export const QRScanScreen = ({ navigation, route }) => {
+  const [currentState, setcurrentState] = useState(StateEnum.DEFAULT);
+
+  const getLatLonFromRouteParams = () => {
+    const params = route && route.params;
+    const latitude = (params && params.latitude) || undefined;
+    const longitude = (params && params.longitude) || undefined;
+    console.log('from route', latitude, longitude);
```
Patrick-Erichsen, line 20: this can be removed

app/views/QRScan.js
```diff
@@ -0,0 +1,117 @@
...
+  return (
+    <Theme use='violet'>
+      <ImageBackground source={BackgroundImage} style={styles.backgroundImage}>
+        <StatusBar
+          barStyle='light-content'
+          backgroundColor='transparent'
+          translucent
+        />
+        {isScanComplete && (
+          <ScanComplete currentState={currentState} onExit={exitQRScan} />
```
Patrick-Erichsen, line 106: beautiful

app/views/QRScan.js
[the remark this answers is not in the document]
Patrick-Erichsen: Cool, that makes sense and looks like it should always be in the right state. 👍

app/views/QRScan.js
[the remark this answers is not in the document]
Patrick-Erichsen: Cool - I'm pretty new to mobile dev so that might be my own ignorance of common terminology as well.

app/views/QRScan.js
Patrick-Erichsen: awesome 👍

Patrick-Erichsen:
Follow up question - are you testing the in-app scanning on an actual device?

tyleryasaka:
> Follow up question - are you testing the in-app scanning on an actual device?

Yes, I've been testing on my Google Pixel 2. I test what I can in the iOS simulator, but I was not able to figure out how to build it for my actual iPhone device.

tyleryasaka:
Ok, conflicts have now been resolved. I'm just saving the location now by calling `NativeModules.SecureStorageManager.importGoogleLocations`, which works perfectly well. The only confusing part is that the method name sounds specific to Google imports. I would suggest renaming that method to be generic at a later point, as it does not do anything that is specific to the Google imports as far as I can tell. But I didn't want to touch the native code in this PR.

Patrick-Erichsen:
Closing this for the time being as we rope in UI/UX resources.

PR #780, opened 2020-05-08
Hide source from dropdown if already selected

Shabanisimo:
https://pathcheck.atlassian.net/browse/SAF-195
photo_2020-05-08_12-33-57
photo_2020-05-08_12-34-00


app/views/ChooseProvider.js
```diff
@@ -367,16 +368,19 @@ class ChooseProviderScreen extends Component {
...
-                      key={key}
-                      onSelect={() => {
-                        this.addAuthorityToState(name);
-                      }}
-                      disabled={this.state.authoritiesList.length === 1}>
-                      <Typography style={styles.menuOptionText} use={'body2'}>
-                        {name}
-                      </Typography>
-                    </MenuOption>
+                    _.findIndex(this.state.selectedAuthorities, ['key', name]) <
```
Patrick-Erichsen: So - this whole section was pretty messy before you arrived - but do you think you could pull it out into a new component, something like `<AuthoritiesMenu />`?  Basically lines 348 - 397. Could also have a few other distinct components in there, e.g. `<GPSFilter />`, `<AuthoritiesMenuOption />`

I know that it's a little out-of-scope with the original ticket, but it feels like a good opportunity to get this cleaned up.

Patrick-Erichsen: Also - could you pull the `_.findIndex()` method into a separate method, something like `isAuthoritySelected()`? It's a little hard to understand what that code is doing at a quick glance.

app/views/ChooseProvider.js
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash';
```
tstirrat: Use the function import instead of the whole library:

`import findIndex from 'lodash/findIndex';`

app/views/ChooseProvider.js
Shabanisimo: > Also - could you pull the `_.findIndex()` method into a separate method, something like `isAuthoritySelected()`? It's a little hard to understand what that code is doing at a quick glance.
Okey, I'll update this component.

Patrick-Erichsen:
Hye @Shabanisimo , thanks for doing some of that cleanup work! Please fix the merge conflict and this is good to go.

app/components/TrustedSourceList.js
```diff
@@ -0,0 +1,88 @@
+import findIndex from 'lodash/findIndex';
+import React from 'react';
+import { StyleSheet, Switch, TouchableOpacity } from 'react-native';
+import { MenuOption } from 'react-native-popup-menu';
+
+import { Typography } from '../components/Typography';
+import Colors from '../constants/colors';
+import languages from '../locales/languages';
+
+export default function TrustedSourceList({
```
tstirrat: use a named export, not default.

app/views/LocationTracking.js
```diff
@@ -506,7 +506,9 @@ const styles = StyleSheet.create({
     top: '-13%',
     left: 0,
     right: 0,
+    flex: 1,
```
tstirrat, line 552: This looks like an unrelated change, what does it do? Perhaps you can revert this change if it's not related.

app/components/TrustedSourceList.js
```diff
@@ -0,0 +1,88 @@
...
+  selectedAuthorities,
+  authoritiesList,
+  isAuthorityFilterActive,
+  toggleFilterAuthoritesByGPSHistory,
+  addAuthorityToState,
+  filterAuthoritesByGPSHistory,
+  onMenuClick,
+}) => {
+  const isAuthoritySelected = name => {
+    return findIndex(selectedAuthorities, ['key', name]) < 0;
```
tstirrat, line 22:
```suggestion
    return findIndex(selectedAuthorities, {key: name}) < 0;
```

This is a little more clear

app/components/TrustedSourceList.js
```diff
@@ -0,0 +1,88 @@
...
+          </Typography>
+          <Switch
+            onValueChange={val => filterAuthoritesByGPSHistory({ val })}
+            value={isAuthorityFilterActive}
+          />
+        </TouchableOpacity>
+      )}
+      {authoritiesList === undefined
+        ? null
+        : authoritiesList.map(item => {
```
tstirrat: There's a lot goin on in this block. Hard to follow. 

This is `.filter()` operation:

```js
const availableAuthorities = authoritiesList.filter(item => !isAuthoritySelected(item));
```

do that above, and then it makes this much clearer:

```js
{availableAuthorities && availableAuthorities.map(item => <MenuOption ... />)}
```

app/components/TrustedSourceList.js
```diff
@@ -0,0 +1,88 @@
+import findIndex from 'lodash/findIndex';
+import React from 'react';
+import { StyleSheet, Switch, TouchableOpacity } from 'react-native';
+import { MenuOption } from 'react-native-popup-menu';
+
+import { Typography } from '../components/Typography';
+import Colors from '../constants/colors';
+import languages from '../locales/languages';
```
tstirrat: use `const {t} = useTranslation()` hook instead

app/components/TrustedSourceList.js
tstirrat: I realise due to the format of the authorities list, this will not work :) never mind.

Shabanisimo:
Updated. I also skipped it right away, but now I fixed the bug with the adding of custom links.

app/components/TrustedSourceList.js
```diff
@@ -16,9 +16,14 @@ export const TrustedSourceList = ({
   filterAuthoritesByGPSHistory,
   onMenuClick,
 }) => {
-  const isAuthoritySelected = name => {
+  const { t } = useTranslation();
+  const isAuthoritySelected = item => {
```
tstirrat, line 20: `isAuthoritySelected` seems to be the opposite of the name..

It returns true if the authority is NOT found in the list.

Either change name to `isAuthorityAvailable`, or change the return value to be `return findIndex(...) >= 0`

app/components/TrustedSourceList.js
```diff
@@ -27,41 +32,35 @@ export const TrustedSourceList = ({
...
           />
         </TouchableOpacity>
       )}
-      {authoritiesList === undefined
-        ? null
-        : authoritiesList.map(item => {
-            let name = Object.keys(item)[0];
-            let key = authoritiesList.indexOf(item);
+      {availableAuthorities &&
+        availableAuthorities.map(item => {
```
tstirrat, line 44: create a unit test for this to show that it renders the correct lists depending on availableAuthorities and selectedAuthorities.

PR #781, opened 2020-05-08
'Red icon flicker issue'

arvind-source:
#### Description:
I have removed the upgrade version of Gradle.


tstirrat:
I suspect this is fixed by the new location status checks in #749. Can you verify?

tstirrat:
I verified this was fixed in the refactor

PR #783, opened 2020-05-09
removed redirect notices for EULA links

pauladam316:
#### Description:

Fixes issue where links within the EULA would trigger redirect notices. It also makes clicking links within the EULA launch in a separate web browser (without this you end up with github pages running inside the license screen of the app with no back functionality to return to the EULA).

#### Linked issues:

https://github.com/Path-Check/covid-safe-paths/issues/739

#### Screenshots:

No visual changes

#### How to test:

In the EULA screen, click any of the two active links on the page. They should launch in your phones default web browser and not give you a redirect notice

### DIFF
--- a/app/components/EulaModal.js
+++ b/app/components/EulaModal.js
@@ -27,12 +27,11 @@ export const EulaModal = ({ selectedLocale, continueFunction }) => {
   const eulaPath = EULA_FILES[selectedLocale] || en;
 
   // Any links inside the EULA should launch a separate browser otherwise you can get stuck inside the app
-  shouldStartLoadWithRequestHandler = (webViewState) => {
+  const shouldStartLoadWithRequestHandler = webViewState => {
     if (webViewState.url != 'about:blank') {
       Linking.openURL(webViewState.url);
       return false;
-    }
-    else return true;
+    } else return true;
   };
 
   // Load the EULA from disk
@@ -63,7 +62,15 @@ export const EulaModal = ({ selectedLocale, continueFunction }) => {
                   accessibilityLabel='Close'
                   onPress={() => setModalVisibility(false)}
                 />
-                {html && <WebView style={{ flex: 1,}} source={{ html }} onShouldStartLoadWithRequest={this.shouldStartLoadWithRequestHandler.bind(this)}/>}
+                {html && (
+                  <WebView
+                    style={{ flex: 1 }}
+                    source={{ html }}
+                    onShouldStartLoadWithRequest={
+                      shouldStartLoadWithRequestHandler
+                    }
+                  />
+                )}
               </View>
             </SafeAreaView>
           </Theme>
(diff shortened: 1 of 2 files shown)


app/locales/eula/en.html
```diff
@@ -98,7 +98,7 @@
       <span class="c8"
         ><a
           class="c11"
-          href="https://www.google.com/url?q=https://github.com/tripleblindmarket/safe-places&amp;sa=D&amp;ust=1587722145899000"
```
Patrick-Erichsen, line 101: @tstirrat  Any idea why we were using a redirect link here? Assuming that it's fine to just use the direct link like @pauladam316 is doing here?

app/components/EulaModal.js
```diff
@@ -26,6 +26,15 @@ export const EulaModal = ({ selectedLocale, continueFunction }) => {
   // Pull the EULA in the correct language, with en as fallback
   const eulaPath = EULA_FILES[selectedLocale] || en;
 
+  // Any links inside the EULA should launch a separate browser otherwise you can get stuck inside the app
+  shouldStartLoadWithRequestHandler = (webViewState) => {
+    if (webViewState.url != 'about:blank') {
```
Patrick-Erichsen, line 31: Could you pull `about:blank` out into a variable with a name along the lines of `const DEFAULT_EULA_URL = 'about:blank'` ? I had no idea that was what that string was until you explained to me so probably would be useful to other folks as well.

Also in general - best practice to do `!==` over `!=` just to be sure we're [forcing strict equality comparison.](https://codeburst.io/javascript-double-equals-vs-triple-equals-61d4ce5a121a)

app/components/EulaModal.js
```diff
@@ -26,6 +26,15 @@ export const EulaModal = ({ selectedLocale, continueFunction }) => {
...
+      Linking.openURL(webViewState.url);
+      return false;
```
Patrick-Erichsen, line 33: Same thing with the boolean values here - kind of hard to understand without looking at the documentation. What do you think about something like this?

```js
const shouldStartLoadWithRequestHandler = webViewState => {
    let shouldLoadRequest = true;

    if (webViewState.url != 'about:blank') {
      Linking.openURL(webViewState.url);

      // Opening the request in the users default browser, so cancel the WebView load
      shouldLoadRequest = false;
    }

    return shouldLoadRequest;
  };
```

app/components/EulaModal.js
Patrick-Erichsen: One other note - I would put this at the module level, e.g. 

```js
const DEFAULT_EULA_URL = 'about:blank'
export const EulaModal = ({ selectedLocale, continueFunction }) => {
```

app/locales/eula/en.html
tstirrat: It was a save from Google doc to html. I believe that was added by Google and we don't need or want it.

PR #788, opened 2020-05-09
Securely Store Location Data + Other improvements

efalkner:
Details are copy 🍝 from https://github.com/Path-Check/covid-safe-paths/pull/747

Securely store locations in encrypted Realm db

This PR securely stores all captured Geolocation coordinates in an encrypted Realm database.

Secures Locations imported via Google
Migrates old locations to new secure storage -> deletes old on success
Still allows for json export
Fixes bug where locations stopped being captured after device restart
Linked issues:
#527
Fixes safe-paths-contrib#29

How to test
For debug builds, you can download the Realm database inside the apps storage. You can open the database using Realm Studio: https://realm.io/products/realm-studio/. Because the database is encrypted, you will need to log a hex-encoded encryption key.

Closes #747


jest/config.js
```diff
@@ -2,5 +2,9 @@ module.exports = {
   preset: 'react-native',
   setupFiles: ['./jest/setupFile.js'],
   rootDir: '../',
-  testPathIgnorePatterns: ['e2e', 'node_modules'],
+  testPathIgnorePatterns: [
+    'e2e',
+    'node_modules',
+    'node_modules/(?!(@react-native-community|react-native)/)',
```
tstirrat, line 8: Oooh this might be the fix we need for `testing-library/react-native`

PR #790, opened 2020-05-10
Language latest from may 10

alpita-masurkar:
#### Description:
The latest pull from Lokalise (May 10)


app/locales/es.json
```diff
@@ -1,5 +1,13 @@
 {
-  "_display_name": "Español",
+  "_display_name": "Inglés",
```
tstirrat, line 2: This needs fixing in Lokalise. It should read "Spanish" not "English"

PR #663, opened 2020-04-24
Feature/filter feedback for url created

Mr-EZG:
Changes were related to GitHub issue #456. The addresses the issue of not filtering or verifying the text used to represent a url for some health authority which could lead to potential server attacks via not regulating the maximum amount of characters that could be used to represent a URL. This implementation limits the character input used to represent a URL to 8,000. This number was chosen as to allow long urls, but still help prevent the opportunity of a server attack due to an extremely long String (See conversation on GitHub issue #456). Another filter was added as to verify the format of the String in that it followed a valid URL format. This was done as an attempt to help prevent hidden characters for server attacks and help increase the likelyhood that the user had inputted a url that is more likely to point towards a valid, live website. If the String sent to be considered as a URL fails any of the two aforementioned conditions, then the URL will not be saved and the user will be notified for which condition the have failed. As a more convenient bonus, the new feature trims leading and trailing whitespaces. It should also be noted that this current implementation also trims any attempted URL to the first 8,000 characters in the event that the input would exceed 8,000 characters.

NOTE: new variables were added to the en.json file. These included "invalid_entry_message", "please_try_again_message", "proceed_message", and "entry_already_used". For more scalable use, the translation of these variables should be included. It should also be noted that I had to ignore the lint issue since it comes from the regex expression. Some escape characters might not be needed, but I this version includes a working regex as is.

Does not allow the user to input an invalid URL and limits the overall length of potential URLs to be considered to 8,000 characters.

#### Linked issues:

GitHub issue #456

This fixes the character limit and adds an additional feature as to help prevent the insertion of hidden characters for injection attacks via a regex pattern. The pattern checks that only certain characters are used and that the overall format matches how urls are defined with slight flexibility with capitolization to match the flexibility of a web browser (Such as Www.google.com is not technically valid since no domain name can have capitol letters, but a browser would still point to www.google.com) This was an important feature as keyboards on phones begin with with typing the first letter as capitalized.

#### How to test

NOTE: due to personal computer restrictions, this was only tested on an Android Emulator. Once able to recreate the results, please do see if they could be recreated on an IOS emulator.

Home-screen -> Dashboard -> Choose Health Authority -> Add trusted Source -> Add authority via URL,

Then:

Attempt to submit a url that does not meet URL format protocol. (NOTE: this does not mean that the url points to a live website, but that the attempted URL is valid and could point to live website)
Attempt to submit a url with more than 8,000 characters
[...]

Issue #456, opened 2020-04-11
BUG: Zero policing of what can be entered as a URL for a Health Authority - DHM-S1-18


diarmidmackenzie:
There is apparently no policing of what text can be entered as a URL.  I was able to paste 1000s of lines of text here.
o   I haven’t explored this, but if there is no policing here, I’d expect it may be possible to attack the app here using SQL injections, very long text etc.

For security reasons, it would seem prudent to put some limits on what can be entered here.

denispapakul:
I can fix it, actually, I need to know max length or URL, which can be entered for a Health Authority. Please, assign me to this ticket.

diarmidmackenzie:
The correct answer for "what is the maximum length of a URL" is complicated:

https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

Key limits are:
- IE/Edge won't allow more than 2000 chars in the address bar.  But most other browsers will. (I can envisage wanting to put a URL in a browser for troubleshooting reasons).
- While RFCs don't set any actual limit, they advise that systems SHOULD support up to 8000 chars.

I think 2000 chars is probably enough.  But there is a risk I'm wrong, and that there is a reason why a URL might need to be longer than that.

For me the most important thing is to have some limit (so that we don't risk a buffer overrun bug).  

So I suggest we take the more conservative position of 8000 chars.

diarmidmackenzie:
I've been thinking about testing, and formed the view that if a tester (manual or automated) wants to test a scenario where the "Healthcare Authority" provides some specific, known, set of data points, a neat way to achieve that would be to encode the full required JSON data set in the configured URL, and have a web server that simply takes that request & serves it back up to the app.

Given this use case, having plenty of room in the URL would be quite useful.  I think 8000 chars is still plenty, but a good reason to go for 8000 rather than 2000.

E3V3A:
iOS and Android browser apps are all different and continuously updated, so there is not much point to comparing it to what your Windows Desktop browsers can handle.

Good practice would be to hard limit the URL length to something sensible, to ensure any browser can handle it. Speaking of, I think the URL is used in a POST request later, independent of browser, is it not? Or are you also talking about the links to documents and authority website info?

diarmidmackenzie:
This is about the limit for a Health Authority URL.  Currently there is no limit, which is an exposure.

I am proposing 8000 chars.

Do you propose a different value?  If so, what value & why?

[...]

PR #791, opened 2020-05-10
Feature/updated url filter

Mr-EZG:
This includes the proposed changes to "Feature/filter feedback for url created #663", which was my first attempt to push the following changes. (Adding a URL sanity check for when adding a new Health Authority such that the attempted saved url matches a similar regex pattern to a web browser and is no more than 8000 characters)

Note: I manually merged many files. I was careful and did my best to only keep newly added code within the app. The App still seemed to run well, but a comparison might want to be made to the original app. The changes of interest occurred in app/views/ChooseProvider.js and app/locales/en.json

Originally, Changes were related to GitHub issue #456. The addresses the issue of not filtering or verifying the text used to represent a url for some health authority which could lead to potential server attacks via not regulating the maximum amount of characters that could be used to represent a URL. This implementation limits the character input used to represent a URL to 8,000. This number was chosen as to allow long urls, but still help prevent the opportunity of a server attack due to an extremely long String (See conversation on GitHub issue #456). Another filter was added as to verify the format of the String in that it followed a valid URL format. This was done as an attempt to help prevent hidden characters for server attacks and help increase the likelyhood that the user had inputted a url that is more likely to point towards a valid, live website. If the String sent to be considered as a URL fails any of the two aforementioned conditions, then the URL will not be saved and the user will be notified for which condition the have failed. As a more convenient bonus, the new feature trims leading and trailing whitespaces. It should also be noted that this current implementation also trims any attempted URL to the first 8,000 characters in the event that the input would exceed 8,000 characters.

NOTE: new variables were added to the en.json file. These included "invalid_url _message" and a new "authorities" key for the son file. For more scalable use, the translation of these variables should be included. 

Does not allow the user to input an invalid URL and limits the overall length of potential URLs to be considered to 8,000 characters.

Linked issues:

GitHub issue #456

The regex pattern checks that only certain patterns are used and that the overall format matches how urls are defined with slight flexibility with capitalization to match the flexibility of a web browser (Such as Www.google.com is not technically valid since no domain name can have capitol letters, but a browser would still point to www.google.com) This was an important feature as keyboards on phones begin with with typing the first letter as capitalized.

How to test

NOTE: due to personal computer restrictions, this was only tested on an Android Emulator. Once able to recreate the results, please do see if they could be recreated on an IOS emulator.

Home-screen -> Dashboard -> Choose Health Authority -> Add trusted Source -> Add authority via URL,

Then:

Attempt to submit a url that does not meet URL format protocol. (NOTE: this does not mean that the url points to a live website, but that the attempted URL is valid and could point to live website)
Attempt to submit a url with more than 8,000 characters
Attempt to submit a url with insert hidden characters into the text-box (as a potential consequence, so may not fully work or be true)

Try to test inputting valid and non-valid urls with edge cases included (such as an empty String)
Then if see if no error pops up for the User if it does not match the aforementioned properties. The pop up messages should be designed as to force the user to notice that their url is in an unacceptable format. There should also be log messages to help verify the above.


Patrick-Erichsen:
@Mr-EZG Please just tag me in a comment when these changes are incorporated so I know when to come back and review 👍

Patrick-Erichsen:
@Mr-EZG we wanted to make sure this work got merged in ASAP, so I branched off of your work here and create #813 .

That PR just incorporated the comments I had above (pulling out logic into `urlValidation.js`, writing unit tests, etc.)

Thank you for getting this done! Greatly appreciated.

PR #801, opened 2020-05-12
Fix: Add trusted source button overlaps "paste url" text box

### LINKED ISSUE #674: Choose health authority : Add trusted source button overlaps the "paste url" text box upon click
Choose health authority : Add trusted source button overlaps the "paste url" text box upon click

Issue type : UI/UX
Safe paths version : 1.0.0

## Version

- Type of phone: Android 
- OS Version: 9
- Model: Moto g 6 plus

## Steps to reproduce

1. Get onboarded.
2. Dashboard -> Choose health authority -> Add authority via URL 
3. Click on the text box, to paste the url

Issue : The "add trusted source " button overlaps the "paste your url" text box.

addauthority

siddhantbenadikar:
#### Description:
Fixes #674 

Notes: Ideally we should get rid of the "Add trusted source" button while adding a new url. While adding a new url the user shouldn't be given options to add more trusted sources. Right now the user can click that button and it presents him with another drop down.

#### Updates:
1. Added a keyboard avoiding view in commons. This new component can be used around other text boxes which potentially may have the same issue.
2. Cleaned up repeated code.

#### How to test:
1. Get onboarded
2. Dashboard -> Choose health auth -> Add auth via URL
3. Click on textbox to enter/paste url
4. Add multiple to see if the view scrolls.
Screen Shot 2020-05-13 at 11 54 33 PM


app/views/common/SPKeyboardAvoidingView.js
```diff
@@ -0,0 +1,29 @@
...
+  );
+};
+
+const styles = StyleSheet.create({
+  keyboardView: {
+    flex: 1,
+  },
+});
+
+export default SPKeyboardAvoidingView;
```
tstirrat, line 29: Use regular named exports

app/views/ChooseProvider.js
```diff
@@ -249,88 +262,56 @@ class ChooseProviderScreen extends Component {
...
+                    </TouchableOpacity>
+                  </View>
+                  <FlatList
+                    data={this.state.selectedAuthorities}
+                    renderItem={({ item }) => (
+                      <View style={styles.flatlistRowView}>
+                        <Typography style={styles.item} use={'body3'}>
+                          {item.key}
+                        </Typography>
+                        <TouchableOpacity
```
tstirrat, line 304: While you're here, can you convert this to an IconButton?

app/views/ChooseProvider.js
```diff
@@ -249,88 +262,56 @@ class ChooseProviderScreen extends Component {
...
+                          urlText: text,
+                        });
+                      }}
+                      value={this.state.urlText}
+                      autoFocus={this.state.urlEntryInProgress}
+                      style={[styles.item, styles.textInput]}
+                      placeholder={languages.t('label.enter_authority_url')}
+                      onSubmitEditing={this.addCustomUrlToState}
+                    />
+                    <TouchableOpacity onPress={this.addCustomUrlToState}>
```
tstirrat, line 293: This is almost an IconButton but with an image. Can you modify IconButton to include image={img} and use that?

app/views/common/SPKeyboardAvoidingView.js
```diff
@@ -0,0 +1,29 @@
...
+      autoScrollToFocusedInput
+      behavior={isiOS ? instanceBehavior : 'height'}>
+      {children}
+    </KeyboardAvoidingView>
+  );
+};
+
+const styles = StyleSheet.create({
+  keyboardView: {
+    flex: 1,
```
tstirrat, line 25: We probably don't always want flex 1 here so I suggest moving that to the outer component and passing through style.

Though if you think we always need flex 1 then put a comment why here.. But also, allow style to be merged in

app/views/ChooseProvider.js
siddhantbenadikar: IconButton seems to be a styled SvgXml, and it won't accept png images. So we might need to find svg alternative for these pngs or bring in another lib like react-native-vector-icons that does the transformation for you

app/views/common/SPKeyboardAvoidingView.js
siddhantbenadikar: since the logic here is that this component renders it's children. Leaving it to default flex will cause its children to shrink. Hence I kept it as 1 default. Without this 1 here the container will shrink and children won't display

siddhantbenadikar:
Also, I don't own a mac so developing on linux atm. And only can do android dev. If someone could verify this on ios

app/views/common/SPKeyboardAvoidingView.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { KeyboardAvoidingView, Platform, StyleSheet } from 'react-native';
+
+export const SPKeyboardAvoidingView = ({ behavior, children }) => {
```
Patrick-Erichsen, line 4: could we just name this `KeyboardAvoidingView` ?  We can import the RN version as `import {KeyboardAvoidingView  as RNKeyboardAvoidingView, ... }`

We aren't using an `SP` prefix for any other genric component wrappers (e.g. `<Button />, <Switch />`, etc.), so I feel like it adds a bit of confusion.

app/views/common/SPKeyboardAvoidingView.js
```diff
@@ -0,0 +1,27 @@
...
+  const isiOS = Platform.OS === 'ios';
```
Patrick-Erichsen, line 5: I think it would be useful to just define this function once and for all in `helpers/General.js`. We have a bunch of different functions in various places that all do the same check. 

Could you just define two `consts` in that file as part of this PR?

```js
const isAndroid = Platform.OS === 'android';
const isIOS = Platform.OS === 'ios'
```

app/views/ChooseProvider.js
```diff
@@ -151,30 +152,19 @@ class ChooseProviderScreen extends Component {
...
       urlEntryInProgress: false,
+      urlText: '',
     });
-  }
 
-  addCustomUrlToState() {
+  addCustomUrlToState = () => {
     let { urlText: url, selectedAuthorities } = this.state;
 
     if (!validUrl.isWebUri(url)) {
```
Patrick-Erichsen, line 170: So this is actually my fault - this should be 

```js
if (!validUrl.isWebUri(url) || this.hasExistingAuthorityWithUrl(url)) {
```

Looks like I accidentally removed that check when I included the `validUrl` library.

app/views/common/SPKeyboardAvoidingView.js
siddhantbenadikar: did some searching and looks like we already have those functions in Utils.js

app/views/ChooseProvider.js
```diff
@@ -151,33 +152,30 @@ class ChooseProviderScreen extends Component {
...
+      urlText: '',
     });
-  }
 
-  addCustomUrlToState() {
+  addCustomUrlToState = () => {
     let { urlText: url, selectedAuthorities } = this.state;
 
-    if (!validUrl.isWebUri(url)) {
+    if (!validUrl.isWebUri(url) || this.hasExistingAuthorityWithUrl(url)) {
```
Patrick-Erichsen, line 178: thanks for fixing my mistake! 😄

app/views/ChooseProvider.js
Patrick-Erichsen: future work.

PR #802, opened 2020-05-12
Feature/add fastlane

efalkner:
#### Description:
Add [Fastlane](https://docs.fastlane.tools/) to facilitate builds and pushes to play store / test flight
Add [Fastlane Match](https://docs.fastlane.tools/actions/match/#match) to manage iOS certificates
Provide workflows to build ios and android staging artifacts on push to develop
Provide workflow to build release ios and android artifacts to push to testflight / playstore beta track

Release workflow is configured to run on merge into develop, master or release branches. It will trigger a version bump task that will use a github action to change the package.json version, create a tag, and write a reference to CHANGELOG.md. iOS and Android versions will be incremented in their build tasks to match, then pushed to the play store / testflight.
Commit messages are scraped, using the following format: https://www.conventionalcommits.org/en/v1.0.0/, to determine if the version is patch, minor or major (with patch as the default version bump)


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,217 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          ref: ${{ github.ref }}
+
+      - name: "cat package.json"
+        run: cat ./package.json
+
+      - name: "Setup Node.js"
```
tstirrat, line 26: this node action often hangs so use our other cached strat from the other jobs

android/app/build.gradle
```diff
@@ -6,6 +6,7 @@ apply plugin: 'realm-android'
 
 project.ext.envConfigFiles = [
     debug: ".env",
+    staging: ".env.beta",
```
tstirrat, line 9: should this be `.env.staging` now since you added that file?

patches/@mauron85+react-native-background-geolocation+0.6.3.patch
```diff
@@ -532,3 +532,73 @@ index ad66b5c..0000000
 -
 -}
 \ No newline at end of file
+<<<<<<< HEAD
```
tstirrat, line 535: I cant tell if this is a merge conflict or part of the patch, but ether way this is the jetifier changes so it's not needed, rever.

android/app/build.gradle
efalkner: 😬 yup

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,212 @@
+name: Publish Release Candidates to Play Store and Testflight
```
tstirrat, line 1: this name is huge which clips the job name in the checks panel on PRs

.github/workflows/test.yml
```diff
@@ -135,15 +135,15 @@ jobs:
...
-      - name: Cache CocoaPods
-        uses: actions/cache@v1
-        with:
-          path: ios/Pods
-          key: ${{ runner.OS }}-pods-${{ hashFiles('**/Podfile.lock') }}
-          restore-keys: |
-            ${{ runner.OS }}-pods-
-
-      - run: pod install --repo-update
+      # - name: Cache CocoaPods
```
tstirrat, line 138: intentional? if so remove commented out code

PR #803, opened 2020-05-12
Start/stop location tracking if needed when checking status

### LINKED ISSUE #797: App no longer tracking locations in debug environment
## Version 
- Debug build version: develop

## Affected Devices
- All

## Affected Languages
- All

## Steps to reproduce

A recent refactor moved from the home screen being LocationTracking.js to Main.js. The Main.js screen never calls LocationService.start() so location tracking will never begin.

Another issue is that the code to migrate to our encrypted db is missing on Main.js.

```
GetStoreData(LOCATION_DATA, false).then(locations => {
      if (Array.isArray(locations) && locations.length > 0) {
        NativeModules.SecureStorageManager.migrateExistingLocations(
          locations,
        ).then(() => {
          RemoveStoreData(LOCATION_DATA);
        });
      }
    });
```

## Expected Behavior

- When the app is installed and location tracking permissions have been granted, the app should start tracking locations.

tstirrat:
This updates the new Main.js tracking code to optionally start and stop the location tracking service when checking status.

#### Linked issues:

Fixes #797

#### Screenshots:

NA

#### How to test:

- Check that the service starts after "All finished" onboarding, then and moving to the main screen.
- Check that the service stops when stopping from app permissions, app settings, and when disabling location tracking at the OS level.
- Check that the service starts up again after toggling in settings.


app/services/LocationService.js
```diff
@@ -105,278 +105,223 @@ export class LocationData {
...
+    }
 
-    BackgroundGeolocation.on('stop', () => {
-      PushNotification.localNotification({
-        title: languages.t('label.location_disabled_title'),
-        message: languages.t('label.location_disabled_message'),
-      });
-      console.log('[INFO] stop');
-    });
+    configureBackroundGeoService();
```
tstirrat, line 212: The diffs are a bit off, but I simply moved all the code here into a method. I did not change it at all.

It was clogging up the general overview of this method which is to configure once + start every time.

tstirrat: I removed this because it made the diff harder to parse

app/services/LocationService.js
```diff
@@ -117,31 +116,30 @@ export default class LocationServices {
 
     PushNotification.configure({
       // (required) Called when a remote or local notification is opened or received
-      onNotification: function(notification) {
+      onNotification(notification) {
```
efalkner, line 119: 😍

app/services/LocationService.js
```diff
@@ -117,31 +116,30 @@ export default class LocationServices {
...
         console.log('NOTIFICATION:', notification);
```
efalkner, line 120: :(

Patrick-Erichsen:
LGTM

PR #808, opened 2020-05-12
Auto close stale PRs after 1 week

tstirrat:
Uses https://github.com/probot/stale to close PRs after 1 week.


tstirrat:
Testing automation

PR #809, opened 2020-05-12
Run iOS unit tests automatically

tstirrat:
Run tests automatically

#### How to test:

Check for their passing. RIP 💀


tstirrat:
Passing test job here: https://github.com/Path-Check/covid-safe-paths/pull/809/checks?check_run_id=668920755

PR #826, opened 2020-05-14
override remote permissions when we're not using them

efalkner:
# Description:
This is a merge into develop to match #825 

react-native-notifications requires firebase integration for Android if you're configuring the notifications with permissions set to true. This work-around comes from their docs, to avoid crashing Android when you're not using firebase.

# Linked issues:
https://pathcheck.atlassian.net/browse/SAF-234

# Screenshots:
image

# How to test:
Run the app on Android.
Visit the Dashboard screen.
Disable & RE-Enable location in the Dashboard screen.
When I re-enabled location, I got a black screen with a Firebase error.
Expect that the app does not crash with the same red box error as shown above
Regression test on iOS

### DIFF
--- a/app/services/LocationService.js
+++ b/app/services/LocationService.js
@@ -121,6 +121,7 @@ export default class LocationServices {
         // required on iOS only (see fetchCompletionHandler docs: https://github.com/react-native-community/react-native-push-notification-ios)
         notification.finish(PushNotificationIOS.FetchResult.NoData);
       },
+      // Setting the permissions to true causes a crash on Android, because that configuration requires Firebase
       requestPermissions: Platform.OS === 'ios',
     });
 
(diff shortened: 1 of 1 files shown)


app/services/LocationService.js
```diff
@@ -121,7 +121,7 @@ export default class LocationServices {
         // required on iOS only (see fetchCompletionHandler docs: https://github.com/react-native-community/react-native-push-notification-ios)
         notification.finish(PushNotificationIOS.FetchResult.NoData);
       },
-      requestPermissions: true,
+      requestPermissions: Platform.OS === 'ios',
```
tstirrat, line 124: Add a comment that `true` causes a crash on Android.

app/services/LocationService.js
```diff
@@ -223,6 +223,8 @@ export default class LocationServices {
     console.log('[INFO] BackgroundGeolocation auth status: ' + authorization);
 
     BackgroundGeolocation.start(); //triggers start on start event
+    // Setting the permissions to true causes a crash on Android, because that configuration requires Firebase
```
Patrick-Erichsen, line 226: Was this the comment that was supposed to be up on L124?

PR #835, opened 2020-05-14
Fix final eslint warnings, install better lint checker

tstirrat:
- Use a eslint action that places annotations inline in the PR
- Set a rule for prefering functional components
- Fix some lint errors

#### Linked issues:

NA

#### Screenshots:

Example when lint fails:

Screen Shot 2020-05-14 at 3 37 24 PM

#### How to test:

lint should pass


Patrick-Erichsen:
Sweet, this is a cool plugin! I'm re-running the builds to hopefully get iOS to pass and then will merge.

PR #836, opened 2020-05-14
Create workspaces for gps, bte, and shared.

mattThousand:
#### Description:

Creates workspaces for `gps`, `bte`, and `shared`

PR #847, opened 2020-05-17
Lokalise latest may 17

alpita-masurkar:
#### Description:
Latest pull from May 17 after the move to gps/app/locales
Because we have haitian creole still present in Lokalise, it gets pulled in. Let me know if we need to cherrypick and remove it before we create a PR. @tstirrat


tstirrat:
Closing this because the file paths were changed in #836  and pull script fixed in #852

PR #849, opened 2020-05-18
Update location notification

Shabanisimo:
Added check if geolocation tracking is turned off and the user navigates to the app that the notification is triggered once a certain period of time


Patrick-Erichsen:
Hey @Shabanisimo - is this ready for review? If so, can you provide a description?

Shabanisimo:
Description updated.

Patrick-Erichsen:
So - I believe this will prevent users from receiving a notification every time that they turn location tracking off. Is that correct? 

If so, I'm not sure that's the behavior we want. I believe our intention is to push an alert every time that location tracking is stopped.

Shabanisimo:
But with such frequent notifications, this can get annoying users.

PR #853, opened 2020-05-18
Create whitelabel builds for bluetooth enabled version of the app.

johnschoeman:
This pr is a handful of commits that work towards introducing a white labeling strategy to build the application for either location based or Bluetooth based contact tracing strategies.  Given that one app cannot have both bluetooth and location based data for privacy concerns we need to be able to build the app binary with different permission settings in info.plist, in order to do this we introduce two build targets in xcode, GPS and BTE, that have different info.plist and bundle ids, but share all of the other native libraries and code. We exposes the tracing strategy that the given build is using to the js side over the bridge via a 'getTracingStrategy' function.

Ive left the commits separate for ease of review and i would recommend reviewing one at a time.

Note that this work only sets up the distinction between the gps and bte based builds. It does not implement any bluetooth permission work, it will just give us some space to do so.

### Running GPS

```
yarn run-ios-gps
```

gps

### Running BTE

```
yarn run-ios-bte
```

bte

## Commit 1:

41fec15
Why:

This commit:

----

This commit is perhaps not necessary, but while getting familiar with the app I ended up using this clean up as a way to build some context, I was also concerned about missing some path updates when moving the assets folder out of shared in the following commit and this refactor made it less likely for me to make a mistake.

## Commit 2

Introduce gps and bte ios targets …
67ade64
Why:
We would like to introduce bluetooth to the application for bluetooth
based contact tracing. Given the privacy concerns of having both
location and bluetooth data available to the same application, google
and apple do not allow for both location and bluetooth to be available
at the same time.

Hence in order to service both location and bluetooth
we will need to provide separate apps with separate info.plists and
bundle identifiers. At the same time we would like to share a much
application logic between the two applications as to reduce the
maintenance and organizational overhead. Ideally as close to 100% of the
javascript layer is common and for the ios layer we only need to change
the info.plists.

This commit:

Introduces two build targets, GPS and BTE, to the xcode project which
are setup to read from the corresponding groups for config and info.plist
A COVIDSafePathsConfig file was introduced to expose a
'getTracingStrategy' function to the react native side so that the js
runtime can determine which tracing strategy the bundle will be using.
This will allow us flexibility to programmatically determine which
tracing strategy is being used either within a given react component or
to conditional render components altogether. This also allows use to
maximize the amount of shared native code.

To build the corresponding targets via the react native cli, you can
pass the --scheme flag with the corresponding target:

```
react-native run-ios --scheme "GPS"
react-native run-ios --scheme "BTE"
```

We updated the bundle ids to `org.pathcheck.covid-safepaths.gps` and
`org.pathcheck.covid-safepaths.bte` respectively. Depending on app store
considerations, it might be easier to revert
`org.pathcheck.covid-safepaths.gps` to `org.pathcheck.covid-safepaths`

Note that theres a fair amount of location information storing logic
currently in AppDelegate, a future commit might want to refactor this so
that the gps specific code is only built in the GPS target builds.

## Commit 3

ollback js code spliting …
d61cf20
Why:
javascript side.

This commit:

## Commit 4

Refactor onboarding5 to use functional component …
4b982d0
Why:
We would like to introduce some information to the component via a
context. It is best practice to use hooks and so we would like to use
the useContext hook for this. Currently the onboarding5 component is
build with the class syntax keeping us from doing this.

This commit:
Refactors the Onbarding5 component to use the functional style. No new
functionality has been added.

## Commit 5

Introduce covid safe paths config …
3c90fba
Why:
We would like to be able to conditionally render react components based
off of the tracing strategy being used by the app. Given that the app
can be built with multiple ios targets we need to fetch this information
from over the bridge.

This commit:
Introduces the COVIDSafePathsConfig to the app, which uses the
NativeModules api to fetch the tracing strategy and stores it in a
config object to be used as needed in components.

Note that this commit only introduces the pattern to a handful of
existing components. It does not exhaustively update every spot that
needs to render conditionally based off of the tracing strategy.

At first we considered using a context for this, we rolled that back
since it seemed to be a bit overkill at this point. It might make sense
to reintroduce in a future commit if there is significant application
logic that needs to be run and shared amongst many components related to
this.

## Commit 5

Update readme, package.json for bte/gps targets …
5ccfb95
This commit updates the readme and package.json to include the info for
building to either the gps or bte target for ios

## How to test:

Run the apps locally with the two schemes.
```
yarn run-ios-gps
```

```
yarn run-ios-bte
```

Remaining work to be done:

- do a similar set of work for the android side

- verify that this setup still works with the fastlane setup and if not adjusting to do so. 
That is, I wasnt able to verify that the production builds are still working, for sake of time. It might makes sense it introduce build targets for development and staging for each of the flavors of the app.

- set up translations for bte / gps
This one could probably use some thought. Something like : t('label.my_text', 'bte') or t('label.my_text', 'gps') might makes sense. We potentially have a combinatoric problem of two app flavors for every language and setting up the translations in a maintainable way will be important.

- consider moving the swift location storage code out of AppDelegate and into the gps group so that it is only included in the gps target build.


tstirrat (changes requested):
There is no way I'm going to be able to track all these changes while minimizing breakage risk..

It looks like you have a chain of things that you want to do in multiple steps, consider using [git-multi-pr](https://github.com/pingpongboss/git-multi-pr) for this, so that you can sync with `develop` and maintain multiple small prs.

It seems to match your workflow of "a single commit per logical change". 

I've used `multi-pr` once before. It assumes `master` is your base branch, but you can edit a single file to change `BASE=master` to `BASE=develop` and everything else works as intended.

johnschoeman:
> There is no way I'm going to be able to track all these changes while minimizing breakage risk..
> 
> It looks like you have a chain of things that you want to do in multiple steps, consider using [git-multi-pr](https://github.com/pingpongboss/git-multi-pr) for this, so that you can sync with `develop` and maintain multiple small prs.
> 
> It seems to match your workflow of "a single commit per logical change".
> 
> I've used `multi-pr` once before. It assumes `master` is your base branch, but you can edit a single file to change `BASE=master` to `BASE=develop` and everything else works as intended.

Yeah theres a lot here, and it has started to grow so that we can hit our deadlines for the bte entitlement submission.

I think it makes sense to break up into multiple prs and introduce incrementally, not sure if we need to use multi-pr, or if we should just open new prs as they make sense like with: https://github.com/Path-Check/covid-safe-paths/pull/864

So potentially this pr will become unnecessary, but I'll leave it open for now until the dust is settled for todays deadline.

johnschoeman:
Closing this pr since there are things that got in to this branch which shouldnt be merged to development without some adjustment. (And it has also gotten much larger over the past few days.) Well open these changes up more incrementally, but continue working on bte as makes sense, if that sounds good to you, @mattThousand , @tparesi

PR #854, opened 2020-05-18
Implement getTracingStrategy in Android native module.

mattThousand:
This pull request implements `getTracingStrategy` inside a small Android native module.


johnschoeman (commented):
LGTM!

Also i tend to not actually approve or request changes on prs, as i dont believe in them, just as a heads up!  let me know if i should be actually approving things.

PR #855, opened 2020-05-18
Incorrect location indicator in settings page due to disabled location permission.

shilpashendre:
Location indicator is set properly in the settings page.


Patrick-Erichsen (changes requested):
Hey @shilpashendre , is there any relevant Github issue you could link on this? I don't quite understand what the fix is for.

Can you also provide testing instructions to verify that the fix is working correctly?

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+      } else {
+        // when user deny
+        setIsLogging(false);
+        LocationServices.stop();
+        await SetStoreData(PARTICIPATE, false);
+      }
+    });
+  };
+
+  // checking whether location permission is enable or disbale from application setting on load of setting's page.
```
Patrick-Erichsen:
```suggestion
  // checking whether location permission is enabled or disabled from application setting on a load of setting's page.
```

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+  const checkIsLocationEnable = () => {
```
Patrick-Erichsen:
```suggestion
  const isLocationTrackingEnabled = () => {
```

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+    LocationService.checkStatus()
```
Patrick-Erichsen: Please use the `async`/`await` pattern here.

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+      .then(async res => {
+        if (res !== undefined) {
```
Patrick-Erichsen: this line can be removed and the next line can just be 
```js
if (res.isRunning) {
```

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+          if (res.isRunning === true) {
+            await SetStoreData(PARTICIPATE, true);
+            setIsLogging(true);
+          }
+        }
+      })
+      .catch(err => {
+        console.log('TCL: SettingsScreen -> err', err);
```
Patrick-Erichsen: What is `TCL`?

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
     };
   }, [navigation]);
 
+  //  This method is used to track whether user has "Allow" or "Deny" location permission system Dialog.
+  const systemDialogPermission = async () => {
+    BackgroundGeolocation.on('authorization', async status => {
```
Patrick-Erichsen: Please use `async`/`await` pattern here.

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+      if (status === BackgroundGeolocation.AUTHORIZED) {
+        // when user allow
```
Patrick-Erichsen: this comment and the one in the `else` branch can be removed.

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
     };
   }, [navigation]);
 
+  //  This method is used to track whether user has "Allow" or "Deny" location permission system Dialog.
+  const systemDialogPermission = async () => {
```
Patrick-Erichsen: I'm confused about what this method is doing. Is there a dialog that is being prompted?

gps/app/views/Settings.js
shilpashendre: Hi, TCL is a Turbo console log extension, use as a shortcut to create console prints.

gps/app/views/Settings.js
shilpashendre: Hi, TCL is a Turbo console log extension, use as a shortcut to create console prints. I will remove it.

gps/app/views/Settings.js
Patrick-Erichsen: I didn't notice in my first review that this is using the `BackgroundGeolocation` library.

The current file is for the `Settings` screen - we want to try and keep this file limited to logic that is only related to the UI content of this screen.

This logic should be moved to the [existing authorization handler in `LocationServices.js`](https://github.com/Path-Check/covid-safe-paths/blob/develop/gps/app/services/LocationService.js#L159).

The view logic for this method (e.g. `setIsLogging()`) should stay in this file but pull from LocationService.

gps/app/views/Settings.js
Patrick-Erichsen: Could you provide a more descriptive name for this method?

gps/app/views/Settings.js
```diff
@@ -55,11 +54,45 @@ export const SettingsScreen = ({ navigation }) => {
...
+        console.log('TCL: SettingsScreen -> err', err);
+      });
+  };
+
   const locationToggleButtonPressed = async () => {
     try {
       isLogging ? LocationServices.stop() : LocationServices.start();
       await SetStoreData(PARTICIPATE, !isLogging);
       setIsLogging(!isLogging);
+      systemDialogPermission();
```
Patrick-Erichsen: Why do we need this method at all if all it is doing is calling (either true of false values)

```js
setIsLogging(false);
LocationServices.stop();
 await SetStoreData(PARTICIPATE, false);
```

which we have already done in the three lines above?

Patrick-Erichsen (changes requested):

app/views/Settings.js
```diff
@@ -53,11 +50,47 @@ export const SettingsScreen = ({ navigation }) => {
...
+          'Something went wrong in system dialog location permission',
+          err,
+        );
+      });
+  };
+
+  // checking whether location permission is enable or disbale from application setting on load of setting's page.
+  const checkIsLocationEnable = async () => {
+    await LocationServices.checkStatus()
+      .then(async res => {
```
Patrick-Erichsen: .then is not needed with await

app/services/LocationService.js
```diff
@@ -258,6 +258,17 @@ export default class LocationServices {
     });
   }
 
+  // To get the system dialog click event
+  static async getSystemLocationPermission() {
```
Patrick-Erichsen, line 262: is this being used anywhere?

Patrick-Erichsen (changes requested):

app/services/LocationService.js
shilpashendre: No. Only one time in the settings page.

app/views/Settings.js
```diff
@@ -55,6 +52,41 @@ export const SettingsScreen = ({ navigation }) => {
...
+      })
+      .catch(err => {
+        console.log('Is background Location enable ', err);
+      });
+  };
+
+  //  This method is used to track whether user has "Allow" or "Deny" location permission system Dialog.
+  const systemDialogLocationPermission = async () => {
+    await LocationServices.getSystemLocationPermission()
+      .then(res => {
```
Patrick-Erichsen: This isn't quite right - I think taking a look at this link may be helpful https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

The short answer is that `await` does not require a `.then()`.

app/services/LocationService.js
Patrick-Erichsen: I still don't quite understand why this is needed. There is no callback that is passed to the authorization handler, so nothing will ever be done

Patrick-Erichsen (changes requested):

Patrick-Erichsen:
Closing as this is no longer relevant with the work being done in #945

PR #856, opened 2020-05-18
Remove unused bluetooth dependency

mattThousand:
This pull request removes `react-native-ble-advertiser` as a dependency, as well as related infrastructure. Keeping the dependency would mean that we'd request bluetooth entitlements on android for both the `gps` and the `bte` product flavors, which we do not want. 

Since the `Exposure Notifications` APIs will eliminate the need for our own bluetooth/ble implementation, we can remove this dependency.


johnschoeman (commented):
LGTM! no comments.

tstirrat (approved):

tstirrat:
There are some extra bits in manifest and ios land that might need to be removed?

https://github.com/vitorpamplona/react-native-ble-advertiser#install

tstirrat:
I think its just the Android bits:

```
<uses-permission android:name="android.permission.BLUETOOTH"/>
 <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
```

https://github.com/Path-Check/covid-safe-paths/blob/2b346de1bd4d949cc1d95ba469555fc49910207b/android/app/src/main/AndroidManifest.xml#L9-L10

mattThousand:
> I think its just the Android bits:
> 
> ```
> <uses-permission android:name="android.permission.BLUETOOTH"/>
>  <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"/>
> ```
> 
> https://github.com/Path-Check/covid-safe-paths/blob/2b346de1bd4d949cc1d95ba469555fc49910207b/android/app/src/main/AndroidManifest.xml#L9-L10

👍 those entitlements were removed from `main/AndroidManifest` in an earlier commit to the target branch:

https://github.com/Path-Check/covid-safe-paths/blob/196a4ebab89b734665250e4385d6d7491bc13c54/android/app/src/main/AndroidManifest.xml#L3

 and the remaining BTE entitlements that we'll need for EN are living in `bte/AndroidManifest`:

https://github.com/Path-Check/covid-safe-paths/blob/196a4ebab89b734665250e4385d6d7491bc13c54/android/app/src/bte/AndroidManifest.xml#L3

tstirrat:
Oh. :) thanks!

PR #858, opened 2020-05-18
clean up beta builds and remove android build job that is duplicated …

efalkner:
…by staging build job

#### Description:
Remove unnecessary beta job and references.
Remove android build job as it's doing the same thing as the staging android job


tstirrat (approved):

tstirrat:
Which checks should be required now?

`build_staging_android`?

PR #861, opened 2020-05-19
Setup Fastlane for gps and bte ios targets

johnschoeman:
Setup Fastlane for gps and bte ios targets …
8f2e18e
Why:
We would like to be able to use fastlane and match for signing and
pushing ipas for both the gps and bte targets of the app.

We would also like for to use github actions so that merges to the bte
branch will build and push the bte version of the app. This is required
to be able to ship the binaries to the app and play store, since all of
the signing certs are managed by a protected git repo that most
developers will not have access to.

Ideally we will eventually move the CD step into the same branches as
the GPS target, currently develop, release-candidate, and mater, so that
we don't have to coordinate as much between the two apps, as the vast
majority of future feature work will apply to both apps.

This commit:
Updates the fastfile and github workflow for distributing the ios
targets. We added a new set of lanes for bte_production and bte_staging.

Future work could include consolidating these lanes to build both or
either target depending on a passed parameter, though it is not clear
that this level of dryness would be worth the effort.

A more important bit of work that should probably be done in a future
commit is to automate the selection of the correct app_identifier on the
match side, currently the develop will need to comment in/out the
correct identifier.

Co-Authored-By: efalkner <elizabethfalkner09@gmail.com>
#### How to test:

Run the fastlane lanes and verify the ipas are successfully complied.

```
fastlane release_bte
fastlane staging_bte
fastlane release
fastlane staging
```

for the github workflow side, im not sure if theres a way to test either than simply running the github action.

Future work would be 

- do a similar set of work on the android side for fastlane. I got it started but am mostly focusing on getting the ios side shipped at this point.


johnschoeman:
@efalkner , @tparesi , @mattThousand , im going to merge this into bte to get going with figuring out the github workflow side, but i would still appreciate a review from yall if you have time today!

PR #864, opened 2020-05-19
Bte jts fast foward develop to bte

johnschoeman:
The purpose of this pr is to merge the first 2 commits of https://github.com/Path-Check/covid-safe-paths/pull/853,  so that develop doesnt lag too far behind on the bluetooth version:

Given that there is some changes to folder structure in these commits, there is high likelihood of confusion and merge conflicts the longer the branches are able to diverge.  Note that these changes only affect the javascript side and we will merge changes affecting the native sides in a future pr:

Commit 1:
41fec15
Why:

This commit:

Commit 2
ollback js code spliting …
d61cf20
Why:
javascript side.

This commit:


app/views/__tests__/__snapshots__/Export.spec.js.snap
```diff
@@ -206,17 +206,6 @@ exports[`<ExportScreen /> renders correctly 1`] = `
               >
                 Share location data
               </Text>
-              <SvgXml
```
tstirrat: missing icon, is this right?

shared/LICENSE.json
```diff
@@ -0,0 +1,6 @@
+{
```
tstirrat: should license.json still be in shared?

app/assets/images/index.js
```diff
@@ -0,0 +1,43 @@
...
+  LaunchScreenBackgroundOverlay,
+  Newspaper,
+  PrivateKitQRCode,
+  SaveIcon,
+  UserGreen,
+  Welcome1,
+  WWW,
+};
+
+export default Images;
```
tstirrat: use named exports everywhere.

app/assets/svgs/index.js
```diff
@@ -0,0 +1,43 @@
...
+  SettingsIcon,
+  StateAtRisk,
+  StateNoContact,
+  StateUnknown,
+  Team,
+  Warning,
+  XmarkIcon,
+};
+
+export default Icons;
```
tstirrat: same re named exports

app/locales/check.sh
```diff
@@ -9,7 +9,7 @@ echo "yarn i18n:extract"
 yarn i18n:extract
 
 echo "Checking for blank keys. Correct any that appear in this list:"
-! grep "\"\"" gps/app/locales/en.json
+! grep "\"\"" app/locales/en.json
```
tstirrat, line 12: You must also update `pull.sh` and `push.sh` and the `i18n.yml` workflow file.

app/views/__tests__/__snapshots__/Licenses.spec.js.snap
```diff
@@ -191,13 +191,7 @@ exports[`renders correctly 1`] = `
...
         }
       >
-        <Image
-          source={
-            Object {
-              "testUri": "../../../shared/assets/images/foreArrow.png",
-            }
-          }
-        />
+        <Image />
```
tstirrat: missing image?

app/views/__tests__/__snapshots__/LocationTracking.spec.js.snap
```diff
@@ -23,11 +23,6 @@ exports[`renders correctly 1`] = `
     }
   >
     <Image
-      source={
-        Object {
-          "testUri": "../../../shared/assets/images/launchScreenBackground.png",
```
tstirrat, line 28: another missing image?

tstirrat (changes requested):

shared/LICENSE.json
johnschoeman: I see no reason why not!, ill move it.

app/locales/check.sh
johnschoeman: good catch!

app/views/__tests__/__snapshots__/Export.spec.js.snap
johnschoeman: It would be nice to have typescript! this was due to a typo i believe.

app/locales/pull.sh
```diff
@@ -60,7 +60,7 @@ lokalise2 file download \
   --replace-breaks=false \
   --bundle-structure "app/locales/%LANG_ISO%.json" \
   --original-filenames=false \
-  --unzip-to=gps \
+  --unzip-to=app \
```
tstirrat, line 63: line 61 needs a change too.. remove app/ from there.

shared/LICENSE.json
johnschoeman: actually this was duplicated, i will delete this extra.

app/views/__tests__/__snapshots__/LocationTracking.spec.js.snap
johnschoeman: caused by a typo.

johnschoeman: caused by a typo, fixed.

johnschoeman:
@tstirrat , thank you for the review! I believe I have captured all of your comments.

app/assets/images/index.js
johnschoeman: Out of curiosity, why is this? Is it to maintain code consistency or is there a technical reason?

Patrick-Erichsen: I think we've been leaning towards named exports for

1. Consistency
2. IDE integration (better autocomplete, etc)

johnschoeman: Even for react components?

tparesi (dismissed):
LGTM

mattThousand (dismissed):

johnschoeman:
Replaced with: https://github.com/Path-Check/covid-safe-paths/pull/876 to get this in since others are waiting to build off of it.  cc: @tstirrat , let me know if you have any other comments/thoughts and we can get updates into a future commit.

PR #868, opened 2020-05-19
Update fastlane (android) for bte

mattThousand:
This pull request accommodates the new `bte` product flavor in the Android codebase, and adds a workflow to publish to the appropriate app on the Google Play Console

### DIFF
--- a/.github/workflows/publish-release-android-bte.yml
+++ b/.github/workflows/publish-release-android-bte.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - bte
-      - release-candidate
-      - master
     paths-ignore:
       - 'package.json'
       - 'CHANGELOG.md'
--- a/.github/workflows/publish-release-android-bte.yml
+++ b/.github/workflows/publish-release-android-bte.yml
@@ -109,13 +109,6 @@ jobs:
           subdirectory: 'android'
           bundle-install-path: 'vendor/bundle'
 
-
-      - name: Upload APK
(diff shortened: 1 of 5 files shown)


.github/workflows/publish-release-android-bte.yml
```diff
@@ -0,0 +1,117 @@
+name: Publish Release Candidates
+
+on:
+  push:
+    branches:
+      - bte
+      - release-candidate
```
johnschoeman: not likely to be an issue, but it might make sense to remove `- release-candidate` and `- master` until we have a plan on how we handle bte / gps deploys. there are some open questions here, like should they always be released together or separate?, etc.

.github/workflows/staging-android-bte.yml
```diff
@@ -0,0 +1,82 @@
+name: Staging Android
+
+on:
+  push:
+    branches:
+      - develop
```
johnschoeman: similar comment here, at this point be probably only want this running on the bte branch.

johnschoeman (approved):
LGTM! a minor comment.

.github/workflows/publish-release-android-bte.yml
mattThousand: 👍

.github/workflows/staging-android-bte.yml
mattThousand: 👍

PR #869, opened 2020-05-19
Differentiate GPS and BTE copy

tparesi:
Why:
In trying to keep the application as similar as possible, most changes are in the
onboarding flow and with copy. This PR adds logic in all places where copy differs
between the two apps. This is a good short term solution, but in the future, the
hope is to come up with a more elegant way to uncouple this logic.

This commit:

The current pattern adds a suffix of either bluetooth or location to all copy where
the two apps will differ (only in en.json).

Update COVIDSafePathsConfig to a name export.
Update Bluetooth in broadcasting service, which may be dead code at this point.


app/views/ExposureHistory/DetailedHistory.js
```diff
@@ -13,6 +14,11 @@ import { SingleExposureDetail } from './SingleExposureDetail';
  */
 export const DetailedHistory = ({ history }) => {
   const exposedDays = history.filter(day => day.exposureMinutes > 0);
+  const wtdmParaText =
```
johnschoeman: what do you think about not using an abbreviation here? something like: ` const whyThisMattersText =`

johnschoeman (approved):
LGTM

totally agree with the idea of taking a pass at finding a better solution for the internationalization stuff.
I would also start thinking about if theres a way we could not pass the tracing strategy all over the place, perhaps once we have a better understanding of the differences in workflows, it might make sense to pull the conditional `tracingStrategy === 'gps'` up the stack and render different screens/components?

app/views/ExposureHistory/DetailedHistory.js
tparesi: I think that is a good idea! Will do

PR #870, opened 2020-05-19
Hide Choose Health Authorities on Settings Page

tparesi:
Why:
The current logic for Choosing a Health Authority involves using location.
Until the new flow is worked out, we will simply hide it.


johnschoeman (approved):
LGTM

PR #876, opened 2020-05-20
Bte jts ff take 4

johnschoeman:
Creating a new pr for PR 864, to be able to merge as we have been waiting for a re-review for about a day. All requested changes were included, and this pr being open is blocking other critical path work.

The purpose of this pr is to merge the first 2 commits of https://github.com/Path-Check/covid-safe-paths/pull/853,  so that develop doesnt lag too far behind on the bluetooth version:

Given that there is some changes to folder structure in these commits, there is high likelihood of confusion and merge conflicts the longer the branches are able to diverge.  Note that these changes only affect the javascript side and we will merge changes affecting the native sides in a future pr:

Commit 1:
41fec15
Why:

This commit:

Commit 2
ollback js code spliting …
d61cf20
Why:
javascript side.

This commit:


jeffstolz (approved):
LGTM

tparesi (approved):
LGTM

PR #878, opened 2020-05-20
Remove workspaces object from package.json

mattThousand:
This pull requests removes the `workspaces` object from `package.json` as it is no longer needed and prevents developers from adding new dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,13 +138,6 @@
       "style": "module"
     }
   },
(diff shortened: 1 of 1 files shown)


yonatanmk (approved):
With this I can add packages again

johnschoeman (commented):
💯

PR #879, opened 2020-05-20
Introduce gps and bte ios targets

johnschoeman:
Why:
We would like to introduce bluetooth to the application for bluetooth
based contact tracing. Given the privacy concerns of having both
location and bluetooth data available to the same application, google
and apple do not allow for both location and bluetooth to be available
at the same time.

Hence in order to service both location and bluetooth
we will need to provide separate apps with separate info.plists and
bundle identifiers. At the same time we would like to share a much
application logic between the two applications as to reduce the
maintenance and organizational overhead. Ideally as close to 100% of the
javascript layer is common and for the ios layer we only need to change
the info.plists.

This commit:

Introduces two build targets, GPS and BTE, to the xcode project which
are setup to read from the corresponding groups for config and info.plist
A COVIDSafePathsConfig file was introduced to expose a
'getTracingStrategy' function to the react native side so that the js
runtime can determine which tracing strategy the bundle will be using.
This will allow us flexibility to programmatically determine which
tracing strategy is being used either within a given react component or
to conditional render components altogether. This also allows use to
maximize the amount of shared native code.

Three schemes where added for each target, a Development, a Staging, and
a Production.

To build the corresponding targets via the react native cli, you can
pass the --scheme flag with the corresponding scheme, e.g:

```
react-native run-ios --scheme "GPS_Development"
react-native run-ios --scheme "BTE_Development"
```

We updated the `yarn run-ios` script in package.json to two scripts for
for the gps based build and the bte based build, `yarn run-ios-bte` and
`yarn run-ios-gps` respectively

Note that theres a fair amount of location information storing logic
currently in AppDelegate, a future commit might want to refactor this so
that the gps specific code is only built in the GPS target builds.

### GPS
Screen Shot 2020-05-20 at 4 19 14 PM

### BTE
Screen Shot 2020-05-20 at 4 17 47 PM 

### Testing

Verify you can run both versions of the app.
- reinstall your pods, clean build folder, remove derived data, etc.
- run `yarn run-ios-gps`
- run `yarn run-ios-bte`


.gitignore
```diff
@@ -44,7 +44,7 @@ buck-out/
 !debug.keystore
 
 # i18next-parser
-gps/app/locales/*_old.json
```
johnschoeman, line 47: this got missed in a previous commit.

ios/GPS/Info.plist
```diff
@@ -2,20 +2,14 @@
...
-	<string/>
 	<key>BGTaskSchedulerPermittedIdentifiers</key>
 	<array>
 		<string>com.transistorsoft.fetch</string>
 	</array>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>Safe Paths</string>
+	<string>Safe Paths - GPS</string>
```
johnschoeman: This shouldnt be changed.

ios/COVIDSafePaths/Base.lproj/LaunchScreen.xib
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
johnschoeman: we shouldnt need to change the launch screen in this one.

johnschoeman:
the next increment to introduce to develop from the bte branch would be the fastlane work.

mattThousand (dismissed):

mattThousand:
LGTM once those unit tests are green 🍏 😄

tstirrat (dismissed):
Would like to get a weigh in from @efalkner in regards to fastlane/workflows too before we merge.

johnschoeman:
> Would like to get a weigh in from @efalkner in regards to fastlane/workflows too before we merge.

@efalkner and i actually paired on the fastlane/workflows on a previous thread! Im currently working on getting a pr specifically for those up, for ease of review with the rest of the team.  That said, this pr wont affect the current github workflow at all and the bte stuff wont auto build on ci.

.github/workflows/test.yml
```diff
@@ -106,9 +106,9 @@ jobs:
       # TODO(https://pathcheck.atlassian.net/browse/SAF-250): remove the skip on COVIDSafePathsTests/testRendersWelcomeScreen once it's passing
       - run: |
           xcodebuild test -workspace COVIDSafePaths.xcworkspace \
-            -scheme COVIDSafePathsTests \
+            -scheme GPS_Tests \
```
johnschoeman, line 109: Ended up needing to split the test targets as well. COVIDSafePathsTests became GPSTests

johnschoeman:
ping @mattThousand, @tstirrat , looks like my for push reset your reviews :/ if i could trouble you for another review. Ive add changes to fix up the issue with the xcode test suite. with a comment in the code.

mattThousand (approved):

PR #880, opened 2020-05-21
[SAF-264] remove WRITE_EXTERNAL_STORAGE permission for Android App

krmannix:
#### Description:

Remove the `WRITE_EXTERNAL_STORAGE` permission from the `AndroidManifest.xml`. The ticket implies it was unnecessary, so I don't believe any application code changes are needed?

#### Linked issues:
https://pathcheck.atlassian.net/projects/SAF/issues/SAF-264

#### How to test:
As this was removing a vestigial permission, I didn't add any tests. I was able to run the app on an emulator. If there's a test I can add, please let me know!

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -2,7 +2,6 @@
   package="org.pathcheck.covidsafepaths">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED"/>
 


tstirrat:
I think this is needed when you export, can you export json and try to save the json on device without this permission?

krmannix:
I was able seemingly able to export the JSON data running this build on the Android Emulator.

At the export screen:
Screen Shot 2020-05-22 at 3 32 16 PM

I did get this permissions pop-up:
Screen Shot 2020-05-22 at 3 32 24 PM

I think this is just because my emulator doesn't have any apps that enable sharing, though I'm not too familiar with Android (I personally have an iOS device). Is there an app I can install to my emulator that will work with this feature?
Screen Shot 2020-05-22 at 3 32 34 PM

I'm able to see the file in the File app, in the Downloads section.
Screen Shot 2020-05-22 at 3 32 53 PM

I'm not sure why it has a `.null` suffix, but this appeared for me both with or without this commit.

The data in the file:
```json
[{"longitude":-122.08400000000002,"latitude":37.421998333333335,"time":1590190260000}]
```

Are there other flows I should test?

tstirrat:
OK thanks for checking.

efalkner:
I think @tstirrat was correct, that [the permission is required for react-native-share](https://github.com/react-native-community/react-native-share/issues/430). I'd recommend testing this on a physical Android device if possible. In addition, I'd check how it works with Android 10 and then Android 9 and below. There were significant changes in Android's APIs around file storage which caused a lot of file access to break, [without updating the manifest accordingly](https://github.com/react-native-community/react-native-share/issues/650)

efalkner:
To test sharing on an emulator, you could share to the messages app, or sign yourself in to google drive. That being said, the file system on the emulator does not always behave the same way that a physical device would, so I'd take the results with a grain of salt. I'm happy to test out on one of my android devices here; we can also send an artifact to the testing team to try out on Perfecto

krmannix:
@efalkner thanks for the comment + info! unfortunately, I don't have a physical Android device handy, and I'd hate to put the burden of testing my own changes on someone else. Happy to close this PR (and update the associated ticket) for now, as it seems that the reward is low compared to the risk of preventing data sharing. Does that sound okay?

Patrick-Erichsen:
Closing out as per @efalkner's request

Thanks for contributing though @krmannix !

PR #883, opened 2020-05-21
Introduce fastlane and github workflow for bte ios 

johnschoeman:
Why:
We would like to be able to use fastlane and match for signing and
pushing ipas for both the gps and bte targets of the app. We would also
like to have CD setup via github workflows for the BTE version of the
app.

This commit:
Introduces workflows for building and deploying the BTE version of the
app equivalent to the the existing workflows for the GPS app.

We set the triggering branch for publishing to be `bte-release`. A
lot of noise and potentially unnecessary builds are made if a new build
is cut with every commit to the develop branch. By having the release
action happen for new commits to bte-release we can better control what
builds get sent to the app and play store. We simply need to fastfoward
the `bte-release` branch to the correct commit on develop, (or ideally
master) I would consider doing a similar strategy for the gps build.

We also update the fastfile for distributing the ios targets. We added a
new set of lanes for bte_production and bte_staging.

Future work could include consolidating these lanes to build both or
either target depending on a passed parameter, though it is not clear
that this level of dryness would be worth the effort.

Co-Authored-By: efalkner <elizabethfalkner09@gmail.com>


.github/workflows/staging-android-bte.yml
```diff
@@ -0,0 +1,76 @@
...
+            ${{ runner.os }}-gems-
+
+      - name: Install Ruby Dependencies
+        if: steps.ruby_cache.outputs.cache-hit != 'true'
+        run: | 
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+        working-directory: ./android
+
+      - name: Build android Staging artifact
```
mattThousand: a few android things snuck in, remove them for now as they've changed on my branch?

.github/workflows/publish-release-bte.yml
```diff
@@ -0,0 +1,113 @@
+name: Publish Release Candidates BTE
+
+on:
+  push:
+    branches:
+      - bte-release
```
johnschoeman, line 6: So the idea here is that instead of pushing a build every time someone merges something into develop, we actively choose when to release a build to the stores by fast forward merging the bte-release and gps-release branches.  This will save us from having a lot extra builds submitted for processing that we arent actually intending to ship.

ios/fastlane/Fastfile
```diff
@@ -134,15 +135,57 @@ platform :ios do
...
+        :DEVELOPMENT_TEAM => TEAM
+      }
+
+      match(
+        type: 'appstore',
+        readonly: false,
+        keychain_name: "tech_at_safepaths_keychain",
+        keychain_password: "meow",
+        git_basic_authorization: ENV["secrets.MATCH_GIT_BASIC_AUTH"],
+        app_identifier: 'org.pathcheck.bt'
```
johnschoeman, line 179: @mattThousand , @efalkner , I think this will solve the bundle id issue we bumped up into for the various flavors/targets.  Curious on you thoughts here.

.github/workflows/publish-release-bte.yml
mattThousand: superb

mattThousand (approved):

PR #873, opened 2020-05-20
Append git sha to continuous build version

tstirrat:
- Sets the version to `v1.0.x-<sha>` so that it's clear how it maps to the repo.

#### Screenshots:

Screen Shot 2020-05-28 at 6 18 11 PM

PR #884, opened 2020-05-21
BTE/GPS android product flavor setup

mattThousand:
This commit:
- Introduces workflows for building and deploying the BTE version of the android
app equivalent to the the existing workflows for the GPS app.

- Sets the triggering branch for publishing to be `bte-release`.

- Updates the `Fastfile` for distributing android `gps` and `bte` product flavors. 

- Adds a native module to determine `tracingStrategy` based on product flavor (`bte` or `gps`)


android/fastlane/Fastfile
```diff
@@ -23,59 +23,98 @@ platform :android do
...
 
-  desc "Build a Release APK and AAB"
-  lane :release do
+  desc "GPS: Build a Release AAB"
+  lane :release_gps do
 
-    increment_version_code(
-      gradle_file_path: "../android/app/build.gradle",
-    )
+    build_number =  ENV["GITHUB_RUN_NUMBER"]
```
mattThousand, line 37: The idea here is to set the version code automatically from the workflow run number. `VersionName` updates are infrequent and can be done manually in `build.gradle`

android/app/src/main/java/org/pathcheck/covidsafepaths/util/COVIDSafePathsConfig.kt
```diff
@@ -0,0 +1,23 @@
+package org.pathcheck.covidsafepaths.util
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.bridge.ReactContextBaseJavaModule
+import com.facebook.react.bridge.ReactMethod
+import org.pathcheck.covidsafepaths.BuildConfig
+
+class COVIDSafePathsConfig(reactContext: ReactApplicationContext) : ReactContextBaseJavaModule(reactContext) {
+
+    override fun getName(): String {
```
johnschoeman, line 9: minor thing, but not sure we are ever going to need this function?

android/fastlane/Fastfile
johnschoeman, line 37: I dig this! i would double check that its not conflicting with some of the work that is being done here: https://github.com/Path-Check/covid-safe-paths/pull/873

let me know if you wanna chat through it some more, but Id love to get the buildnumbers for the ios side incrementing and independent of the version number as well, its a bit odd to have the version number change along with the build number and means that we have many many versions.

.github/workflows/staging-android-bte.yml
```diff
@@ -0,0 +1,82 @@
+name: Staging Android
```
johnschoeman:
```suggestion
name: Staging Android BTE
```

.github/workflows/staging-android-bte.yml
```diff
@@ -0,0 +1,82 @@
...
+
+on:
+  push:
+    branches:
+      - develop
+      - master
+      - release-candidate
```
johnschoeman:
```suggestion

```

johnschoeman (dismissed):
Looking good to me!  A few minor minor comments, if youre interested.

johnschoeman:
> @mattThousand mattThousand dismissed johnschoeman’s stale review via 7a37b05 34 seconds ago

oh man i hate that this happens on every code push, ping me when its ready to go and i can re-review!

android/app/src/main/java/org/pathcheck/covidsafepaths/util/COVIDSafePathsConfig.kt
mattThousand: good question: it's used to expose the name of the module to the `JS` layer, so we are stuck with it I believe

android/fastlane/Fastfile
mattThousand: I like @tstirrat's approach in #873 - we'd read the new version name from `package.json`. The `versionCode` on android and the `build number` on iOS should be independent, and increment automatically on each workflow run. What do you think?

johnschoeman: tbh, i think we could be manually upping the build numbers, that said, having some automation is nice and if tim's strategy works then it works!

johnschoeman (approved):
💯 🚢

android/fastlane/Fastfile
mattThousand: sounds good! this PR won't address versionName; in the interest of time we can keep the versionCode incrementation logic here and we can all discuss the versionName piece or revisit this question as part of #873

tstirrat: Yeah, version codes can be whaterver as long as they always increase. Seems good.

android/fastlane/Fastfile
```diff
@@ -23,59 +23,96 @@ platform :android do
...
+  end 
+
+  desc "Build a Release AAB"
+  lane :release_bte do
+
+    build_number =  ENV["GITHUB_RUN_NUMBER"]
+
+    increment_version_code(
+      gradle_file_path: "app/build.gradle",
+      version_code: build_number
```
tstirrat, line 85: This would have been clearer as `github_run_number`

PR #886, opened 2020-05-21
[Build] Show eslint output on failure

tstirrat:
The eslint checker adds annotations to the source, but does not show what failed in github actions. This should fix that.

#### How to test:

Now you can see warnings in the job log: https://github.com/Path-Check/covid-safe-paths/pull/886/checks?check_run_id=697716251


JacobJaffe (approved):
🎉 thanks for making this debugging easier

PR #887, opened 2020-05-21
GPS/BT copy and UI changes

tparesi:
#### Commit 1: Split GPS and BT functionality 

Why:
In trying to keep the application as similar as possible, the changes are
kept to a minimum within components. This PR adds logic in all places
where copy and UI differs between the two apps. This is a good short term
solution, but in the future, the hope is to come up with a more elegant
way to uncouple this logic.

This commit:

The current pattern adds a suffix of either bluetooth or location to all
copy where the two apps will differ (only in en.json).

Update Bluetooth in broadcasting service, which may be dead code at this point.

#### Commit 2: Remove Permissions Onboarding step from BT flow

Why:
Since no notifications or bluetooth functionality is working, remove the
permissions flow from the BT application so there are no hold ups in the
app store.

This commit:
Check tracing strategy before going to the permissions page, and route
back to main when using BT strategy.

#### Commit 3: Add a Not Available page for Exposure Notifications

Why:
When the user does not have a healthcare authority chosen, a screen is
needed to notify the user that the service is not available.

Update colors for WCAG AA accessibility

#### Commit 4: Make visual and copy changes for Setting related screens

This commit:

Change the Settings screen to hide non BT functional or related items.
Change the Choose HA page for BT.
Update the Legal and About page's styling.
Update copy.
Change Navigation Title styling.
Remove LICENSE.json.
Hide GoogleMapsImport for BT

### Screenshots:

Attached below

Co-authored-by: jeffstolz <jstolz123@gmail.com>


johnschoeman (approved):
LGTM! Theres a lot here, but the general patterns seem fine enough for now. Once we are ready for BT app store stuff we should make sure to do another pass at the factoring here.

PR #888, opened 2020-05-21
Fix app identifier typo in ios matchfile

johnschoeman:
Fix app identifier typo in ios matchfile …
9ae783e
Why:
Currently there is a typo in the match file, which would cause gps
releases to fail.

Co-Authored-By: tparesi <tparesi@gmail.com>


tparesi (approved):
LGTM

mattThousand (approved):

PR #891, opened 2020-05-22
Revert changes to onboarding5

johnschoeman:
Why:
A recent commit, 08ceca29e42c115b9afa9f8a48757c0d02815853, that
introduced ui changes for the bluetooth app introduced a bug that caused
the permission requesting screen for the gps version of the build to
flicker and not allow the user to progress.

This commit:
Reverts the changes in that screen. We will eventually need to
reintroduce these changes, but the component is rather complicated and
would take some time to fix, by reverting we can get a working
experience out and reintroduce the changes without the bug.

Addresses: https://github.com/Path-Check/covid-safe-paths/issues/890

### Before

flicker_before

### After

flicker_afert

### DIFF
--- a/app/views/onboarding/Onboarding5.js
+++ b/app/views/onboarding/Onboarding5.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { Component } from 'react';
 import {
   Dimensions,
   ImageBackground,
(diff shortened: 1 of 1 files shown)


tparesi (approved):
Bummer we had to completely roll it back, but makes sense since we are not using this screen for the BT app at the moment anyways with the current flow.

johnschoeman:
> Bummer we had to completely roll it back, but makes sense since we are not using this screen for the BT app at the moment anyways with the current flow.

yeah this component is just too gnarly to track down whats actually happening quickly. When we get around to adding the bluetooth side of the permissions we can reintroduce the changes.

PR #892, opened 2020-05-22
Remove old LocationTracking code and flag

tstirrat:
Removes the old LocationTracking code which has no tests and mingles state and UI together.

The new Main.js has been the dev and staging mode default for at least 2 weeks now.

Also, while here:
* Makes `<Entry />` a functional component
* Consolidates `headerShown: true` to reduce repetition (copied rom @JacobJaffe)
* Makes the `onboardingDone` logic more concise.

#### Linked issues:

Originally introduced in #749 

#### Screenshots:

No change

#### How to test:

Check that the location tracking screen loads and acts as expected.


tstirrat:
the e2e are failing, not sure if that's new to this PR.

app/Entry.js
```diff
@@ -27,127 +27,58 @@ import { SettingsScreen } from './views/Settings';
...
-      .then(onboardingDone => {
-        this.setState({
-          initialRouteName: onboardingDone,
-        });
-      })
-      .catch(error => console.log(error));
-  }
+  useEffect(() => {
+    async function checkDone() {
+      const flag = await GetStoreData('ONBOARDING_DONE');
```
Patrick-Erichsen:
```suggestion
      const flag = await GetStoreData('ONBOARDING_DONE', false);
```

should use `JSON.parse()` to convert that to a regular boolean.

Also - maybe throw `ONBOARDING_DONE` into `constants/storage.js` ?

app/Entry.js
```diff
@@ -27,127 +27,58 @@ import { SettingsScreen } from './views/Settings';
...
 
-class Entry extends Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      initialRouteName: '',
-    };
-  }
+export const Entry = () => {
+  const [onboardingDone, setOnboardingDone] = useState(false);
```
JacobJaffe, line 32: I think this should be a tertiary state, with `useState(null)` for the loading. 
We're really requesting an async load of app state before mounting the app, which would be good to have a splash mock handler for:

```js 
const appLoading === onboardingDone === null || ...otherAsyncLoads;

if (appLoading) return <SplashScreenMock/>

return <Navigation>....
```

Currently the app flashes because of the switch between onboardnigDone going from false to true on the useEffect.

app/Entry.js
tstirrat: Done

Patrick-Erichsen (dismissed):

Patrick-Erichsen (approved):

PR #900, opened 2020-05-22
Fix getTracingStrategy on Android

mattThousand:
Ensures that `getTracingStrategy()` will not return `undefined` when calling into the native layer on Android.


efalkner (commented):
In which cases was this method returning undefined? Would it make sense to use the Context.getPackageName() method provided at the native layer? [React-Native's comments](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactMethod.java#L27) around the annotation suggest avoiding the blocking synchronous context.

mattThousand:
> In which cases was this method returning undefined? Would it make sense to use the Context.getPackageName() method provided at the native layer? [React-Native's comments](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/ReactMethod.java#L27) around the annotation suggest avoiding the blocking synchronous context.

All cases :) (blocking vs non-blocking below):
Screen Shot 2020-05-23 at 10 15 25 AM

In certain cases a synchronous call would indeed cause performance penalties (per the RN team's note) but we're just checking a string against the build config here so I wouldn't worry about that in this case. If we were accessing local data, making a network call, or doing computationally expensive or time-intensive work, I wouldn't reach for this annotation

Thinking about how the JS layer is calling this method, we should probably just make a single call on app launch and cache the result, as it will not change. But that's for another PR :)

tstirrat (dismissed):
Approving this to fix GPS for now, but it does feel like a massive code smell having this blocking method. Would like us to explore a different approach..

We already use `react-native-config` which allows passing a flag from build time into the app, and it can be configured per flavor [here](https://github.com/Path-Check/covid-safe-paths/blob/develop/android/app/build.gradle#L7-L11) ... maybe that's enough?

efalkner:
Sounds good to me. Let's revisit in a separate PR. Thanks for the explanation @mattThousand

efalkner:
;) just saw the new commits. checking it out locally  :D thanks

tstirrat (dismissed):

efalkner (dismissed):
LGTM.

tstirrat:
I'm making some tweaks to `.env` file config in `app/build.gradle` so that it automatically picks files:

```
$ ./gradlew assembleGpsRelease --stacktrace

> Configure project :app
Reading env from: .env.gps.release  <-----
```

```
$ ./gradlew assembleBteDebug

> Configure project :app
Reading env from: .env.bte
```

```
$ yarn run-android-bte
yarn run v1.22.4
$ react-native run-android --variant=bteDebug --appId=org.pathcheck.covidsafepathsBte
<  snip >
info JS server already running.
info Installing the app...

> Configure project :app
Reading env from: .env.bte <--------
```

Result:
Screen Shot 2020-05-24 at 8 44 10 AM

```
$ yarn run-android-gps
yarn run v1.22.4
$ react-native run-android --variant=bteDebug --appId=org.pathcheck.covidsafepathsBte
<  snip >
info JS server already running.
info Installing the app...

> Configure project :app
Reading env from: .env.gps <--------
```

```
$ yarn run-ios-gps
yarn run v1.22.4
$ yarn install:pod && ENVFILE=.env.gps react-native run-ios --scheme "GPS_Development"
```

.github/workflows/create-release.yml
```diff
@@ -1,95 +0,0 @@
-name: Create Release Assets
```
tstirrat, line 1: I deleted this because it's wayyyyyyy out of date/irrelevant now.

tstirrat:
```
$ yarn run-ios-gps
yarn run v1.22.4
$ yarn install:pod && ENVFILE=.env.gps react-native run-ios --scheme "GPS_Development"
```

Screen Shot 2020-05-24 at 8 56 23 AM


```
$ yarn run-ios-bte
yarn run v1.22.4
$ yarn install:pod && ENVFILE=.env.bte react-native run-ios --scheme "BTE_Development"
```

Screen Shot 2020-05-24 at 9 00 51 AM

tstirrat (approved):

Issue #589, opened 2020-04-20
Implement a good way to mock tests with timezone and DST dependencies


kenpugsley:
Pure tech debt item.  We have a few unit tests (notably LocationService.spec.js and ExposureHistory.spec.js) that must mock the current datetime, but are sensitive to the timezone and daylight savings time.  We need to come up with a good solution for unit testing the edge cases around these tests, as well as ensuring that the tests don't have any lurking system time / timezone sensitivity.

lewisPrescott707:
I can have a look at this tomorrow

Patrick-Erichsen:
@lewisPrescott707 any progress here?

lewisP707:
> @lewisPrescott707 any progress here?

Not yet I'm afraid, as not had time.
Looks like timezone is not supported by dayjs which makes mocking tricky.
Will take another look at the weekend

lewisP707:
@kenpugsley @Patrick-Erichsen sorry I don't really have the knowledge of the Intersect.js to be able to create the data for the tests. So I have created a PR to demonstrate how to mock the timezones with dayjs. [PR-906](https://github.com/Path-Check/covid-safe-paths/pull/906)

PR #906, opened 2020-05-25
Test - Add ability to mock timezones

lewisP707:
#### Description:

Mock Timezone in Intersect.spec.js.

I don't fully understand how the bins are created, so would need someone with knowledge of Intersect.js to be able to provide the expected results for the daylight saving tests to be added

#### Linked issues:

Adds ability to mock timezone: #589


Patrick-Erichsen:
Hey @lewisPrescott707 , thanks for opening the PR! I'm a little confused about what new functionality is being introduced here though. It seems like this is just refactoring up some existing logic?

lewisP707:
@Patrick-Erichsen - @tstirrat hopefully can shed some light on how he thought about setting up this scenario:
```
// A daylight savings time change should cause either a 23 or 25 hour day.
    // TODO: Once we've figured out a good way to mock the timezone and DST, make this test work
```

PR #907, opened 2020-05-25
Export location saving to JS

Tambet-Ingo-RocketInsights:
Add a method to save locations from JS.
Update database schema to store location hashes.
Save/restore location hashes to/from database.
Add a test to verify the functionality.
Remove code to save background locations from Android application.

### DIFF
--- a/android/app/src/main/java/org/pathcheck/covidsafepaths/MainApplication.java
+++ b/android/app/src/main/java/org/pathcheck/covidsafepaths/MainApplication.java
@@ -4,7 +4,6 @@ import android.app.Application;
 import android.content.Context;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
-import com.marianhello.bgloc.BackgroundGeolocationFacade;
 import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
 import com.facebook.soloader.SoLoader;
@@ -54,7 +53,6 @@ public class MainApplication extends Application implements ReactApplication {
     SoLoader.init(this, /* native exopackage */ false);
     initializeFlipper(this); // Remove this line if you don't want Flipper enabled
     Realm.init(this);
-    initializeGeolocationTransformer();
     // Ignore assignment. Creating to begin heavy encryption work
     SecureStorage wrapper = SecureStorage.INSTANCE;
   }
@@ -83,17 +81,6 @@ public class MainApplication extends Application implements ReactApplication {
         e.printStackTrace();
       }
     }
-  }
-
-  private void initializeGeolocationTransformer() {
-    BackgroundGeolocationFacade.setLocationTransform((context, location) -> {
-
-      // Save Location in encrypted realm db
-      SecureStorage.INSTANCE.saveDeviceLocation(location);
-
-      // Always return null. We never want to store data in the libraries SQLite DB
-      return null;
-    });
   }
 
   public static Context getContext() {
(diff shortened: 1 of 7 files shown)


android/app/src/main/java/org/pathcheck/covidsafepaths/bridge/SecureStorageManager.kt
```diff
@@ -15,6 +17,11 @@ class SecureStorageManager(reactContext: ReactApplicationContext) : ReactContext
     return "SecureStorageManager"
   }
 
+  @ReactMethod
+  fun saveLocation(location: ReadableMap, promise: Promise) {
```
tstirrat, line 21: Where/when is this called from js land?

android/app/src/main/java/org/pathcheck/covidsafepaths/storage/Location.kt
```diff
@@ -77,6 +117,11 @@ open class Location(
         }
         val latitude = map.getDouble(KEY_LATITUDE)
         val longitude = map.getDouble(KEY_LONGITUDE)
+        val hashes = map.getArrayOrNull(KEY_HASHES).toRealmStringList()
+        val accuracy = map.getDoubleOrNull(KEY_ACCURACY)?.toFloat()
+        val altitude = map.getDoubleOrNull(KEY_ALTITUDE)
```
tstirrat, line 122: After a recent privacy review, we removed altitude because we were not using it at the time...

It's unclear here if we're using these new keys: altitude, bearing or speed.

I'd recommend leaving them out until we use them.

@AdamLeonSmith may want to chime in

android/app/src/main/java/org/pathcheck/covidsafepaths/MainApplication.java
```diff
@@ -85,17 +83,6 @@ private static void initializeFlipper(Context context) {
     }
   }
 
-  private void initializeGeolocationTransformer() {
```
tstirrat, line 88: I think removing this restores default storage behaviour, which will leave all the unencrypted data in the underlying `react-native-background-geolocation` storage (sqlite)..


We probably still want this override to return `null`.

@AdamLeonSmith

tstirrat (changes requested):
This looks to remove the `BackgroundGeolocationFacade.setLocationTransform`, yet there is no js land code calling `saveLocation`. From what I can tell, this stops logging locations altogether.. is that right?

I might be missing some context
Where/when does the new location data get stored?

android/app/src/main/java/org/pathcheck/covidsafepaths/bridge/SecureStorageManager.kt
Tambet-Ingo-RocketInsights: The JS caller is in another commit c1939894889bec15c4bec6d474c40de777e36d24 in  `ymk/encrypt-location-data` branch.

android/app/src/main/java/org/pathcheck/covidsafepaths/storage/Location.kt
Tambet-Ingo-RocketInsights: Ok, I'll remove  altitude, bearing and speed.

android/app/src/main/java/org/pathcheck/covidsafepaths/MainApplication.java
Tambet-Ingo-RocketInsights: If it "transforms" the location to `null`, it won't get emitted on `BackgroundGeolocation.on('location'...)`:

https://github.com/mauron85/background-geolocation-android/blob/2693e07587eea08746004bd3e72dde276b04b4d2/src/main/java/com/marianhello/bgloc/service/LocationServiceImpl.java#L534

Tambet-Ingo-RocketInsights: But you're right, `react-native-background-geolocation` will persist it in its own storage now, which is also not what we want...

Tambet-Ingo-RocketInsights:
Closing this one, new attempt at #925.

PR #909, opened 2020-05-25
Update permissions onboarding screen for bt

johnschoeman:
### Commit 1:

Refactor Onboarding5 screen …
b7c8b7b
Why:
We would like to update the onboarding5 screen to conditionally render
the location permission request buttons and location permissions logic
depending on the tracing strategy. That is if the tracing strategy is
bluetooth, we do not want to request location permissions for users
during the onboarding process.

Currently the onboarding5 screen, which requests permissions from the
user, is factored in such a way that making updates to the logic is
challenging. Specifically the view rendering logic is tightly coupled to
the screen state and the handling of permission request side effects.

This commit:
Is the hard refactor to make the change we want easy; no new
functionality has been added.

Specifically, we refactored the Onboarding5 screen to separate the requesting
permissions from the screen itself and to separate the component state
of which permission step they are on with the rendering of the
permission status indicator and request button.

We introduced a react context, PermissionsContext, to contain the
permissions states and the side effects of requesting permissions. By
placing this logic in a context, we fully separate the concerns of
requesting permissions, with rendering screens based off the current
status of permissions and allow for multiple screens to share this
logic.

In order to use this context we refactored Onboarding5 to be functional
component.

In the screen itself, we introduced the abstraction of a step, which
contains all of data needed for a particular step to be properly
displayed to the user. By separating the required data of a step and the
logic to render the step, we were able to remove a lot switch statements
and in consolidate a lot of the conditional logic.

Finally we changed the component name from Onboarding5 to
OnboardingPermissions to better describe the intent of the screen.

### Commit 2:

Only request location permission in gps app …
ebd7288
Why:
We only want to be requesting location permissions from the user in the
gps version of the app.

This commit:
Conditionally renders the location permissions indicator and enable
button such that it only appears in the gps build.

---

### GPS_Development

Screen Shot 2020-05-25 at 10 36 13 AM

### BTE_Development

Screen Shot 2020-05-25 at 10 39 48 AM


app/PermissionsContext.js
```diff
@@ -0,0 +1,141 @@
...
+    setLocationPermission(statusToEnum(status));
+  };
+
+  const checkNotificationStatus = async () => {
+    const { status } = await checkNotifications();
+    setNotificationPermission(statusToEnum(status));
+  };
+
+  const checkSubsriptionStatus = async () => {
+    const hasUserSetSubscription = await HCAService.hasUserSetSubscription();
```
johnschoeman, line 102: When pulling out this stuff, I was a little confused at what was going on, since it defaulted to 'granted' no matter what and didn't seem to be doing much yet.  So I tried to keep feature parity, but there wasnt a clear way for me to tell if anything changed/broke.

That said, given that this is only happening on the dev builds i didnt spend too much time digging into it, so if anyone has some context here and could verify if didnt break something for someone that would be much appreciated.

app/PermissionsContext.js
```diff
@@ -0,0 +1,141 @@
+import React, { createContext, useEffect, useState } from 'react';
```
johnschoeman, line 1: An alternative to the context could have been some custom hooks. I think the idea of a context for this makes sense though since it seems like there are a lot of desperate screens/components that want to be interacting with react-native-permissions and a context is a good way to keep a single source of truth for this situation.

That said, I'm not sure what the plan is for permission management is in general, so please let me know if this is butting up against other changes in the works and i can update it follow the preferred pattern.

Finally, while i think its best practice to not have all this conflated with the view logic, as a worst case scenario if this conflicts with other work to a large degree, i can inline this logic back into the screen and we can find a way to pull it out more holistically in a latter commit.

ios/Podfile
```diff
@@ -41,9 +41,9 @@ def shared_pods
   pod 'RNReanimated', :path => '../node_modules/react-native-reanimated'
 
   permissions_path = '../node_modules/react-native-permissions/ios'
-  pod 'Permission-BluetoothPeripheral', :path => "#{permissions_path}/BluetoothPeripheral.podspec"
-  pod 'Permission-Notifications', :path => "#{permissions_path}/Notifications.podspec"
+  pod 'Permission-LocationWhenInUse', :path => "#{permissions_path}/LocationWhenInUse.podspec"
   pod 'Permission-LocationAlways', :path => "#{permissions_path}/LocationAlways.podspec"
+  pod 'Permission-Notifications', :path => "#{permissions_path}/Notifications.podspec"
```
mattThousand, line 46: only tangentially related to this PR, but the `Permission-LocationWhenInUse` pod should be moved over to the `gps` target, and any residual permissions for location or bluetooth should be removed from the `BTE` `info.plist`

johnschoeman: yup, we'll need to tackle this in a future commit.

app/PermissionsContext.js
tstirrat: @Patrick-Erichsen should have more context here

maybe it was an oversight, or intentional? either way lets add a comment to the code if it's intentional.

For now, to unblock, you could add a `// TODO(<SAF-XX-link>): blah` and add a jira task to follow up and fix/clarify the HCA auto sub logic.

__mocks__/react-native-permissions.ts
```diff
@@ -1,12 +1,13 @@
 import * as RNPermission from 'react-native-permissions/lib/typescript';
```
tstirrat: This is where the eslint is failing... there's an annotation.. 

```
'RNPermission' is defined but never used.
```

silly gh action didn't make this clear in its logs, though.

app/Entry.js
```diff
@@ -23,7 +23,7 @@ import Onboarding1 from './views/onboarding/Onboarding1';
 import Onboarding2 from './views/onboarding/Onboarding2';
 import Onboarding3 from './views/onboarding/Onboarding3';
 import Onboarding4 from './views/onboarding/Onboarding4';
-import Onboarding5 from './views/onboarding/Onboarding5';
+import OnboardingPermissions from './views/onboarding/OnboardingPermissions';
```
tstirrat: Yes! renaming this is helpful. I have never remembered which OnboardingX is which.

app/PermissionsContext.js
tstirrat: Agreed on this not being in view logic. A context seems fine for now. 


One thing to note, we are already checking location permissions in the `LocationServices.checkStatus()`. It uses a slightly different method than you have here 

(Hopefully they yield the same results.. but they're technically 2 different modules).

Probably no action here, but wanted to make you aware.

app/views/onboarding/Onboarding5.js
```diff
@@ -1,519 +0,0 @@
-import React, { Component } from 'react';
```
tstirrat, line 1: If there's a way to force this file to be marked as "moved", that would be helpful. Not sure if git can do it though.

app/views/onboarding/OnboardingPermissions.js
```diff
@@ -0,0 +1,264 @@
...
+  permissionIcon: {
+    alignSelf: 'center',
+  },
+  skipThisStepBtn: {
+    color: Colors.DIVIDER,
+    paddingTop: 15,
+  },
+});
+
+export default OnboardingPermissions;
```
tstirrat: Use a named export.

tparesi (dismissed):
This makes a lot of sense to me. Thanks for taking the time to refactor this!

app/PermissionsContext.js
```diff
@@ -0,0 +1,158 @@
...
+
+  useEffect(() => {
+    const checkAllPermissions = async () => {
+      const isGPS = config.tracingStrategy === 'gps';
+      const isiOS = Platform.OS === 'ios';
+      const isDev = __DEV__;
+      await Promise.all([
+        isGPS ? checkLocationPermission() : null,
+        isiOS ? checkNotificationPermission() : null,
+        isDev && isiOS ? checkAuthSubscriptionPermission() : null,
```
tstirrat, line 80: Add a TODO here:

```
// TODO(https://pathcheck.atlassian.net/browse/SAF-232): Put HCA auto sub logic behind a feature flag
```

tstirrat (changes requested):

__mocks__/react-native-permissions.ts
johnschoeman: awesome! great catch

app/views/onboarding/Onboarding5.js
johnschoeman: probably, not sure if its worth the effort to figure out though.

app/PermissionsContext.js
johnschoeman: got it, havent had a chance to dig into the location service, so dont have any strong opinions, but it might makes sense to consolidate and have the location service defer to the permissions context for permission stuff.

johnschoeman:
any more thoughts on this one: @tstirrat ?

app/PermissionsContext.js
Patrick-Erichsen: How come this is iOS specific?

app/PermissionsContext.js
Patrick-Erichsen: So there are two confusing parts about this.

One is that this isn't a device permission, it's just a flag in state. We also didn't implement a modal for this request yet, so it's just a button, which implies that if a user presses the `Enable Autosubscription` button, it should just default to granting the permission.

The second is that the subscription state is one of three values: `true`, `false`, or `undefined`. True/False represents an explicit approve/deny that a user has granted, while `undefined` implies that the user hasn't been prompted yet.

Given that the only place we really need to check this is onboarding, I think it makes sense to just remove the `undefined` state, i.e.

```js
const checkAuthSubscriptionPermission = async () => {
  const isEnabled = await HCAService.isAutosubscriptionEnabled();
   const status = isEnabled
        ? PermissionStatus.GRANTED
        : PermissionStatus.DENIED;

    setAuthSubscriptionPermission(statusToEnum(status));
}
```

app/PermissionsContext.js
```diff
@@ -0,0 +1,160 @@
...
+        isiOS ? checkNotificationPermission() : null,
+        // TODO(https://pathcheck.atlassian.net/browse/SAF-232): Put HCA auto sub logic behind a feature flag
+        isDev && isiOS ? checkAuthSubscriptionPermission() : null,
+      ]);
+    };
+    checkAllPermissions();
+  }, []);
+
+  const checkLocationPermission = async () => {
+    const status = await check(PERMISSIONS.IOS.LOCATION_ALWAYS);
```
Patrick-Erichsen: Do we need to check OS here? What about `PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION` ?

app/PermissionsContext.js
Patrick-Erichsen: So - I think there is going to be some differences between the permissions here and those returned from `LocationService.checkStatus()`.

For example, this is the enum we are using for reasons that location tracking could be disabled

```js
export const Reason = {
  LOCATION_OFF: 'LOCATION_OFF',
  NOT_AUTHORIZED: 'NOT_AUTHORIZED',
  USER_OFF: 'USER_OFF',
};
```

In this case, these represent the following:

- `LOCATION_OFF`: Location services are disabled for the device 
- `NOT_AUTHORIZED`: Location services are not enabled for this app
- `USER_OFF`: Location services are enabled for this app, but the user has requested that we stop tracking (e.g. by toggling the `Location active` button on the `Settings` screen

In addition, the `BackgroundGeoloation` library provides these statuses for authorization states 

```ts
NOT_AUTHORIZED: AuthorizationStatus;
AUTHORIZED: AuthorizationStatus;
AUTHORIZED_FOREGROUND: AuthorizationStatus;
```

`LocationService.checkStatus()` ends up returning an object with other additional info rolled into, e.g. 

```js
if (authorization != BackgroundGeolocation.AUTHORIZED) {
      return {
        canTrack: false,
        reason: Reason.NOT_AUTHORIZED,
        hasPotentialExposure,
        isRunning,
      };
    }
```

To summarize - I think both this file and `LocationService` are providing two slightly different sets of permission info. This ContextProvider is purely app-specific device permissions, whereas `LocationService` also takes into account device and user-level statuses.

It definitely feels like we need to consolidate, and I think this ContextProvider is where it should happen.

app/PermissionsContext.js
```diff
@@ -0,0 +1,160 @@
...
+
+  const requestAuthSubscriptionPermission = async () => {
+    await HCAService.enableAutoSubscription();
+    const status = PermissionStatus.GRANTED;
+    setAuthSubscriptionPermission(status);
+  };
+
+  return (
+    <PermissionsContext.Provider
+      value={{
```
Patrick-Erichsen, line 142: Is there a reason we can't just set `intialState` to these values? With the status of `PermissionStatus.UNKNOWN` for each one

app/views/onboarding/OnboardingPermissions.js
```diff
@@ -0,0 +1,262 @@
...
+    PermissionsContext,
+  );
+  const [currentStep, setCurrentStep] = useState(0);
+  const isGPS = config.tracingStrategy === 'gps';
+  const isiOS = Platform.OS === 'ios';
+  const isDev = __DEV__;
+
+  const handleRequestLocation = async () => {
+    await location.request();
+    setCurrentStep(currentStep + 1);
```
Patrick-Erichsen: If we're just doing a `currentStep + 1` could we make it more descriptive?  e.g. 
```js
const moveToNextStep = () => {
  setCurrentStep(currentStep + 1)
}
```

Patrick-Erichsen (dismissed):
This cleanup is beautiful... 100000x easier to understand. Great job!

I left some comments in regards to location tracking differences between the new Provider and `LocationService`, but I think this is good to merge. Feels like it should be future work to circle back and consolidate things.

app/PermissionsContext.js
johnschoeman: Thanks for the context!

Given that it only exists during the onboarding flow and is not really a device permission would it make sense to remove it from the permissions context?

app/views/onboarding/OnboardingPermissions.js
johnschoeman: sure, looks a bit ruby-ish to my eye, but happy to go with preferring the one liners.

app/PermissionsContext.js
johnschoeman: Thanks for the write up!

> To summarize - I think both this file and LocationService are providing two slightly different sets of permission info. This ContextProvider is purely app-specific device permissions, whereas LocationService also takes into account device and user-level statuses.

This makes a lot of sense to me as a distinction.

app/PermissionsContext.js
johnschoeman: yes! good catch.

app/PermissionsContext.js
johnschoeman: I'm not sure I follow the question, but i dont believe so.  as these status are determined by the components state and the initialState assigned above is not.

the initial state i have above is mostly forward looking to if typescript is ever added to this file. that is we _could_ not provide an initial state as the value being passed to the provider on the render step will replace it immediately after the useEffect resolves, but this wouldnt type check, hence the need to initialize the context with an initial state of the same structure.

app/PermissionsContext.js
```diff
@@ -0,0 +1,170 @@
...
+      const isDev = __DEV__;
+      await Promise.all([
+        isGPS ? checkLocationPermission() : null,
+        isiOS ? checkNotificationPermission() : null,
+        // TODO(https://pathcheck.atlassian.net/browse/SAF-232): Put HCA auto sub logic behind a feature flag
+        isDev && isiOS ? checkAuthSubscriptionPermission() : null,
+      ]);
+    };
+    checkAllPermissions();
+  }, []);
```
tstirrat, line 84: Enable eslint in your editor to catch these:

```
React Hook useEffect has a missing dependency: 'checkLocationPermission'. Either include it or remove the dependency array.
```

app/PermissionsContext.js
tstirrat: Yes, it makes sense to remove it from "permissions". But I think it is helpful to know if the value is set to false or not (undefined).. i.e. when upgrading to a new version

tstirrat (dismissed):

app/PermissionsContext.js
johnschoeman: hmmm, not sure whats going on with the linting rule but I think its mistaken, checkLocationPermission is a constant and wont make sense to include as a dependency to the useEffect hook?

johnschoeman: yeah, fiddled around with it a bit. looks like the rule is inferring if a variable can change over the lifecycle by absence of any hooks or something.  I refactored the code to get around this, but in general i dont think its best practice to blindly follow linters, ultimately a human wrote them and they can be fallible.

mattThousand (approved):

app/PermissionsContext.js
Patrick-Erichsen: Ah yeah I see where I was confused. 👍

PR #912, opened 2020-05-26
Update ios icons and splash screen for bt app

johnschoeman:
Why:
Currently but the gps and bte versions of the app are using the same
icons and splash screen assets. We would like to have the apps be
visually distinct.

This commit:
Adds a new asset set for the bte target and moves the existing asset set
to be only included in the gps target.

### icons on screen.

Screen Shot 2020-05-26 at 12 19 10 PM


tparesi (dismissed):
LGTM

tstirrat (approved):

PR #913, opened 2020-05-26
Remove eslint action from workflow

johnschoeman:
Why:
Currently the tinovyatkin/action-eslint@8b32772f075413f9192f43641e03d9bd0abae621
action fails silently and does not fail for the same reasons as `yarn
lint`

This commit:
Removes the action until we can troubleshoot it.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,12 +23,6 @@ jobs:
 
       - run: yarn --frozen-lockfile
 
-      # v1 is broken, the given sha was the latest master (but we dont want to track master)
-      - uses: tinovyatkin/action-eslint@8b32772f075413f9192f43641e03d9bd0abae621
-        with:
-          repo-token: ${{secrets.GITHUB_TOKEN}}
-          check-name: eslint # this is the job name from above 👆
-
       # The above eslint action does not show you in the job log what failed,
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -23,12 +23,6 @@ jobs:
 
       - run: yarn --frozen-lockfile
 
-      # v1 is broken, the given sha was the latest master (but we dont want to track master)
-      - uses: tinovyatkin/action-eslint@8b32772f075413f9192f43641e03d9bd0abae621
-        with:
-          repo-token: ${{secrets.GITHUB_TOKEN}}
-          check-name: eslint # this is the job name from above 👆
```
tstirrat, line 30: It's nice to get the annotations inline, so perhaps we can do `continue-on-error`:

https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

And then we can rely on `yarn lint` to pass/fail the check (remove the `if: always` from that check)

tstirrat (changes requested):

johnschoeman:
figured out the issue on this one: https://github.com/Path-Check/covid-safe-paths/pull/909, closing as this is no longer necessary.

PR #914, opened 2020-05-26
chore(Fix build) Fix iOS staging ipa artifact path

tstirrat:
Fix the broken CI on merge into `develop`.

[Failing because of an incorrect file path](https://github.com/Path-Check/covid-safe-paths/runs/709914047)

#### Linked issues:

NA

#### How to test:

Merge and let's see! :)


johnschoeman (approved):
🥇

PR #916, opened 2020-05-27
Lokalise latest may 27

alpita-masurkar:
1) Added a new language file- Spanish (Latin America)- es_419 and updated languages.js. Please review this to make sure it is added correctly.
2) NOTE: I have not updated ios/COVIDSafePaths.xcodeproj/project.pbxproj. This file is missing the mapping of several languages including the ones I had submitted in an older PR to fix this (so these languages might not appear for ios). I don't know if it was intentional to not have those languages. But if we need this, I can add them in this PR or create a new one for the ios mapping.


app/locales/es_419.json
```diff
@@ -0,0 +1,25 @@
+{
```
tstirrat, line 1: please ensure that the `_display_name` key is provided in lokalise

tstirrat (changes requested):

app/locales/es_419.json
alpita-masurkar: added it.

tstirrat (approved):

PR #917, opened 2020-05-27
Update display and executable names for ios

johnschoeman:
Update display and executable names for ios …
d1dbc5a
Why:
Currently the display name and product name for both the GPS and BT
version of the app are both 'COVIDSafePaths'. We would like the display
names to be Path Check GPS and Path Check BT respectively for the GPS
and BT targets. Additionally, while it technically is okay to have the
product/executable names be the same for the targets, this can be
confusing.

This commit:
Updates the Product Name and Bundle Display Names for the two targets in
xcode. We also removed the COVIDSafePathsConfig.m files for the targets since we are now using react-native config to set .env variables in the preprocessing step rather than at run time over the bridge. No new functionality has been added.

### Before

Screen Shot 2020-05-26 at 12 19 10 PM

### After

Screen Shot 2020-05-27 at 10 39 59 AM

Screen Shot 2020-05-27 at 10 44 44 AM


mattThousand (dismissed):

mattThousand (dismissed):

mattThousand (approved):

johnschoeman:
lol, thank you @mattThousand !!!

PR #920, opened 2020-05-27
Tp bt onboard copy updates

tparesi:
#### Description:

This PR addresses two things:

Commit 1: Updates copy for the BT app from [Figma](https://www.figma.com/file/79OwaKqHvSeriExpYzAvVN/PATHS-%2F-Exploration-%2F-April-%26-May-2020?node-id=3208%3A140)

Commit 2: Moves the GPS/BT decision logic to a single location for copy changes. This will make it easier to track these differences in a central location. There are still a few places of conditional component rendering at this time.

#### Linked Cards:

Addresses Jira Card: [SAF-507](https://pathcheck.atlassian.net/secure/RapidBoard.jspa?rapidView=28&projectKey=SAF&modal=detail&selectedIssue=SAF-507)

#### How to test:

Click through the GPS and BT app to make sure there are no copy or UI regressions.


mattThousand (dismissed):

app/views/Settings.js
```diff
@@ -22,14 +21,18 @@ import LocationServices from '../services/LocationService';
 import { FEATURE_FLAG_SCREEN_NAME } from '../views/FeatureFlagToggles';
 import { GoogleMapsImport } from './Settings/GoogleMapsImport';
 import { SettingsItem as Item } from './Settings/SettingsItem';
+import {
+  isGPS,
+  settingsLoggingActiveText,
```
johnschoeman: What do you think about keeping these scoped in an object?

something like
```
import { Copy, Assets } from '../TracingStrategyAssets'
...

return (
  <View>
      <Image src={Assets.Onboarding1Backround} />
       <Text>{Copy.Onboarding1Title}</Text>
...
```

This makes is a little easier to keep undefineds and nulls from floating around and it tends to result in less churn in the imports.

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,80 @@
+import { Images } from './assets';
+import { config } from './COVIDSafePathsConfig';
+import languages from './locales/languages';
+
+export const isGPS = config.tracingStrategy === 'gps';
+
```
johnschoeman, line 7: I see this file as potentially being a pain to maintain over time.  Do you have a sense of exactly how much of the copy will be different between the two apps at this point?  if its about this much then i wouldnt worry about it, but if its going to be a lot i might consider trying to get something more programatic working.

do you think something like this would be viable?

```javascript
const tracingTranslate = (translation) => {
  const translation = isGPS ? languages.t(`${translation}_location`) : languages.t(`${translation}_bluetooth`)
  return translation
}
```

I think your idea from before was really good as well and it would be ideal if we could get it to work. I think the issue we had may have been resolved with the switch from passing the tracingStrategy from over the bridge to having it be a part of the preprocess step via react-native-config.

that is, something like: 

```javascript
const tracingTranslate = (translation) => {
  const translation = isGPS ? languages.t(translation).gps : languages.t(translation).bte
  return translation
}
```

or if we could get an api like this to work that would be even better:

```
\\ en.json
{
  ...
   "label": {
       "foo": "Foo"
       "launch_access": {
            "gps": "Enable Location",
            "bte": "Enable Bluetooth",
        },
      ...
```

```javascript
\\ MyComponent.js
  <Text>{t('label.foo.bar')[tracingStrategy]}</Text>
```
or maybe
```javascript
\\ MyComponent.js
  <Text>{t(`label.launch_access.${tracingStrategy}`)}</Text>
```


Didn't actually try it myself, so Ill defer to you if this is a good idea or not, but it would be nice to not need to maintain an additional assets file.

johnschoeman (dismissed):
Had a few thoughts on style and the api, if you're interested.

This looks good to me to ship though!

Even if we do want to make a pass at updating the api that we use to accomplish the switching of the copy per the tracing strategy, it might make sense to ship this pr as is to get the copy changes in sooner than later and make any refactor in a follow up commit.

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,80 @@
...
+export const onboarding4HeaderText = isGPS
+  ? languages.t('label.launch_screen4_header_location')
+  : languages.t('label.launch_screen4_header_bluetooth');
+export const onboarding4SubheaderText = isGPS
+  ? languages.t('label.launch_screen4_subheader_location')
+  : languages.t('label.launch_screen4_subheader_bluetooth');
+export const onboarding4ButtonText = isGPS
+  ? languages.t('label.launch_set_up_phone_location')
+  : languages.t('label.launch_set_up_phone_bluetooth');
+export const onboarding4NavDestination = isGPS ? 'Onboarding5' : 'Main';
```
johnschoeman:
```suggestion
export const onboarding4NavDestination = isGPS ? 'OnboardingPermissions' : 'Main';
```

app/TracingStrategyAssets.js
tstirrat: Please do not do string manipulation or use variables inside `t(x)` otherwise the `yarn i18n:extract` will not find things. This tool helps keep us sane with so many strings around.

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,80 @@
...
+import languages from './locales/languages';
+
+export const isGPS = config.tracingStrategy === 'gps';
+
+// Onboarding2
+export const onboarding2BackgroundImage = isGPS
+  ? Images.LaunchScreen2
+  : Images.LaunchScreen2BT; 
+export const onboarding2HeaderText = isGPS
+  ? languages.t('label.launch_screen2_header_location')
```
tstirrat: Generally `languages.t` is being phased out because its not dynamic especially in this exact use case.

Because these `t(...)` are all called outside of a component, they are fixed to the language at module load time.

Need to find a way to make the `t()` dynamic at usage/render time otherwise language changes in Settings.js will not be reflected dynamically.

Its **not pretty** but perhaps this module can export a _function_ that takes the `t` value:

```js
import { getAppSpecificAssets } from '../TracingStrategyAssets';

const {t} = useTranslation();
const { aboutHeaderText } = getAppSpecificAssets(t);
```

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,80 @@
+import { Images } from './assets';
+import { config } from './COVIDSafePathsConfig';
+import languages from './locales/languages';
+
+export const isGPS = config.tracingStrategy === 'gps';
```
tstirrat: move `isGPS` and `isBTE` to the config file.

tstirrat (changes requested):

app/TracingStrategyAssets.js
johnschoeman: This looks good to me!

app/TracingStrategyAssets.js
johnschoeman: good to know, I take back all my suggestions!

app/views/Settings.js
tparesi: I decided to go with a function, so that I can import `useTranslation` and load these values dynamically. Let me know what you think.

app/TracingStrategyAssets.js
tparesi: Thanks - I decided to make the TracingStrategyAssets a function so I could import and use the useTranslation hook. Let me know your thoughts about these changes. Thanks!

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,105 @@
+import { useTranslation } from 'react-i18next';
+import { Images } from './assets';
+import { config } from './COVIDSafePathsConfig';
+
+export function AppSpecificAssets() {
```
tstirrat: This is basically a hook now, so call it useAssets or similar.

Knowing it's a hook means it's important that it's always run in a predictable order.

app/TracingStrategyAssets.js
```diff
@@ -0,0 +1,105 @@
...
+  const { isGPS } = config;
```
tstirrat: Does this work?

`import {isGPS} from '../blah';`

If so that's more conventional.

Change here and throughout

app/COVIDSafePathsConfig.js
```diff
@@ -1,5 +1,6 @@
 import Config from 'react-native-config';
 
 export const config = {
+  isGPS: Config.TRACING_STRATEGY === 'gps',
```
tstirrat: Is tracingStrategy read anywhere anymore? Maybe all we need is plain isGPS/isBT exports? Those seem simpler to grok anyway

tstirrat (dismissed):
Great change, cleans up a lot of stuff. Thanks!

app/TracingStrategyAssets.js
tparesi: Thanks - will do!

tstirrat (dismissed):

mattThousand (approved):

PR #923, opened 2020-05-28
Global Store Introduction

JacobJaffe:
#### Description:

Adds a global redux store to the app, with persistence & an action / reducer for replacing the on boarding async storage flow. 

The motivation for this is to set up the infrastructure for sharing of cross-component state in a way that is more easily consumable by different screens & components, ensures a shared, normalized data layer for our upcoming APIs, and allows us to more gracefully handle (and ommit) loading states.

Additionally, we can debug our coupled state better (I've set up RN debugger integrations for Redux Dev Tools here). 

Replacing the onboarding async store with a persited redux state allows us to have a state of "onboardingComplete" at mount time for the navigator, stopping navigator flash between onboarding and main. Follow up is needed to keep the splash screen up during the persist gate loading.

#### Screenshots:
Behavior of the app load with the persist gate (note the button text flash on the before):

Before             |  After
:-------------------------:|:-------------------------:
 | 

#### How to test:
Ensure that going through the onboarding flow can only occur once. 
No app functionality should change.
Can additionally now use Redux Dev Tools / RN Debugger to ensure that redux state is persisted / updated properly after onboarding flow.


app/Entry.js
```diff
@@ -28,58 +27,59 @@ import { SettingsScreen } from './views/Settings';
...
+  headerShown: false,
+};
 
-  useEffect(() => {
-    async function checkDone() {
-      /** @type {boolean | undefined | null} */
-      const flag = await GetStoreData(ONBOARDING_DONE, false /* isString */);
-      setOnboardingDone(!!flag);
-    }
+const MainApp = () => (
```
JacobJaffe, line 77: This abstraction will make swapping out for a Tab Navigator easier in upcoming design changes

app/store/migrations/index.js
```diff
@@ -0,0 +1,5 @@
+const migrations = {
+  0: prevState => prevState,
```
JacobJaffe: no-op setup so that we have boilerplate setup

app/views/onboarding/Onboarding4.js
```diff
@@ -97,4 +97,7 @@ const styles = StyleSheet.create({
   },
 });
 
-export default Onboarding;
+const mapDispatchToProps = dispatch => ({
```
johnschoeman: what do you think about using useDisptach and and useReducer hooks? I have been seeing most projects moving away from the connected component style now that hooks are available as they easier to reason about.

a blog post that explains the reasoning a bit if youre interested: https://thoughtbot.com/blog/using-redux-with-react-hooks

app/store/reducers/onboardingReducer.js
```diff
@@ -0,0 +1,19 @@
+import { ONBOARDING_COMPLETE } from '../actions/constants';
+
+const initialState = {
+  complete: false,
+};
+
+const onboardingReducer = (state = initialState, action) => {
```
johnschoeman:
```suggestion
const onboardingReducer = (state = initialState, action) => {
  Object.freeze(state)
```

johnschoeman (dismissed):
Looks good! A few minor comments.

---

I tend to be bearish on introducing redux as i find it tends to become a mess as its challenging to be diligent about making eventful actions rather than commandful actions and you loss the benefit of Pub/Sub design and just have a lot of boilerplate to maintain. And it inevitably tends to attract more and more of the application state over time. (its also a pain to use with typescript)

I've been seeing a lot of projects move away from redux for these reasons.

That said I do think redux has its place, but mostly for very complicated state management things, like undo/redo, drag n drop list reordering, etc.

But I do find it overkill if all you need is to share some state between components on the tree.

> The motivation for this is to set up the infrastructure for sharing of cross-component state in a way that is more easily consumable by different screens & components, ensures a shared, normalized data layer for our upcoming APIs, and allows us to more gracefully handle (and ommit) loading states.

Given your goal is to share information between components, what do you think about using just the state sharing portion of redux, that is a react context, directly to solve this?

app/views/onboarding/Onboarding4.js
JacobJaffe: I'm a fan of the hooks -- thought that onboaridng4 was a Class component still, I'll swap that over!

JacobJaffe:
@johnschoeman I fully get the bearishness on introducing redux.

> loss the benefit of Pub/Sub design

Generally I agree with this, but we've already lost it with our abstractions to classes to share the state via store writes (e.g. previous onboarding status), and for holding api states (e.g. HA list)

I think this is an improvement here because the shared data was _already_ floating around without strong couplling, but now we can have stronger invariants on it, debugging tools, and the ability to write stronger tests (as well as mock app states easier)

>  Given your goal is to share information between components, what do you think about using just the state sharing portion of redux, that is a react context, directly to solve this?

This is half the goal (fully the goal in _this_ pr, which aims to set up the infrastructure). The other half is to provide infrastructure to interact with HA API data in a consistent way. E.g. for the HA List that is needed for the subscription & the sharing step, to provide both a context provider for the data _and_ an api / action layer to ensure we are consistent in how we are fetching / storing from the api.

johnschoeman:
> This is half the goal (fully the goal in _this_ pr, which aims to set up the infrastructure). The other half is to provide infrastructure to interact with HA API data in a consistent way. E.g. for the HA List that is needed for the subscription & the sharing step, to provide both a context provider for the data _and_ an api / action layer to ensure we are consistent in how we are fetching / storing from the api.

Got it, thanks for the explanation! I still have my concerns with redux, but I certainly defer to your thoughts on if its a good fit for the use case.

app/Entry.js
```diff
@@ -28,61 +27,62 @@ import { SettingsScreen } from './views/Settings';
 
 const Stack = createStackNavigator();
 
-export const Entry = () => {
-  const [onboardingDone, setOnboardingDone] = useState(false);
+const screenOptions = {
```
tstirrat: CONST_CASE for module scope that will never change.

app/Entry.js
tstirrat: Love it, thanks

app/Entry.js
```diff
@@ -28,61 +27,62 @@ import { SettingsScreen } from './views/Settings';
...
+    <Stack.Screen name='Onboarding4' component={Onboarding4} />
+    <Stack.Screen
+      name='OnboardingPermissions'
+      component={OnboardingPermissions}
+    />
+  </Stack.Navigator>
+);
+
+export const Entry = () => {
+  const onboardingComplete = useSelector(state => state.onboarding.complete);
```
tstirrat: Extract the selector into a function, so that store structure doesn't matter to the consumer and could be refactored under the hood.

Also, in tests you can:

```js
jest.spyOn(selectors, 'getOnboardingComplete').mockReturnValue(true);
```

Side note. does useSelector do memoization? or will we need reselect?

Update: did some research, it [does not do memoization](https://react-redux.js.org/api/hooks#equality-comparisons-and-updates). Let's use a basic select function for now, and look at reselect in a follow up PR

app/store/actions/onboardingAction.js
```diff
@@ -0,0 +1,7 @@
+import { ONBOARDING_COMPLETE } from './constants';
+
+const onboardingCompleteAction = () => ({
```
tstirrat: Consider redux-actions: https://redux-actions.js.org/api/createaction

app/store/reducers/onboardingReducer.js
tstirrat: I think we can turn on some dev mode only enhancers that enforce for immutable states. Rather than doing it manually in every reducer.. thoughts?

tstirrat (changes requested):
Great stuff! This will be excellent for improving app perceived responsiveness

app/store/actions/onboardingAction.js
tstirrat: Mostly suggesting this so there isnt a file for action names, and another for action creators and payloads. etc.

tstirrat: Redux toolkit is officially recommended, has createAction with payloads, typesafe createReducer handlers, and includes reselect for selectors:

https://redux-toolkit.js.org/api/createAction

app/store/reducers/onboardingReducer.js
tstirrat: redux toolkit includes some middleware that checks immutable state:

https://redux-toolkit.js.org/api/getDefaultMiddleware#included-default-middleware

tstirrat:
Also checked and redux toolkit uses an `immer.js` like pattern where you can mutate the state object without worry and it will always store new refs:

https://redux-toolkit.js.org/api/createReducer#direct-state-mutation

JacobJaffe:
@tstirrat  updated with Redux Tool kit!
@johnschoeman in doing so introduced the freezing as middleware

this also now includes a bunch of prettier changes because of the husky pre commit hook

johnschoeman (approved):

tstirrat (approved):

PR #924, opened 2020-05-28
General Code Cleanup

Patrick-Erichsen:
#### Description:

**Note**: This PR was originally to fix bugs on the Location Status indicator, but the UI for that element has changed and we are removing it altogether. I would still like to include some general cleanup that I did in the PR however.

- Fixes onboarding bug where selecting `Deny` on location permissions blocks the user from continuing
- Moves `useFlags` into a new `services/hooks` folder
- Converts `LocationData` into a few functions instead of a class that needs to be instantiated
- Creates an `index.js` for the components folder to allow for group imports. Cleans up all files that use a component to use a singular group import

#### Linked issues:

- https://pathcheck.atlassian.net/browse/SAF-239

#### Screenshots:

N/A

#### How to test:

Walk through app and ensure no screens get a red box.


app/constants/__tests__/Theme.spec.js
```diff
@@ -1,7 +1,7 @@
 import { render } from '@testing-library/react-native';
 import React from 'react';
 
-import { Typography } from '../../components/Typography';
+import { Typography } from '../../components';
```
johnschoeman, line 4: 💯

app/views/Settings/GoogleMapsImport.js
```diff
@@ -14,30 +14,35 @@ export const GoogleMapsImport = ({ navigation }) => {
...
-      <ParagraphContainer>
-        <Typography use='body3' secondary monospace>
-          {t('import.google.disclaimer')}
-        </Typography>
-      </ParagraphContainer>
-    </>
-  );
+  // GoogleMapsImport is only enabled on the GPS app
+  if (config.tracingStrategy === 'gps') {
+    return (
```
johnschoeman: For component returns I find it nice to avoid statements and use expressions if possible as they tend to be easier to reason about/less simple bugs like forgetting a 'return'.  What do you think about something like:
```javascript
return (
  <>
    { config.tracingStrategy === 'gps' ? <GoogleMapsImport /> : null }
  </>
```

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+        style={{ width: '100%' }}
+        secondary
+        small
+      />
+    </>
+  );
+};
+
+const LocationStatusSwitch = ({ canTrack, setLocTrackingStatus }) => {
+  console.log('LocationStatusSwitch!');
```
johnschoeman: extra console.log?

app/views/onboarding/Onboarding5.js
```diff
@@ -19,10 +19,12 @@ import { SvgXml } from 'react-native-svg';
 
 import { isPlatformiOS } from './../../Util';
```
johnschoeman: Sadly a lot of this file will be in conflict with #909 , let me know if you'd like to pair/help on the updates there!

app/views/Settings/GoogleMapsImport.js
johnschoeman: Also what do you think about pulling the conditional up altogether?  That is the settings page is the component that determines if the GoogleImports should be rendered at all.

johnschoeman (dismissed):
Overall looks good to me! a few minor comments, if you are interested.

Thanks for grabbing the the components imports, that refactor is 💯 

I dont have a lot of context on the location stuff, so I cant speak to much more than the code quality, but it all reads really well to me. Awesome work!

app/views/onboarding/Onboarding5.js
Patrick-Erichsen: Yeah I started this before I saw that PR. I plan on just copy/pasting all of your changes over mine 👍

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: Yup! Thanks

app/views/Settings/GoogleMapsImport.js
Patrick-Erichsen: Yeah, I was curious to hear other's thoughts on this. Agreed on the ugly conditional render - my thought with pulling it into the component rather than the parent Settings page was that regardless of where the `<GoogleMapsImport />` is, it should never render is it isn't the GPS build.

The big tradeoff there is that it's way less apparent in the parent component. 

I think I prefer your snippet - same thing goes for the new `<LocationPermission />` component being introduced here, so I'll make the same update there.

app/constants/storage.js
```diff
@@ -1,6 +1,7 @@
 export const LOCATION_DATA = 'LOCATION_DATA';
 export const CONTACT_DATA = 'CONTACT_DATA';
-export const PARTICIPATE = 'PARTICIPATE';
+export const IS_APP_LOCATION_TRACKING_ENABLED =
```
tstirrat: This could still be confused with app level location permissions... Perhaps a rename to `IS_LOCATION_TRACKING_OPT_IN`

app/services/HCAService.js
```diff
@@ -9,7 +10,7 @@ import {
 } from '../constants/storage';
 import { GetStoreData, SetStoreData } from '../helpers/General';
 import languages from '../locales/languages';
-import { LocationData } from './LocationService';
+import LocationService from '../services/LocationService';
```
tstirrat: same dir, so `./`

app/services/HCAService.js
```diff
@@ -149,7 +189,7 @@ class HCAService {
    * @returns {[{authority_name: [{url: string}, {bounds: Object}]}]} List of health care authorities
    */
   async getAuthoritiesFromUserLocHistory() {
-    const locData = await new LocationData().getLocationData();
+    const locData = await LocationService.getLocationData();
```
tstirrat, line 192: Yes, love this, was never sure whether LocationData needed to be a singleton.

app/services/LocationService.js
```diff
@@ -3,107 +3,35 @@ import PushNotificationIOS from '@react-native-community/push-notification-ios';
...
-   * A valid box has a `ne` and `sw` field that each contain a valid GPS point
-   * @param {*} region - Object to validate
-   */
-  isValidBoundingBox(region) {
-    if (!region.ne || !this.isValidPoint(region.ne)) {
-      console.error(`invalid 'ne' field for bounding box: ${region.ne}`);
-      return false;
-    }
+export const Reason = {
+  //Location services are disabled for the device
```
tstirrat: If you make these `/** */` comments then vscode will annotate the enums on hover.

app/services/LocationService.js
```diff
@@ -226,27 +152,33 @@ export default class LocationServices {
...
   }
 
   static async getHasPotentialExposure() {
     const dayBin = await GetStoreData(CROSSED_PATHS, false);
     return !!dayBin && dayBin.some(exposure => exposure > 0);
   }
 
-  static async getParticpating() {
-    return await GetStoreData(PARTICIPATE, false);
+  static async setStoreLocationTracking(isEnabled) {
```
tstirrat: Make this more clear this is the user opt in. `setUserOptIn()` perhaps?

app/services/LocationService.js
```diff
@@ -226,27 +152,33 @@ export default class LocationServices {
...
+    await SetStoreData(IS_APP_LOCATION_TRACKING_ENABLED, isEnabled);
+  }
+
+  static async getAppLocTrackingStatus() {
```
tstirrat: `getUserOptInStatus`?

app/services/LocationService.js
```diff
@@ -260,44 +192,44 @@ export default class LocationServices {
...
         canTrack: false,
-        reason: Reason.LOCATION_OFF,
+        reason: Reason.APP_NOT_AUTHORIZED,
         hasPotentialExposure,
         isRunning,
       };
     }
 
-    if (authorization != BackgroundGeolocation.AUTHORIZED) {
+    if (!locTrackingStatus) {
```
tstirrat: See here is where its ambiguous.. which status?

app/services/LocationService.js
```diff
@@ -260,44 +192,44 @@ export default class LocationServices {
...
       return {
         canTrack: false,
-        reason: Reason.NOT_AUTHORIZED,
+        reason: Reason.USER_DISABLED,
```
tstirrat: Perhaps since you're renaming, `USER_OPT_OUT`

app/services/LocationService.js
```diff
@@ -260,44 +192,44 @@ export default class LocationServices {
...
         hasPotentialExposure,
         isRunning,
       };
     }
 
     return {
       canTrack: true,
-      reason: '',
+      reason: Reason.USER_ENABLED,
```
tstirrat: Perhaps this can be undefined.... its not really true because of only this.. its true because of all 3 perconditions met: enabled location, app has permissions, user is opt in.

app/services/LocationService.js
```diff
@@ -324,4 +256,17 @@ export default class LocationServices {
...
+
+  static async getLocationData() {
+    return NativeModules.SecureStorageManager.getLocations();
+  }
+
+  /**
+   * Returns the most recent point of location data for a user.
+   * This is the last item in the location data array.
+   */
+  static async getMostRecentUserLoc() {
```
tstirrat: Use the whole `Location`

app/services/hooks/useLocTrackingStatus.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Hook used to check the location tracking status of a user
+ * @param {{canTrack: boolean, reason: Reason, isRunning: boolean, hasPotentialExposure: boolean}} status
+ */
+export const useLocTrackingStatus = () => {
+  const initialState = {
+    canTrack: undefined,
+    reason: undefined,
+    isRunning: undefined,
+    hasPotentialExposure: undefined,
```
tstirrat: This probably shouldn't be here anymore... but.. can remove in a different PR

app/services/hooks/useLocTrackingStatus.js
```diff
@@ -0,0 +1,54 @@
+import BackgroundGeolocation from '@mauron85/react-native-background-geolocation';
+import { useEffect, useState } from 'react';
+
+import LocationService from '../LocationService';
+
+/**
+ * Hook used to check the location tracking status of a user
+ * @param {{canTrack: boolean, reason: Reason, isRunning: boolean, hasPotentialExposure: boolean}} status
+ */
+export const useLocTrackingStatus = () => {
```
tstirrat: `Loc` isnt great. maybe just `useTrackingStatus` or `useGpsTrackingStatus`

app/services/hooks/useLocTrackingStatus.js
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  const [locTrackingStatus, setStatus] = useState(initialState);
+
+  /**
+   * Set the location tracking status, and start/stop BackgroundGeolocation accordingly
+   * @param {boolean} newIsTracking
+   * @returns {{canTrack: boolean, reason: Reason, isRunning: boolean, hasPotentialExposure: boolean}}
+   */
+  const setLocTrackingStatus = async newIsTracking => {
```
tstirrat: `const setUserOptIn = async userOptIn => { ... }`

app/views/Settings/GoogleMapsImport.js
tstirrat: I also prefer: `{isGPS && <GoogleMapsImport />}`

and +100 to pulling up:

It makes more sense to me that `Settings.js` owns the switch logic, not this component:

- It's odd for a component to conditionally switch itself entirely to null.
- Settings has far more reason to know about gps/bte context.

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+    reason === Reason.USER_ENABLED || reason === Reason.USER_DISABLED;
+
+  const isMissingLocPermissions =
+    reason === Reason.DEVICE_LOCATION_OFF ||
+    reason === Reason.APP_NOT_AUTHORIZED;
+
+  const getCurrentRowDirection = i18n.dir() === 'rtl' ? 'row-reverse' : 'row';
+
+  const getLocStatusElement = () => {
+    console.log(`reason: ${reason}`);
```
tstirrat: remove console logs

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+    console.log(
+      `hasLocTrackingPermissions: ${hasLocTrackingPermissions} isMissingLocPermissions: ${isMissingLocPermissions}`,
+    );
+
+    if (hasLocTrackingPermissions) {
```
tstirrat: Already inside the locStatus component, this could be just hasPermissions.

Personally, prefer zero abbreviations outside of well known acronyms. The code is minified, so you'll save the space in the real app.

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+export const LocationTrackingStatus = () => {
+  const { i18n } = useTranslation();
+  const [locTrackingStatus, setLocTrackingStatus] = useLocTrackingStatus();
+  console.log(locTrackingStatus);
+
+  const { reason, canTrack } = locTrackingStatus;
+
+  const isGPS = config.tracingStrategy === 'gps';
+
+  const hasLocTrackingPermissions =
```
tstirrat: Can this just be the inverse of the isMissingPermissions?

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+        />
+      );
+    }
+
+    if (isMissingLocPermissions) {
+      return <EnableLocationButton reason={reason} />;
+    }
+  };
+
+  if (isGPS) {
```
tstirrat: Here;s the gps/bte switch thing again... move up to Settings.js

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+        const deviceSettingsUrl = 'App-prefs:';
+        Linking.openURL(deviceSettingsUrl);
+      } else {
+        DeviceSettings.location();
+      }
+    }
+  };
+
+  return (
+    <>
```
tstirrat: fragment seems not needed.

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+    </>
+  );
+};
+
+const LocationStatusSwitch = ({ canTrack, setLocTrackingStatus }) => {
+  console.log('LocationStatusSwitch!');
+  const { t, i18n } = useTranslation();
+  const [label, setLabel] = useState('');
+
+  useEffect(() => {
```
tstirrat: Extract this whole text mutation/animation thing into a hook/component so that the state and timers are all isolated.

```jsx
const loadingLabel = useInProgressText(label, 1000);

// or

<Typography><AnimatedText value={label} interval={1000} /></Typography>
```

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+    let labelInterval;
+
+    /**
+     * Provide a basic animation to show that location logging
```
tstirrat: This comment was slightly unclear so I had to read the code anyway.

Perhaps:

```
Takes an input label and animates a 3 dot suffix, updating each second.

Example:

Loading
Loading.
Loading..
Loading...
Loading
Loading.
```

app/views/About.js
```diff
@@ -38,9 +37,10 @@ export const AboutScreen = ({ navigation }) => {
       BackHandler.removeEventListener('hardwareBackPress', handleBackPress);
   }, [navigation]);
 
-  const aboutHeaderText = config.tracingStrategy === 'gps'
-    ? t('label.about_header_location')
-    : t('label.about_header_bluetooth');
+  const aboutHeaderText =
+    config.tracingStrategy === 'gps'
```
tstirrat: Used a lot, move to `import {isGPS} from 'config';`

app/views/Settings/LocationTrackingStatus.js
```diff
@@ -0,0 +1,184 @@
...
+    const createLabelInterval = () => {
+      return setInterval(() => {
+        setLabel(label => {
+          if (label.slice(-3) === '...') {
+            setLabel(t('label.logging_active_location'));
+          } else {
+            setLabel(label + '.');
+          }
+        });
+      }, 1000);
```
tstirrat: Perhaps the interval can be configured when it's a hook.

tstirrat (changes requested):
Great work! Looking good.

app/views/Settings/GoogleMapsImport.js
Patrick-Erichsen: Cool - thanks for the feedback 👍 I have implemented this.

app/constants/storage.js
Patrick-Erichsen: 👍

app/services/LocationService.js
Patrick-Erichsen: Good call

app/services/LocationService.js
Patrick-Erichsen: I like it

app/services/LocationService.js
Patrick-Erichsen: 👍

app/services/LocationService.js
Patrick-Erichsen: Also renamed `USER_ENABLED` to `USER_OPT_IN`

app/services/hooks/useLocTrackingStatus.js
Patrick-Erichsen: I'm fine to just remove now. The only reason I included it is because it currently matches the shape of the object returned from `LocationService.checkStatus()`, so I thought it was clearer.

But if we just remove it here there isn't really a downside, so if we plan to remove it from `checkStatus()`, I'll just remove this.

app/services/hooks/useLocTrackingStatus.js
Patrick-Erichsen: Agreed - I like `useGpsTrackingStatus`

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: Yup - `loc` feels awkward all around.

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: So the way that it works now is that since `setUserOptIn()` is async, it is initially `undefined`.

So we essentially have three return cases:

1. User has permissions
2. User does not have permissions
3. Waiting on the permission status to resolve

In cases 1 & 2, we return a component. In case 3, we don't want to return anything or else we end up with a flicker between the button and the switch components if the state is different once resolved.

Feels hacky/unclear though so open to feedback on how to make it clear.

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: 👍

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: Good catch

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: I like the animated text approach. Going to throw it in `/components`, but let me know if you think it should just stay in this file.

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: I think this is a good solution 

```js
const getGpsStatusElement = () => {
    if (reason === undefined) {
      return;
    }

    if (hasPermission) {
      return (
        <GpsStatusSwitch canTrack={canTrack} setUserOptIn={setUserOptIn} />
      );
    } else {
      return <EnableGpsButton reason={reason} />;
    }
  };
```

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: Thanks for pushing back on this - ended up being simple to implement and is 10x cleaner/easier to understand.

I went with `<AnimatedText />` but the return value feels a little awkward:

```js
return <>{text}</>;
```

app/views/Settings/LocationTrackingStatus.js
Patrick-Erichsen: Done

app/views/About.js
Patrick-Erichsen: Good call

app/services/LocationService.js
Patrick-Erichsen: As in - just get rid of this method altogether?

app/services/LocationService.js
Patrick-Erichsen: This gets kind of tricky for the reasons mentioned in [this comment ](https://github.com/Path-Check/covid-safe-paths/pull/924#discussion_r432173992).

What about just giving it a somewhat better description of `ALL_CONDITIONS_MET`? Still vague, but at least it isn't misleading.

app/services/LocationService.js
tstirrat: I meant dont shorten the word to "Loc"

tstirrat (dismissed):

Patrick-Erichsen:
Closing this PR and will pass off relevant logic to @vlasac once he begins work on https://pathcheck.atlassian.net/browse/SAF-284?jql=text%20~%20%22home%20indicator%22

Patrick-Erichsen:
After some discussion, reopening as per the updated description.

Going to wait to merge until after #945 to avoid creating conflicts on that PR.

app/services/hooks/useAssets.js
```diff
@@ -1,14 +1,15 @@
 import { useTranslation } from 'react-i18next';
-import { Images } from './assets';
-import { isGPS } from './COVIDSafePathsConfig';
+
+import { Images } from '../../assets';
```
johnschoeman: Im less of a fan of grouping by implementation and more of a fan of grouping by feature. And in general, I try to keep relative paths shorter as a metric to determine approachability to the code base.

What do you think about not moving this file?

johnschoeman (dismissed):
Continues to look good to me!

One thought, we could have shipped the component file reorganization, which is 💯 btw, in a separate commit/pr as its logically separate form the location changes. This helps keep a git history that is easy for future readers to understand and allows us to ship more frequently, which is especially useful on projects with a lot of contributors. I wouldnt change anything on this pr though!

Patrick-Erichsen:
@johnschoeman yup, totally agree on your comment about doing the component file cleanup in a separate PR, this one got a little sloppy. 

I also agree with your sentiment on keeping `TracingContactStrategy` in the root and have reverted that change 👍 

This should be ready for another rubber stamp in the morning and hopefully merging.

johnschoeman (approved):
🚢

PR #927, opened 2020-05-28
Update Lokalise push to use --poll

tstirrat:
The cli api from 2.5.0+ uses non-blocking operations be default. `--poll` is now required to wait until the upload is complete.

#### How to test:

Ensure the push job completes after the files are uploaded.


tstirrat:
Successful push here: https://github.com/Path-Check/covid-safe-paths/pull/927/checks?check_run_id=718031674

efalkner (approved):
✅

PR #919, opened 2020-05-27
Feature/ios/location hashing

MichaelNeas:
#### Description:

Adds the native iOS component to support storing and retrieving scrypt hashes based on geohash and timestamp. Includes Realm migration with the addition of a hashes column, which is a List of Strings containing all possible hashes for the given timestamp. No migration code needs to go inside the block but the schema version must increase and there must exist a migrationBlock. https://realm.io/docs/swift/latest/#migrations

This PR brings in and slightly modifies a geohash struct found here: https://github.com/nh7a/Geohash/blob/master/Sources/Geohash/Geohash.swift

CryptoSwift implementation of Scrypt was performing too poorly.  Added Swift Package Managed wrapper above libscrypt found here: https://github.com/greymass/swift-scrypt

#### Linked issues:

Android PR with corrected comments: https://github.com/Path-Check/covid-safe-paths/pull/907
Previous iOS PR: https://github.com/Path-Check/covid-safe-paths/pull/898
Previously the thought was to bridge a saveLocation method from the react side but that brought up the issue that the RN dependency actually saves it's own SQLite DB of location data.  This required us to provide a transform the the library and implement the hashing natively on iOS and Android.

iOS component for:
#807
#806

#### How to test:

Run  ios/COVIDSafePathsTests/

PR #898, opened 2020-05-22
Feature/ios/secure store hash

MichaelNeas:
#### Description:

Adds the native iOS component to support storing hashes from JS.  Includes Realm migration with the addition of hashes column, which is a List of Strings containing all possible hashes for the given timestamp. No migration code needs to go inside the block but the schema version must increase and there must exist a migrationBlock. https://realm.io/docs/swift/latest/#migrations 

The new bridged saveLocation method takes a dictionary and parses location data coming from JS.  The parse method refactors the previous import logic to support all properties from the background location callback and maintain backwards compatibility for minimum import requirements.

#### Linked issues:

iOS component for:
https://github.com/Path-Check/covid-safe-paths/issues/807
https://github.com/Path-Check/covid-safe-paths/issues/806

#### Screenshots:

N/A

#### How to test:

Run iOS XCTests
RealmSecureStorageTest.testSaveLocation()
LocationTests.testImportParsing()
LocationTests.testSaveParsing()

Issue #807, opened 2020-05-12
[FEATURE-REQ] SHA256 of geohashes in the intersection code


tremblerz:
## Issue type

This builds on top of #806. This is going to be a change that we need to apply to the intersect logic. See the old PR for reference - [relevant file diff from the PR](https://github.com/Path-Check/covid-safe-paths/pull/248/files#diff-1980a789ee94a1b6e356aeeccc4f87aa)
When we load the location data from the Asyncstorage, instead of reading the raw data, we need to read from Geohashes. The next step is as follows -
1. read geohashes
2. Append timestamp to the geohash
3. Apply SHA256([geohash, timestamp])
4. Compare it with the received points in the intersect.js

diarmidmackenzie:
My view is that these should use a slow hash algorithm, and scrypt is my favoured approach for various reasons detailed here:

https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/61112371/Design+for+MVP1+HA+JSON+Changes

summetj:
@tremblerz  - If I'm reading your issue description correctly, #806 is the first step in this process that can be done independently, and this can be added only after #806 is finished. Did I understand that correctly?

diarmidmackenzie:
For anyone who wants to work on tis please see Jira SAF-268 for a proposed alternate plan of attack.

Same basic outcome, but proposing different sequencing to ensure Safe Paths & Safe Places retain compatibility as much as possible during this change.

yonatanmk:
To clarify, we're going to hash the concatenation of the geohash and timestamp using Scrypt, not SHA256

Issue #806, opened 2020-05-12
[FEATURE-REQ]: geo-hashing and SHA256


tremblerz:
Right now we obtain GPS data in raw (latitude, longitude) and store it in the encrypted format (Not sure if PR is merged). Another thing we want to add in this storage is the storage of Geohash of the raw data in **7 bit** precision (which is equivalent of the 76m accuracy). For the first pass let's only store one geohash and in subsequent PR we will be storing geohash of the corners or edges of the squares.

## Issue type

Feature request for relatively tighter security.

diarmidmackenzie:
I am now quite strongly against 7 char precision, and think it needs to be 8 char (even though 8 char precision depends on GPS accuracy in Safe Paths that we have not achieved yet - see SAF-175).

A 7 char geohash in an urban center will cover a dozen or more businesses.  See illustrations in here:
https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/61145553/Moving+Safe+Paths+Safe+Places+to+use+geohashes+natively

Even moving to a 30 min detection window won't prevent false positives from people who spent 30 mins in the other 11 or so uninfected businesses.

Also, 7 char geohashes reduce entropy by 32x and hence make the hashed data 32x easier to crack - so actually makes privacy worse than the more specific 8 char geohashes.

diarmidmackenzie:
For anyone who wants to work on tis please see Jira SAF-268 for a proposed alternate plan of attack.

Same basic outcome, but proposing different sequencing to ensure Safe Paths & Safe Places retain compatibility as much as possible during this change.

tstirrat:
https://pathcheck.atlassian.net/browse/SAF-268
https://pathcheck.atlassian.net/browse/SAF-269

PR #928, opened 2020-05-29
Feature/ios/scrypt hashes

MichaelNeas:
Description:

Previous PR got a little messy, getting the c library to work nicely.  I remade the PR for better commits.

Adds the native iOS component to support storing and retrieving scrypt hashes based on geohash and timestamp. Includes Realm migration with the addition of a hashes column, which is a List of Strings containing all possible hashes for the given timestamp. No migration code needs to go inside the block but the schema version must increase and there must exist a migrationBlock. https://realm.io/docs/swift/latest/#migrations

This PR brings in and slightly modifies a geohash struct found here: https://github.com/nh7a/Geohash/blob/master/Sources/Geohash/Geohash.swift

CryptoSwift implementation of Scrypt was performing too poorly (290 seconds for 4 hashes with the suggested N size). Added Swift Package Managed wrapper above libscrypt found here: https://github.com/greymass/swift-scrypt
Linked issues:

Android PR with corrected comments: #907
Previous iOS PR: #898 #919 
Previously the thought was to bridge a saveLocation method from the react side but that brought up the issue that the RN dependency actually saves it's own SQLite DB of location data. This required us to provide a transform the the library and implement the hashing natively on iOS and Android.

iOS component for:
#807
#806
How to test:

Run ios/COVIDSafePathsTests/


tstirrat (approved):

yonatanmk (approved):
I was able to get the location hashing working locally.

PR #932, opened 2020-05-29
Fix/permission flow

Namit-Agrawal:
#### Description:
Fixed the permission flow for android where screen advances immediately without waiting for the user to approve/deny location permissions

#### Screenshots:

Permission


app/PermissionsContext.js
```diff
@@ -87,9 +87,13 @@ const PermissionsProvider = ({ children }) => {
     if (Platform.OS === 'ios') {
       const status = await check(PERMISSIONS.IOS.LOCATION_ALWAYS);
       setLocationPermission(statusToEnum(status));
-    } else if (Platform.OS === 'andriod') {
+    } else if (Platform.OS === 'android') {
       const status = await check(PERMISSIONS.ANDROID.ACCESS_FINE_LOCATION);
-      setLocationPermission(statusToEnum(status));
+      if (locationPermission === PermissionStatus.UNKNOWN) {
```
tstirrat: should this be checking `status`?

Why doesn't `setLocationPermission(statusToEnum(status));` work?

tstirrat (changes requested):

app/PermissionsContext.js
Namit-Agrawal: @tstirrat It definitely works! The only reason I put that there was because there was initially a typo of "andriod" which meant it never entered that condition whenever the android app checked for permissions. When I fixed that, it always came out to denied initially when the android app booted which I am assuming is not the preferred behavior. So i added that if condition so that permissions for android is shown as "unknown" when the user first installs the app instead of denied. I can definitely change it back to the original.

johnschoeman: I read this as, if the location permission is unknown, then set the location permission to unknown, regardless of what the actual permission setting being found by check is. Which i dont think is the behavior we want in all cases. I would remove the if else statement here and address the initializing as denied bug in a separate pr.

One possibility on that would be going through the onboarding flow multiple times, are you reseting the permissions on each test?

Namit-Agrawal: @johnschoeman I am! I clear the data for each run I do so that I can get to the permission screens. I will remove the if and else and try to see why android is initializing it as denied

johnschoeman: Not sure exactly how you are clearing the data, but I believe that adb has a utility for reseting permissions: https://gist.github.com/Pulimet/5013acf2cd5b28e55036c82c91bd56d8#file-adbcommands-L73

I would give that a try and if the issue persists then you have found a bug!  and I would open an issue to capture that and we can go from there to fix it. Thanks for digging into this!

Namit-Agrawal: I see! I am essentially going to the app info > storage > clear data. This resets all the permissions for me.

johnschoeman (approved):
LGTM!

one thought, which i dont think you should implement here, but for any future readers, if we want to be able to allow the context consumer to chose whether or not to wait for the permission request, it might make sense to have `requestLocationPermission` itself be awaited at the call site. NBD at this point, but wanted to mention the option!

app/PermissionsContext.js
Patrick-Erichsen: So @namit-agrawal, I am also seeing that issue. The diagram below from [react-native-permissions](https://github.com/react-native-community/react-native-permissions#android-flow) should clarify why that is happening.
Screen Shot 2020-06-02 at 9 38 28 AM

Seems like a legitimate bug that we should address in another PR.

Patrick-Erichsen:
LGTM as well, nice work!

@tstirrat - I believe this is just waiting on a resolution of your previously requested changes.

Patrick-Erichsen:
@tstirrat mind taking one more peek?

johnschoeman:
@Namit-Agrawal sorry about this but all of these changes have now been added in other prs.  (and theres no way for me to dismiss @tstirrat 's review.) so going to close this.

Namit-Agrawal:
No worries!

PR #933, opened 2020-05-29
Remove location hashes from location upload JSON

yonatanmk:
#### Description:
https://github.com/Path-Check/covid-safe-paths/pull/925 and https://github.com/Path-Check/covid-safe-paths/pull/928 recently added location hashes to each location stored on the device. 

When we retrieve these locations to export them to safe places we want to make sure we do not upload the hashes as well. This PR makes sure that we only upload a location's `latitude` `longitude`, and `time` to safe places.

#### How to test:
You can call `console.log(jsonData)` in the `onShare` method in Export.js to verify that hashes are not uploaded


tstirrat (dismissed):

yonatanmk:
@tstirrat I had to rebase this branch since it had merge conflicts with develop. Could I get another 👍  on this and then I can merge it?

JacobJaffe (approved):

PR #934, opened 2020-05-29
E2E Transfer Flow

JacobJaffe:
#### Description:
Wires up the flow for the e2e. This sets them against our apiary server to handle as the api contracts are expected. 

This is all behind a feature flag. When the staging server is set up, we can add it to the HA list in a way that we can dynamically route the upload to HA's based on their api url's in the yaml. For now, this is hard coded to the apiary. 

1. Select an HA
2. Enter code
3. Consent to both agreements
4. Upload data

https://www.figma.com/file/79OwaKqHvSeriExpYzAvVN/PATHS-%2F-Exploration-%2F-April-%26-May-2020?node-id=3600%3A26288

#### Screenshots:

image
image
image
image
image
image
image

image

>What's with all these funky `flexGrows` ? 
Accessibility hacks for scroll views for content: 
image
 

#### How to test:
The primary thing to test with these is the UI/UX of the flows. We're awaiting the staging server to be able to handle a full e2e test with this.

- Test the primary flow, of going through the flow screen by screen. With the mock api's, any code will work right now.

- To test the fail cases, disable wifi while going through the screens

- To test the failed code (because of a _bad_ code, rather than a generic failure), you've gotta either change the endpoint, or change the condition in the code screen to fail on the 200. This will be easier when we have a working endpoint.

- This should be good across a variety of phone sizes. I've tested this on iPhone 11, 8 & 6, and Pixel 2


app/views/Settings.js
```diff
@@ -137,12 +137,23 @@ export const SettingsScreen = ({ navigation }) => {
                 description={t('label.event_history_subtitle')}
                 onPress={() => navigation.navigate('ExposureHistoryScreen')}
               />
-              <Item
-                label={t('share.title')}
-                description={t('share.subtitle')}
-                onPress={() => navigation.navigate('ExportScreen')}
-                last
-              />
+              <FeatureFlag
```
JacobJaffe, line 140: Keep our old mechanism of downloading, but default to this new one with feature flags.

app/views/Export/ExportSelectHA.js
```diff
@@ -0,0 +1,164 @@
...
+
+  return (
+    <Theme use='default'>
+      <StatusBar
+        barStyle='dark-content'
+        backgroundColor={Colors.VIOLET_BUTTON}
+        translucent={false}
+      />
+      <View style={styles.wrapper}>
+        <SafeAreaView style={{ flex: 1, paddingBottom: 0 }}>
```
JacobJaffe, line 56: The version of `SafeAreaView` we're using requires doing this double nesting of views.

app/views/Export/ExportSelectHA.js
```diff
@@ -0,0 +1,164 @@
...
+  wrapper: { flex: 1, backgroundColor: Colors.INTRO_WHITE_BG },
+  exportSectionTitles: {
+    fontWeight: '500',
+    color: Colors.VIOLET_TEXT_DARK,
+    fontFamily: fontFamily.primaryMedium,
+  },
+  card: {
+    backgroundColor: Colors.WHITE,
+    shadowColor: Colors.BLACK,
+    shadowOffset: {
```
JacobJaffe, line 153: This is a pretty arbitrary shadow. 
I like using https://ethercreative.github.io/react-native-shadow-generator/

app/Entry.js
```diff
@@ -28,6 +38,41 @@ import { SettingsScreen } from './views/Settings';
...
+
+const ExportStack = () => (
+  <Stack.Navigator
+    mode='modal'
+    screenOptions={{
+      ...screenOptions,
+      cardStyleInterpolator: fade,
+      gestureEnabled: false,
+    }}>
+    <Stack.Screen name='ExportStart' component={ExportStart} />
```
tstirrat, line 59: +1 to not naming them Export1, Export2, Export3

app/Entry.js
```diff
@@ -28,6 +38,41 @@ import { SettingsScreen } from './views/Settings';
...
+
+const screenOptions = {
+  cardStyleInterpolator: CardStyleInterpolators.forHorizontalIOS,
+  cardStyle: {
+    backgroundColor: 'transparent', // prevent white flash on Android
+  },
+  headerShown: false,
+};
+
+const ExportStack = () => (
```
tstirrat, line 51: This seems like it could go in Export.js as `getExportStack(Navigator, screenOptions)`, perhaps?

Happy to be overruled, it just seems weird to pull in all the steps and assemble them here, when it could be done there.

(And yes, maybe Onboarding dir should own it's stack)

app/views/Export/ExportCodeInput.js
```diff
@@ -0,0 +1,235 @@
...
+    <View style={{ flexDirection: 'row', flexShrink: 1 }}>
+      {[...digits].map((digit, i) => (
+        <TextInput
+          ref={ref => (digitRefs.current[i] = ref)}
+          key={`${i}CodeDigit`}
+          value={digit}
+          style={[
+            styles.digitInput,
+            {
+              borderColor: digit
```
tstirrat: if digit is 0 it will be false. perhaps `digit >= 0`

app/views/Export/ExportCodeInput.js
```diff
@@ -0,0 +1,235 @@
...
+                ? Colors.VIOLET_BUTTON
```
tstirrat: can you use theme colors here? That way it can work on dark/light/charcoal theme backrounds.

https://emotion.sh/docs/theming#usetheme-hook

```js
const theme = useTheme()

borderColor: digit ? theme.primary : theme.textSecondaryOnBackground
```

You may want to use theme.disabled or add some new constant to `theme.js`

app/views/Export/ExportCodeInput.js
```diff
@@ -0,0 +1,235 @@
...
+      Alert.alert(t('common.something_went_wrong'));
+      setIsCheckingCode(false);
+    }
+  };
+
+  return (
+    <Theme use='default'>
+      <StatusBar
+        barStyle='dark-content'
+        backgroundColor={Colors.VIOLET_BUTTON}
```
tstirrat: You can get the theme bg color via `useTheme().background`

app/views/Export/ExportConfirmUpload.js
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Alert, NativeModules } from 'react-native';
+
+import { Icons } from '../../assets';
+import exitWarningAlert from './exitWarningAlert';
+import ExportTemplate from './ExportTemplate';
+
+const MOCK_ENDPOINT =
+  'https://private-anon-da01e87e46-safeplaces.apiary-mock.com/upload';
```
tstirrat, line 10: maybe these urls could be a `.env` var?

General question.. when might we change them? would they change per environment/flavour.. or?

If static, leave as is

app/views/Export/ExportConfirmUpload.js
```diff
@@ -0,0 +1,58 @@
...
+  const [isUploading, setIsUploading] = useState(false);
+  const onClose = () => exitWarningAlert(navigation);
+
+  const { selectedAuthority, code } = route.params;
+
+  const upload = async () => {
+    setIsUploading(true);
+    try {
+      const concernPoints = await NativeModules.SecureStorageManager.getLocations();
+      console.log({ concernPoints });
```
tstirrat: remove console.log

app/views/Export/ExportConfirmUpload.js
```diff
@@ -0,0 +1,58 @@
...
+
+      const res = await fetch(`${MOCK_ENDPOINT}`, {
```
tstirrat: unnecessary template string.

app/views/Export/ExportLocally.js
```diff
@@ -15,18 +15,21 @@ import LinearGradient from 'react-native-linear-gradient';
...
+import { Theme } from '../../constants/themes';
+import { isPlatformiOS } from '../../Util';
 
 const base64 = RNFetchBlob.base64;
 
-export const ExportScreen = ({ navigation }) => {
+// NOTE:
+// This is the old way we export. This is still the default, but will
+// become flipped behind the feature flag once we have staging for uploading.
+const ExportLocally = ({ navigation }) => {
```
tstirrat, line 32: awesome, appreciate the rename to make stuff more clear.

app/views/Export/index.js
```diff
@@ -0,0 +1,8 @@
+export { default as ExportCodeInput } from './ExportCodeInput';
```
tstirrat, line 1: Why not just export named consts from the sub modules if you're gonna do the dance here?

tstirrat (dismissed):

app/views/Export/ExportConfirmUpload.js
JacobJaffe: These will be part of the selected HA eventually -- the HA retrieval will need to be a two step process to 
1. query the yaml file
2. query the HA's config endpoint

Once we have the staging server wired with a `cursor.json` file with the configuration meta data, we can wire this up completely

app/views/Export/ExportCodeInput.js
JacobJaffe: Misnomer on my part. It's really a character input, these are stored as a string for any generic character, so the evaluation is on existence. I updated the name to be all about `character` rather than `digit`

app/Entry.js
JacobJaffe: IMO it would be best to have a `Navigation` directory that handles constructing the `<Navigation/>` component for `App.js`. 

I think keeping the navigation creators together is useful so that we ensure we navigate between sections of the app consistently (e.g. with screen options), as navigator inheritance can be messy:

I think we should revisit this when we do the tab restructure, so it can be something like: 
```
navigation/
    index.js
    onboarding/
    mainTabs/
    export/
```

So that we can group construction under the `navigation` folder, and keep abstracted from views

app/views/Export/ExportCodeInput.js
JacobJaffe: It looks like the background mixes backgroundColor & gradient styled instructions?

I updated the color here to be consistent with the hard coded theming color of the background, for consistency on android with the background color

app/views/Export/index.js
JacobJaffe: Personally I like the internal / external abstraction better. 
IMO `ExportCodeInput` should export the view as the default without any context of how it'll be imported outside of the file to make the file more readable. 

I'm fine changing it to another export structure if we want though!

app/views/Export/ExportConfirmUpload.js
tstirrat: Ah, right. makes sense

app/views/Export/index.js
tstirrat: I mostly care that you can type `ExportCodeInput` and VSCode autocomplete will find and import the correct file. If your current export pattern works, great.

Generally default exports only work 50% of the time.

I get the impression this index.js will have sufficient enough naming that it will work.

app/views/Export/ExportCodeInput.js
```diff
@@ -88,22 +90,20 @@ const CodeInput = ({ code, length, setCode }) => {
 
   return (
     <View style={{ flexDirection: 'row', flexShrink: 1 }}>
-      {[...digits].map((digit, i) => (
+      {[...characters].map((character, i) => (
```
tstirrat: Why the array clone? .map will not mutate the original array.

tstirrat (dismissed):

app/Entry.js
tstirrat: Ah, makes sense. Thanks for the thoughts

app/views/Export/index.js
JacobJaffe: Ah, gothcya. Yeah, this works for auto importiing

app/views/Export/ExportCodeInput.js
JacobJaffe: good catch

tstirrat (approved):

PR #937, opened 2020-06-01
Update ios productname/headers to separate targets 

johnschoeman:
Why:
We would like to be able to build the BT version of the app without
including any GPS location specific code. Currently there is some
location specific code in AppDelegate.m and some location specific swift
files being included in the BT build. We would like to start the process
of segmenting the application code by target.

This commit:
Moves Location specific code to the GPS group. It separates
AppDelegate.m into two AppDelegates, one for each target. We set up
swift bridging headers for each target. Additionally we restored some
defaults to the library and framework search paths and updated the
Product names to match the Target names of GPS and BTE as this is the
default that is expected by xcode build process and allows us to use the
default search paths provided by xcode.

Note that this commit changes no functionality.

Note that the BTE target is still building with GPS only pods currently
that are being defined from the react-native side and will need to be
removed in the BTE target in a future commit. Specifically RN
LocationAlways and @maurn85/react-native-background-geolocation

#### How to test:

Build and run both the GPS and BT versions of the app


johnschoeman:
@MichaelNeas , we are having a bit of trouble getting Scrypt, which was introduced by your commit here: https://github.com/Path-Check/covid-safe-paths/commit/6e6b00988572500db0449fd3165159b9fc454be2, to play nicely with the ios builds here.

Do you have any time today to assist?

MichaelNeas:
@johnschoeman : https://github.com/Path-Check/covid-safe-paths/pull/939

ios/Podfile.lock
```diff
@@ -514,6 +508,6 @@ SPEC CHECKSUMS:
   SSZipArchive: fa16b8cc4cdeceb698e5e5d9f67e9558532fbf23
   Yoga: f2a7cd4280bfe2cca5a7aed98ba0eb3d1310f18b
 
-PODFILE CHECKSUM: da76f32974b5fc83f551b81bc61d97931055a2ec
+PODFILE CHECKSUM: 310a468cb175b9046abc49527d4f28cb0a5a27ab
 
-COCOAPODS: 1.9.1
+COCOAPODS: 1.8.4
```
tstirrat: `cd ios && bundle`

Should give you a later cocoapods

tstirrat (dismissed):

ios/COVIDSafePaths.xcodeproj/project.pbxproj
```diff
@@ -403,7 +400,7 @@
 				C53FD09724719AD1006D3268 /* CoreGraphics.framework in Frameworks */,
 				C53FD09824719AD1006D3268 /* CoreData.framework in Frameworks */,
 				C53FD09924719AD1006D3268 /* Accelerate.framework in Frameworks */,
-				68EFB21324856EC6003D84F3 /* Scrypt in Frameworks */,
+				68EFB21324856EC6003D84F3 /* (null) in Frameworks */,
```
tstirrat: these (null) in x are suspicious

package.json
```diff
@@ -153,47 +153,47 @@
     },
     "configurations": {
       "iphone11.sim": {
-        "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/COVIDSafePaths.app",
+        "binaryPath": "ios/build/Build/Products/Release-iphonesimulator/GPS.app",
```
tstirrat: `test.yml` needs to be updated too:

https://github.com/Path-Check/covid-safe-paths/blob/develop/.github/workflows/test.yml#L217-L220

tstirrat (dismissed):

ios/COVIDSafePaths.xcodeproj/project.pbxproj
johnschoeman: they are! working through getting it all in order.

johnschoeman:
@mattThousand , when you have a moment, could I get a review? Was able to figure out the ios tests, but the android checks are still broken, as this is just updating ios things, I think its better to get this in and tackle the android issues in a separate pr. thoughts?

ios/BTE/Info.plist
```diff
@@ -42,19 +42,11 @@
...
+	<string>This app uses Bluetooth to securely collect and share random IDs with nearby devices. The app can use these IDs to notify you if you&apos;ve been exposed to COVID-19</string>
 	<key>NSBluetoothPeripheralUsageDescription</key>
-	<string>This app uses Bluetooth to securely collect and share random IDs with nearby devices. The app can use these IDs to notify you if you've been exposed to COVID-19</string>
-	<key>NSLocationAlwaysAndWhenInUseUsageDescription</key>
-	<string>Your location history will be saved on your device, and only shared with others if you choose to do so.</string>
-	<key>NSLocationAlwaysUsageDescription</key>
-	<string>Your location history will be saved on your device, and only shared with others if you choose to do so.</string>
-	<key>NSLocationWhenInUseUsageDescription</key>
-	<string>Your location history will be saved on your device, and only shared with others if you choose to do so.</string>
+	<string>This app uses Bluetooth to securely collect and share random IDs with nearby devices. The app can use these IDs to notify you if you&apos;ve been exposed to COVID-19</string>
```
mattThousand: Do we still need this permission string given that we're shifting over to `EN`?

ios/GPS/Tmp.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  Tmp.swift
+//  GPS
+//
+//  Created by John Schoeman on 5/29/20.
+//  Copyright © 2020 Path Check Inc. All rights reserved.
+//
+
+import Foundation
```
mattThousand: Is this file still necessary or can it be deleted?

mattThousand:
2 small questions, otherwise LGTM 🎉

mattThousand (dismissed):

mattThousand:
> @mattThousand , when you have a moment, could I get a review? Was able to figure out the ios tests, but the android checks are still broken, as this is just updating ios things, I think its better to get this in and tackle the android issues in a separate pr. thoughts?

@johnschoeman yes - the android tests may wind up being a larger chunk of work. Cut a ticket for us to address it in the near future if you would be so kind?

johnschoeman:
@mattThousand and changes are in! sorry to keep buggin ya, but could i have another review?

mattThousand (approved):

johnschoeman:
actually noticing that reruning the test work flows fixed it, perhaps we have a flakey set of tests  😭

PR #938, opened 2020-06-01
[SAF-272] - Safe Paths updates intersection logic to use hashes instead of GPS points

vlasac:
#### Description:

This PR changes the intersection logic used to determine if the app user has crossed paths with an infected person. It is an implementation of the algorithm described in [this Confluence doc](https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/81231882/Hashing+of+GPS+points+-+Design) created by Diarmid Mackenzie.

This PR also includes logic for paginated fetching of HA data described in [SAF-275](https://pathcheck.atlassian.net/browse/SAF-275).

#### How to test:
- install the app and wait for it to record some local data points
- manually add a mock HA server url to the list of subscribed HAs
- wait for the app to trigger the intersection calculation
- if the random data published by the mock HA matches enough of your local data point hashes, you should get an exposure notification


app/helpers/Intersect.js
```diff
@@ -207,11 +52,17 @@ export function normalizeAndSortLocations(arr) {
   if (arr) {
     for (let i = 0; i < arr.length; i++) {
       let elem = arr[i];
-      if ('time' in elem && 'latitude' in elem && 'longitude' in elem) {
+      if (
+        'time' in elem &&
+        'latitude' in elem &&
+        'longitude' in elem &&
+        'hashes' in elem
+      ) {
```
JacobJaffe, line 60: I realize you didn't write this original code, but I'd suggest breaking this into two steps for readability here:

```js
cosnt isLocationValid = (location) => (
        'time' in location &&
        'latitude' in location &&
        'longitude' in location &&
        'hashes' in location
);

const normalizeLocation = (location) => (
 ... 
);

export function normalizeAndSortLocations(locations) {
  return locaitons
               .filter(isLocationValid)
               .map(normalizeLocation)
               .sort((a, b) => a.time - b.time))
}
```
Then these can be broken out of the intersect logic, and abstracted to the file reading or api service layer, to deal directly with data.

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
-      i = k + 1;
-    } else if (cmp < 0) {
-      n = k - 1;
+/**
+ * Returns an array of data points within the exposure window
+ *
+ * @param {array} localDataPoints
+ * @param {int} exposureWindowDays
+ */
+export function discardOldData(
```
JacobJaffe, line 81: Does this need to be exported?

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
+  localDataPoints = [],
```
JacobJaffe, line 82: If this isn't used outside of this file, maybe these default params can be removed to reduce confusion?

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
+  exposureWindowDays = MAX_EXPOSURE_WINDOW_DAYS,
+) {
+  dayjs.extend(dayOfYear);
```
JacobJaffe, line 85: This is a plugin config I think?
Can this happen at the import at top of file, instead of in the function?

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
+ * data points for that day, otherwise zero.
+ *
+ * This method should be called before filling the gaps
+ * in the local points data because the filled data would
+ * generate the same result as filling day bins with zeros.
+ *
+ * @param {array} localDataPoints
+ * @param {int} exposureWindowDays
+ */
+export function initLocationBins(
```
JacobJaffe, line 105: Same thoughts around default params + exports

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
+ *
+ * @param {array} localData - array of stored gps points with gaps
+ * @param {int} gpsPeriodMS - local points sampling period
+ */
+export function fillLocationGaps(
+  localData,
+  gpsPeriodMS = MIN_LOCATION_UPDATE_MS,
+) {
+  // helper function that creates and populates an array with correct timestamps
+  const generateGapPoints = (startTime, gapSize, gpsPeriodMS) =>
```
JacobJaffe, line 137: can this be defined outside of this function?

app/helpers/Intersect.js
```diff
@@ -221,32 +72,233 @@ export function normalizeAndSortLocations(arr) {
...
+ * that are in the HA's list of concern points.
+ *
+ * This method does not calculate the exposure durations,
+ * as that can be done only after all of the HA data chunks
+ * have been analyzed.
+ *
+ * @param {array} localGPSDataPoints - array of recorded GPS data points.
+ * @param {Set} concernPointHashes - set of hashed concern points issued by health authorities
+ */
+export function updateMatchFlags(localGPSDataPoints, concernPointHashes) {
```
JacobJaffe, line 174: Does this need to be done mutatively?
Looking at the usage, I'm worried this adds brittleness because of the order of operations needed. 
Could this return a list of matching points instead?

app/helpers/Intersect.js
```diff
@@ -309,44 +361,83 @@ async function asyncCheckIntersect() {
...
-        // Update each day's bin with the result from the intersection.  To keep the
-        //  difference between no data (==-1) and exposure data (>=0), there
-        //  are a total of 3 cases to consider.
-        dayBins = dayBins.map((currentValue, i) => {
-          if (currentValue < 0) return tempDayBin[i];
-          if (tempDayBin[i] < 0) return currentValue;
-          return currentValue + tempDayBin[i];
-        });
       } catch (error) {
         // TODO: We silently fail.  Could be a JSON parsing issue, could be a network issue, etc.
```
JacobJaffe: We've introduced more places this can fail now (more async online calls. Is there a game plan for improving the network call retries?)

app/helpers/Intersect.js
```diff
@@ -309,44 +361,83 @@ async function asyncCheckIntersect() {
...
 
-  // get the health authorities
-  let authority_list = await GetStoreData(AUTHORITY_SOURCE_SETTINGS);
+  const authorities = await getHealthAuthorities();
 
-  if (authority_list) {
-    // Parse the registered health authorities
-    authority_list = JSON.parse(authority_list);
+  // we also need locally saved data so we can know the last read page for each HA
+  let localHAData = await GetStoreData(AUTHORITY_SOURCE_SETTINGS, false);
```
JacobJaffe, line 380: Can we give this a more descriptive name / What kind of data is this? 
Maybe we can call it `lastReadPageByHa`?

app/helpers/Intersect.js
```diff
@@ -309,44 +361,83 @@ async function asyncCheckIntersect() {
...
+  // get the saved set of locations for the user, already sorted, and fill in the gaps
   let locationArray = await NativeModules.SecureStorageManager.getLocations();
+  locationArray = discardOldData(locationArray, MAX_EXPOSURE_WINDOW_DAYS);
+  // generate an array with the asked number of day bins
+  let tempDayBins = initLocationBins(MAX_EXPOSURE_WINDOW_DAYS, locationArray);
+  locationArray = fillLocationGaps(locationArray, gpsPeriodMS);
 
-  // get the health authorities
-  let authority_list = await GetStoreData(AUTHORITY_SOURCE_SETTINGS);
+  const authorities = await getHealthAuthorities();
```
JacobJaffe: need error handling around this

app/helpers/Intersect.js
```diff
@@ -390,12 +493,11 @@ function notifyUserOfRisk() {
 /**
  * Return Json retrieved from a URL
  *
- * @param {*} url
+ * @param {string} url
  */
 async function retrieveUrlAsJson(url) {
```
JacobJaffe, line 498: Can we add this to the API folder?

JacobJaffe (changes requested):
Various comments --

Also, it's really difficult currently knowing where the test coverage is for all the util functions -- 
For all the helper functions, what do you think about we abstracting these into their own files with their own spec files?


e.g. 
```
interset/
  fillDayBins/
    index.js
    spec.js
```
Or something akin. Mostly to solve being able to ensure we have unit tests for all these really specific utill helper functions on core app functionality

app/helpers/Intersect.js
vlasac: I totally agree, nice improvement :)

app/helpers/Intersect.js
vlasac: Nope, only other place where this is called is the test file.

app/helpers/Intersect.js
vlasac: I can make not-mutative, but it will still return the whole list of points with the `hasMatch` attribute set to true or false. Is that behaviour acceptable?

JacobJaffe (approved):

JacobJaffe (approved):

PR #943, opened 2020-06-02
statically linked libscrypt

MichaelNeas:
#### Description:

Due to continuous issues with pulling in libscrypt using SPM we need to use a statically linked version of the c library.  All the files are directly ported from: https://github.com/technion/libscrypt/tree/a402f4116245ce8677b3d9f4f87096b5ccbe26e9

The issue that required this change is due to: https://forums.swift.org/t/swift-package-manager-and-custom-build-configurations/29181/12

In our build machines we use XCode11 which at the time, SPM did not have support for any configurations outside of Debug and Release.  The solutions revolving around making SPM work would be incredibly hacky. 

By including the C library directly we gain the ability to use the Scrypt implementation without the SPM overhead.

The reason we have to use Scrypt is outlined here: https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/81231882/Hashing+of+GPS+points+-+Design#The-Hash-Calculation

Any swift implementation proved to not meet our performance requirements.  So we have to use a lower level language to not drain battery and waste cycles on users phones.

#### Linked issues:

https://github.com/Path-Check/covid-safe-paths/pull/928 : describes more reasons for bringing this in, in the first place
https://github.com/Path-Check/covid-safe-paths/pull/937 : brings up new reasons to port over to a static lib

#### How to test:

All tests remain unchanged, this is essentially a "hot swap" of how the project knows where our library is.


MichaelNeas:
@johnschoeman @JacobJaffe @yonatanmk @efalkner @tstirrat @mattThousand This should fix up the staging build artifact issue.

johnschoeman (approved):
LGTM,

So another option here could have been to use a more recent version on xcode in our CD boxes?

would it makes sense to do this if it allows us to use the swift package manager?

My understanding of the ExposureNotification service that we are working on implementing is that it's required to be using xcode 11.5.1 beta so this is something that we will likely need to do anyways for the BT work.

MichaelNeas:
@johnschoeman if we do upgrade to xcode 11.5.1 beta the SPM integration would be flawless!  I was able to pin down the xcode build issues by using different xcode versions locally.  The future integration would only require deleting the libscrypt folder and adding back the SPM module :) 

I also don't have write access so feel free to merge this in.

I don't believe the `Test / e2e-iphone-se` cancel/fail is due to these changes.

johnschoeman:
> @johnschoeman if we do upgrade to xcode 11.5.1 beta the SPM integration would be flawless! I was able to pin down the xcode build issues by using different xcode versions locally. The future integration would only require deleting the libscrypt folder and adding back the SPM module :)

Sounds good @tstirrat @efalkner I think well need to get 11.5.1 beta into the github workflow for the building is this something you can help out on?

> I also don't have write access so feel free to merge this in.
> 
> I don't believe the `Test / e2e-iphone-se` cancel/fail is due to these changes.

Currently we are blocking all merges that dont pass these required tests, .... so we will need to actually get that to pass before merging in. :/

MichaelNeas:
@johnschoeman https://github.com/Path-Check/covid-safe-paths/pull/937/checks?check_run_id=731238470#step:11:9 demonstrated the same behavior

johnschoeman:
> @johnschoeman https://github.com/Path-Check/covid-safe-paths/pull/937/checks?check_run_id=731238470#step:11:9 demonstrated the same behavior

Yeah, I think the requirement on this one building is new.  It makes sense to fix the test in a separate pr, imo: https://github.com/Path-Check/covid-safe-paths/pull/947

johnschoeman:
@MichaelNeas , I have this pr up for using xcode 11.5: https://github.com/Path-Check/covid-safe-paths/pull/948.

Assuming that theres no conflict in our devops setup and we are able to make the switch, what do you think about not merging this pr in?  Is there any work here that would be valuable outside of removing SPM?

MichaelNeas:
@johnschoeman That'd be sweet, no value in merging this PR if all goes well with XCode 11.5, also thanks for upgrading the xcode version! 😄

johnschoeman:
@MichaelNeas , #948. was merged, so closing this now!  Let me know if you need any more assistance getting swift stuff setup.  thanks again!

PR #944, opened 2020-06-02
Reintroduce manual link for mauron85

johnschoeman:
Why:
A previous commit introduced a manual link for
@mauron85/react-native-background-geolocation so that we could build it
only to the GPS target. This has caused some build issues for others.
And for sake of time we are going to autolink the library again and take
another pass at manually linking the library in a future commit.

This commit:
Reverts the manual linking of the library.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -223,6 +223,7 @@ android {
 }
 
 dependencies {
+    implementation project(':@mauron85_react-native-background-geolocation')
     // loading html files
     compile project(':react-native-local-resource')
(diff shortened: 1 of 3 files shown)


mattThousand (dismissed):

johnschoeman:
@JacobJaffe , are you still branching off of this pr? is it still necessary to have this merged in?

mattThousand (approved):

johnschoeman:
closing as this as the linking was not actually an issue.

PR #945, opened 2020-06-02
Remove In-App Location Permission Setting

JacobJaffe:
#### Description:
Removals enabling / disabling locations in the app, instead deferring to the app's permissions. 
Removes the `USER_OFF` reason overall from the app for location services.

#### Linked issues:

https://pathcheck.atlassian.net/browse/SAF-521

#### Screenshots:

image

#### How to test:
Disabling / enabling location services from app setting should now trigger the off screen. 
Current test suite for location services should still pass (removed tests for flows not used now)


app/views/main/UnknownPage.js
```diff
@@ -49,15 +46,15 @@ export const UnknownPage = () => {
           </View>
           <View style={styles.contentBelowPulse}>
             <Text style={styles.mainTextBelow}>
-              {t('label.home_unknown_header')}
+              {t('label.home_setting_off_header')}
             </Text>
             <Typography style={styles.subheaderText}>
               {t('label.home_unknown_subtext')}
```
JacobJaffe: Separate issue, this doesn't have an attached `.en` string. 
This file should never be hit also.
 
I think ideally we should scrap this file and have pages for:
1. location off (for phone)
2. location permissions disabled for the app

app/views/Main.js
```diff
@@ -95,9 +95,13 @@ export const Main = () => {
...
     }
   } else {
-    if (location.reason === Reason.USER_OFF) {
+    if (
+      location.reason === Reason.LOCATION_OFF ||
+      location.reason === Reason.NOT_AUTHORIZED
+    ) {
       page = <OffPage />;
     } else {
+      // Invariant violation if this occurs
```
johnschoeman, line 104: Its not clear to me what Invariant violation means here?

app/views/main/UnknownPage.js
```diff
@@ -20,9 +20,6 @@ import { styles } from './style';
 export const UnknownPage = () => {
   const { t } = useTranslation();
   const buttonLabel = t('label.home_enable_location');
-  const handleEnableLocationPress = () => {
```
johnschoeman, line 23: 💯

johnschoeman (approved):
LGTM!

Patrick-Erichsen:
Thanks for doing this @JacobJaffe ! I have a question though - has it been decided that we want to remove the location status indicator on `<Settings />` altogether? 

My understanding was that the scope of this PR was just to remove the storage setting for user opt in/out. I feel like if the user has turned off location tracking for either this particular app, or their entire device, it still is valuable to show that in the UI. 

Thoughts?

Patrick-Erichsen (approved):
LGTM as per our conversation in standup.

PR #948, opened 2020-06-02
Use xcode 11.5 for workflow ios builds

johnschoeman:
Why:
We would like to be able to use the swift package manager and be on the
latest version of xcode to have the most up to date support and
security patches. Currently all of our github actions are building with
xcode 11.

This commit:
Updates the various actions to use Xcode_11.5.app, which is the most
recent version available on github virtual environments at the moment:

https://github.com/actions/virtual-environments/blob/43e26fa96b8e36c96dc4f825b349bd417625b92c/images/macos/macos-10.15-Readme.md

Note that for us to be able to use the ExposureNotification entitlement
recently released by apple we will need to build with Xcode 11.5 beta,
which isn't available on github actions at the moment.

#### How to test

Notice that the github actions that include building an ios binary work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,7 +156,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
-        run: sudo xcode-select -s /Applications/Xcode_11.app
+        run: sudo xcode-select -s /Applications/Xcode_11.5.app
 
       - name: Cache node_modules/
         uses: actions/cache@v1
@@ -231,7 +231,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
-        run: sudo xcode-select -s /Applications/Xcode_11.app
+        run: sudo xcode-select -s /Applications/Xcode_11.5.app
 
       - name: Cache node_modules/
         uses: actions/cache@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -74,7 +74,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
-        run: sudo xcode-select -s /Applications/Xcode_11.app
+        run: sudo xcode-select -s /Applications/Xcode_11.5.app
 
       - name: Cache node_modules/
         uses: actions/cache@v1
@@ -126,7 +126,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
-        run: sudo xcode-select -s /Applications/Xcode_11.app
+        run: sudo xcode-select -s /Applications/Xcode_11.5.app
 
       - name: Cache node_modules/
         uses: actions/cache@v1
@@ -182,7 +182,7 @@ jobs:
     needs: e2e-ios-build
     timeout-minutes: 15
     env:
-      DEVELOPER_DIR: /Applications/Xcode_11.4.app
+      DEVELOPER_DIR: /Applications/Xcode_11.5.app
     steps:
       - uses: actions/checkout@master
       - name: Cache node_modules/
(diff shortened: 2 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -182,7 +182,7 @@ jobs:
     needs: e2e-ios-build
     timeout-minutes: 15
     env:
-      DEVELOPER_DIR: /Applications/Xcode_11.4.app
+      DEVELOPER_DIR: /Applications/Xcode_11.5.app
```
johnschoeman, line 185: noticing this one is finding the application via setting the env.  We could probably change it to use the xcode-select utility, but I'm kinda hesitant to make more changes to build pipeline stuff than is necessary.

johnschoeman:
> Note that for us to be able to use the ExposureNotification entitlement
recently released by apple we will need to build with Xcode 11.5 beta,
which isn't available on github actions at the moment.

https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md#xcode

@mattThousand , @tstirrat , @efalkner this is something we will need to figure out.  Not sure how viable it is to reach out to github to include xcode beta in the virtual machines, we may need to switch to building locally for BT.

.github/workflows/build.yml
```diff
@@ -156,7 +156,7 @@ jobs:
       - uses: actions/checkout@master
 
       - name: Set XCode Version
-        run: sudo xcode-select -s /Applications/Xcode_11.app
+        run: sudo xcode-select -s /Applications/Xcode_11.5.app
```
efalkner, line 159: Does this suffice for the .beta need?

.github/workflows/test.yml
efalkner: I'd say if it passes with the same command as the other tools that's fine by me :) It may be that detox requires the environment variables set, but the easiest way to confirm that would be a test run, I think :)

efalkner:
@johnschoeman looks like 🍎 would reject an app build with a beta version of XCode
https://developer.apple.com/support/xcode/

efalkner (commented):
Looks like the simctl command needs to be updated too:

```
run: xcrun simctl create "iPhone SE" com.apple.CoreSimulator.SimDeviceType.iPhone-SE com.apple.CoreSimulator.SimRuntime.iOS-13-4
```

to

```
run: xcrun simctl create "iPhone SE" com.apple.CoreSimulator.SimDeviceType.iPhone-SE com.apple.CoreSimulator.SimRuntime.iOS-13-5
```

.github/workflows/build.yml
johnschoeman: I would be surprised if it did.

johnschoeman:
> Looks like the simctl command needs to be updated too:
> 
> ```
> run: xcrun simctl create "iPhone SE" com.apple.CoreSimulator.SimDeviceType.iPhone-SE com.apple.CoreSimulator.SimRuntime.iOS-13-4
> ```
> 
> to
> 
> ```
> run: xcrun simctl create "iPhone SE" com.apple.CoreSimulator.SimDeviceType.iPhone-SE com.apple.CoreSimulator.SimRuntime.iOS-13-5
> ```

thanks!

johnschoeman:
> @johnschoeman looks like 🍎 would reject an app build with a beta version of XCode
> https://developer.apple.com/support/xcode/

For the BT version of the app inorder to use the entitlement for the Exposure Notification service we will need to build with the Beta per apples instructions, so I think they will make an exception for contact tracing app submissions.  But you are correct, we will not be able to use the beta for the GPS version of the app.

tstirrat (dismissed):

tparesi (approved):
LGTM

PR #951, opened 2020-06-03
Update project tooling: typescript, prettier, husky

johnschoeman:
This pr introduces 3 changes to the projects developer tooling.

Given the large number of contributors to the project any change the the developer environment will likely cause some friction in the adoption process, so I am quite open to any or all these not being included, though I do believe that they will help our development process overall.

In particular adding typescript is a bigger change and I would like to make sure that theres no objections there before moving forward with it. Luckily typescript has a very good introduction story and is easy to give a try before committing to it for the entirety of the project.

Any or all of them could be rejected, or we could chose to introduce these changes separately.

### Commit 1
Introduce typescript …
e363304
Why:
Currently the project has struggled with a handful of bugs and
inconsistencies. Given the medical nature of the application it is
essential that we have confidence in our codebase acting in a reliable
way.

While critically important, high level end to end testing is difficult
and expensive to implement and maintain in the javascript ecosystem,
especially with cross platform technologies like React Native. We
currently have some test coverage, but we continue to need to add more.
Given the cost and difficulty of end to end test, is best practice to
rely on it for only the critical user flows and rely on unit tests and
static analysis for further test coverage as this balances the need to
maintain quality assurance and productivity in the development process.

Unit tests are easier to maintain, but currently our codebase does not
have every component or module covered. And again, these test, while
critically important, take time to implement and maintain. Additionally
unit test are better suited for ensuring higher level behavior of the
system, they are less suited for protecting against simple bugs such as
typos and implicit type coercion.

By introducing typescript we can quickly introduce some typesafety that
will help us maintain confidence that we are not shipping code that has
unexpected behavior. Typescript is incrementally adoptable and no
existing code needs to be updated immediately. To make a file be
captured by the typescript complier we need only change the extension
from .js to .ts.

This commit:
Adds the typescript configuration and adds a 'validate' script to
package json which will run the ts complier. Note that metro bundler
ships with typescript support and we are not introducing any new build
infrastructure.

A single file was chosen to convert to .ts arbitrarily to validate the
tsc process works.

A future commit should include running this validation in CI.

Note that given the build processes provided by metro bundler, there is
no need to emit js code into a bundle as one would do in a web context.
Hence the inclusion of the --noEmit flag.

### Commit 2

Include prettier in the validation script …
40fa850
Why:
Currently there are a handful of files that are not following the
project's prettierrc. This is likely due to individual developer
environments which are running prettier based off of global prettierrc
files rather than the project's. While not of major concern for the
functionality of the application, this leads to a lot of git churn and
is ultimately a drag on productivity due to unnecessary overhead in the
code review process.

This commit:
Adds a prettier check to the validation script, such that it will fail
if the files are not formatted according to prettierrc.

### Commit 3

Update husky to run validation script on push …
e9d0dd3
Why:
Currently husky is running eslint and prettier on commit. This can be a
drag on committing frequently discourages the practice. Additionally the
husky hooks do not run the jest test suite, meaning that it's likely
that prs can be opened which do not pass the jest test suite. Given that
this could be checked locally first, we are spending more CI cycles than
are necessary.

This commit:
Changes the husky git hoot to be pre-push rather than pre-commit and
adds yarn test to the process.


tparesi (dismissed):
I would be fine to add all three of these changes. Thanks for doing that.

efalkner (dismissed):
I'm all about this! Thanks for the introductions!

jest.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
Patrick-Erichsen: There is an existing config file in the `jest` folder I think we can just add the `moduleExtensions` to. 

https://github.com/Path-Check/covid-safe-paths/blob/develop/jest/config.js

tsconfig.json
```diff
@@ -0,0 +1,73 @@
...
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  },
+  "exclude": [
+    "node_modules",
+    "babel.config.js",
+    "metro.config.js",
+    "jest.config.js",
```
Patrick-Erichsen: If we decide to keep using just the `jest/config.js` file

```suggestion
    "jest/config.js",
```

Patrick-Erichsen (dismissed):
All of this sounds great to me! I think TS will be fantastic to have in a big & fast-moving project like this. 

The only comment I have is that I think we should include a bit of info about our usage of TS in the Readme and link to some official docs, e.g. https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html

The info about metro handling TS out of the box would probably be useful as well since that was one of the first questions that came to my mind, e.g. "Why are we using --noEmit?"

JacobJaffe (approved):
This is terrific!

> Update husky to run validation script on push …
Thanks especially for this. Breaking up PRs into commits has been difficult when appeasing lint. 

On TS, I agree this will be a big win for stability -- I think it would be valuable if we follow this up with an assessment of the most fragile data consuming layers to migrate to ts. What comes to mind immediately for me is the API payloads from the HA service, the published data, the concern points, and the exposure status.

johnschoeman:
> On TS, I agree this will be a big win for stability -- I think it would be valuable if we follow this up with an assessment of the most fragile data consuming layers to migrate to ts. What comes to mind immediately for me is the API payloads from the HA service, the published data, the concern points, and the exposure status.

Totally agree!  I would want to introduce ts to existing code as we need to update it for feature development though. Our main priority should still be to ship this to users asap, and I wouldn't want to lose sight of that by revisiting existing code *just* for the sake of typescript.

PR #953, opened 2020-06-03
Fixes incorrect app strings for Android.

mattThousand:
Fix incorrect strings (`app_name`, `app_name_short`, `notification_channel_description`) for GPS and BTE product flavors.


JacobJaffe (approved):
Thanks!

PR #954, opened 2020-06-03
Update Android Unit Tests with Timestamp Examples in Milliseconds

yonatanmk:
#### Description:
The location hashing algorithm overview has been updated to use UTC timestamps in units of milliseconds instead of seconds
https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/81231882/Hashing+of+GPS+points+-+Design#The-Hash-Calculation

This PR updates a single android unit tests that had an example location timestamp in seconds. This should clarify any confusion about what unit timestamps should be in. Moving forward all location timestamps should be in milliseconds in both Safe Paths and Safe Places

#### How to test:

Run the android unit tests locally. More specifically you could run the single test I've modified in `GeohashTests.kt`


johnschoeman (approved):
Dont have a lot of context here, but looks good to me!

PR #956, opened 2020-06-04
Fix GPS Staging Android APK builds

JacobJaffe:
The bundle step wasn't occurring for GPS staging!
The generated apk should work now!

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -93,10 +93,12 @@ import com.android.build.OutputFile
 
 project.ext.react = [
     entryFile: "index.js",
-    bundleInDevDebug: false,
-    bundleInStagingDebug: true,
+    bundleInGpsDebug: false,
+    bundleInBteDebug: false,
     bundleInGpsStaging: true,
-    devDisabledInStagingDebug: true,
+    bundleInBteStaging: true,
+    devDisabledInGpsStaging: true,
+    devDisabledInBteStaging: true,
     enableHermes: false,  // clean and rebuild if changing
 ]
 
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -95,6 +95,7 @@ project.ext.react = [
     entryFile: "index.js",
     bundleInDevDebug: false,
     bundleInStagingDebug: true,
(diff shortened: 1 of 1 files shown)


android/app/build.gradle
```diff
@@ -95,6 +95,7 @@ project.ext.react = [
     entryFile: "index.js",
     bundleInDevDebug: false,
     bundleInStagingDebug: true,
```
JacobJaffe: @efalkner I think we might be able to get rid of `bundleInStagingDebug`, if that's no longer used? double checking it's not used by BT

johnschoeman (dismissed):
Nice!

android/app/build.gradle
mattThousand: @JacobJaffe @efalkner presumably `bundleInBteStaging` should be added too? and `bundleInDevDebug` can also be removed as it is no longer a `buildvariantProductflavor` combination (`bundleInGpsDebug` and `bundleInBteDebug` should be added)

android/app/build.gradle
```diff
@@ -95,6 +95,7 @@ project.ext.react = [
...
+    bundleInGpsStaging: true,
     devDisabledInStagingDebug: true,
```
mattThousand, line 99: `devDisabledInStagingDebug` `devDisabledInGpsStaging`

also add `devDisabledInBteStaging`

android/app/build.gradle
```diff
@@ -95,6 +95,7 @@ project.ext.react = [
     entryFile: "index.js",
     bundleInDevDebug: false,
```
mattThousand:
```
project.ext.react = [
    entryFile: "index.js",
    bundleInGpsDebug: false,
    bundleInBteDebug: false,
    bundleInGpsStaging: true,
    bundleInBteStaging: true,
    devDisabledInGpsStaging: true,
    devDisabledInBteStaging: true,
    enableHermes: false,  // clean and rebuild if changing
]
```

mattThousand (changes requested):

mattThousand (approved):

PR #958, opened 2020-06-05
Jts update eslint and prettier

johnschoeman:
#### Commit 1

Update to prettier 2.0 and eslint with ts support …
517512d
Why:
Currently our prettier and eslint setup do not support JSX within .tsx
files. We would like to update our prettier and eslint config to handle
this.

Prettier 2.0 was released in march and includes better typescript
support and eslint support.

This commit:
Updates prettier to 2.0 and adds all the required eslint plugins to
support tsx files.

Note that the prettier-plugin-import-sort is not updated for prettier
2.0 yet so it was removed.

#### Commit 2

Use prettier 2.0 default settings …
2567c9c
Why:
The previous commit updated prettier to version 2.0.
https://prettier.io/blog/2020/03/21/2.0.0.html

The update included some changes to the default settings which largely
improve the development experience for typescript and we would like to
use these new defaults.

This commit:
Updates all of the files to use the new default settings.


mattThousand (dismissed):

tparesi (dismissed):
LGTM, but you will prob need another approval once the android unit tests pass. Just let me know!

mattThousand (approved):

PR #957, opened 2020-06-04
Request exposure notification authorization bt ios

johnschoeman:
Why:
We would like to use the exposure notification service provided by
apple in the application. In order to do so users must enable
authorization to the app. We would like to have this authorization
requested during the onboarding flow.

This commit:
Introduces a screen that will render in the BT app which will prompt the
user to enable Exposure Notifications. We introduced an
ExposureNotificationContext that contains the authorization state on the
js side and provides a way for components to request EN authorization by
calling into a Native Module which exposes a method for setting the
exposure notification to true in the iOS layer.

We also introduced the start of the Exposure Notification Service in the
BTE target and included the entitlements for Exposure Notifications.

Co-Authored-By: mattthousand <matt.buckley@rocketinsights.com>
Co-Authored-By: tparesi <thomas.paresi@thoughtbot.com>

#### iPhone

onboarding-en

PR #960, opened 2020-06-05
iOS Key Server API Client

mattThousand:
Why:
We would like the app to be able to communicate with the key server over the network.

This commit:
Introduces an interface between the Exposure Notifications iOS API on the device and the key server. It also introduces infrastructure to store Exposure Notifications-related data needed for future use on device.


mattThousand:
This PR is waiting on #957

johnschoeman (commented):
LGTM! no comments!

johnschoeman (approved):
LGTM! no comments!

PR #962, opened 2020-06-05
Add types to permissions context

johnschoeman:
Why:
We would like to have some type safety in the permissions context.

This commit:
Converts PermissionsContext.js to a tsx file.

### DIFF
--- a/app/services/HCAService.d.ts
+++ b/app/services/HCAService.d.ts
@@ -1,0 +1,3 @@
+import { HCAService as service } from './HCAService';
+
+export const HCAService = service;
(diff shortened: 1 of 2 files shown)


app/services/HCAService.d.ts
```diff
@@ -0,0 +1,3 @@
+import { HCAService as service } from './HCAService';
+
+export const HCAService = service;
```
johnschoeman, line 3: This is pretty lazy, but for sake of time I didn't do any actual declarations here. just got an inferred `any` by defining a new constant.

tparesi (approved):
LGTM

PR #963, opened 2020-06-05
New end-user consent screens

yonatanmk:
#### Description:
Implements new screens for onboarding end-user consent screens following Figma designs.

Current screen:
Screen Shot 2020-06-09 at 10 16 06 AM

New Screens:
Screen Shot 2020-06-09 at 10 14 38 AM
Screen Shot 2020-06-09 at 10 14 50 AM

#### Linked issues:
https://pathcheck.atlassian.net/browse/SAF-434


app/PermissionsContext.js
```diff
@@ -114,9 +114,9 @@ const PermissionsProvider = ({ children }) => {
 
   const requestLocationPermission = async () => {
     if (Platform.OS === 'ios') {
-      requestLocationForPlatform(PERMISSIONS.IOS.LOCATION_ALWAYS);
+      await requestLocationForPlatform(PERMISSIONS.IOS.LOCATION_ALWAYS);
```
johnschoeman: What is the reason for introducing this change? Was it to have the screen transition immediately? is this the intended behavior?

app/views/onboarding/Onboarding4.js
```diff
@@ -25,10 +27,11 @@ const Onboarding = (props) => {
   } = useAssets();
 
   const handleOnPressNext = () => {
-    const nextScreen = isGPS
-      ? 'OnboardingPermissions'
-      : 'EnableExposureNotifications';
-    props.navigation.replace(nextScreen);
+    if (isGPS) {
+      props.navigation.replace('OnboardingPermissions', { step: isiOS ? 'notification' : 'location' });
```
johnschoeman: is there a way we can do this without manually passing data through react-navigation? could the screen itself handle its conditional logic?

app/views/onboarding/OnboardingPermissions.js
```diff
@@ -16,37 +14,27 @@ import { Typography } from '../../components/Typography';
...
 import { SetStoreData } from '../../helpers/General';
 import languages from '../../locales/languages';
 import PermissionsContext, { PermissionStatus } from '../../PermissionsContext';
 import { sharedStyles } from './styles';
+import fontFamily from '../../constants/fonts';
 
 const width = Dimensions.get('window').width;
 
-export const OnboardingPermissions = ({ navigation }) => {
+export const OnboardingPermissions = ({ route, navigation }) => {
```
johnschoeman: why are we passing a route here?

app/views/onboarding/OnboardingPermissions.js
```diff
@@ -55,73 +43,48 @@ export const OnboardingPermissions = ({ navigation }) => {
...
+    nextStepParam: null,
+    handleButtonPress: handleRequestLocation,
+    icon: Icons.LocationPin,
+    buttonLabel: languages.t('label.launch_allow_location')
+  }
+  const steps = {
+    notification: notificationStep,
+    location: locationStep,
+  }
+  const currentStep = steps[route.params.step]
```
johnschoeman: ah I see the idea now. Yeah this totally works! Though in general i try to avoid relying on the navigation components for handling application data/logic, especially for mobile apps, as it increases the coupling between the screens making it harder to update flows in the future.  And its highly likely that we'll be updating theses flows as the product evolves.

What do you think about having the screen component itself determine what step it should be on?

johnschoeman (approved):
Overall looks good to me! 

One thought on the data passing.

And its somewhat hard to review the style changes without a before and after picture, could you add some screenshots?

Also, just to check, given the layout changes, did you take a look at the screen in both small and large format screens?

app/views/onboarding/OnboardingPermissions.js
yonatanmk: That certainly is an option.

The reason I've built it this way is so there's a clear transition between the request notification permission screen and the request location permission screen using the stack navigator.

app/PermissionsContext.js
yonatanmk: Doing this makes the `requestLocationPermission` logic consistent with the `requestNotificationPermission` so that the app doesn't progress to the next screen until the user selects an option from the location permission popup that appears.

app/views/onboarding/OnboardingPermissions.js
yonatanmk: I agree that this screen will probably be updated often in the future so I can look into changing it as you described.

PR #964, opened 2020-06-06
Connect Export To Staging Server & Redux Typescript

JacobJaffe:
#### Description:
Connects export to Staging Server!
Adds connection to the new trusted-authorities repo for HA selection during export.
Next steps are to use that everywhere. 

Currently the HA yaml is hard coded (fine because this is all feature flagged still), this should ideally be a env variable to point to different yamls depending on lane. 

This also makes everything under the `/store` folder typescript, so that our data layer is less brittle.

Adds linking to privacy policy. Some design divergence (whole line is a link instead of just 'here') here to make this accessible across localization. We can revisit this and introduce the i18next `trans` component as a follow up if we want to specify a word across locals.

#### Screenshots:

Redux Store Shape:
image (4)

image

#### How to test:
Go through the export flow! Using the staging server, create an access code for a case, and the upload process should work! There's a postman collection to make that easier if needed for creating)


App.js
```diff
@@ -1,13 +1,33 @@
...
+import { Provider } from 'react-redux';
+import { PersistGate } from 'redux-persist/integration/react';
 
 import { Theme } from './app/constants/themes';
 import { Entry } from './app/Entry';
 import { ExposureNotificationsProvider } from './app/ExposureNotificationContext';
 import { FlagsProvider } from './app/helpers/Flags';
 import { PermissionsProvider } from './app/PermissionsContext';
 import VersionCheckService from './app/services/VersionCheckService';
+import { createPersistedStore } from './app/store';
```
johnschoeman: looks like you'll need to rebase? wasnt this added in your previous commit?

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
johnschoeman: We can turn this lint rule off if we want, I was on the fence about adding it tbh.  I go back and forth between using type inference on function return types.  I hear arguments both ways, and dont really have a strong opinion.  Seemed okay to try out for exports, but if its slowing us down and not really adding much, we should remove it.

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -0,0 +1,52 @@
...
+    res.text(),
+  );
+  const { authorities } = Yaml.safeLoad(yamlString);
+  if (!Array.isArray(authorities)) {
+    throw new Error('authorities yaml did not return an array of authorities');
+  }
+  return authorities;
+};
+
+const getHealthcareAuthoritiesAction = () => async (dispatch: any) => {
```
johnschoeman: I believe the redux package exports a `Dispatch` type, could we remove the any here?

app/store/index.ts
```diff
@@ -1,18 +1,15 @@
...
 
-const composeEnhancers =
-  // eslint-disable-next-line no-underscore-dangle
-  (typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__) ||
-  compose;
-
-const enhancers = composeEnhancers(
-  applyMiddleware(createImmutableStateInvariantMiddleware()),
+const enhancers = composeWithDevTools(
+  applyMiddleware(thunk, createImmutableStateInvariantMiddleware()),
```
JacobJaffe, line 12: This `composeWithDevTools` is the exact same thing as the previous `composeEnhancers`, but removes the headache of us manually typing the `window` global for typescript

app/store/migrations/index.ts
```diff
@@ -0,0 +1,7 @@
+import type { MigrationManifest, PersistedState } from 'redux-persist';
```
JacobJaffe, line 1: Just a typescript migration

app/store/selectors/isOnboardingCompleteSelector.ts
```diff
@@ -0,0 +1,5 @@
+import type { RootState } from '../types';
```
JacobJaffe, line 1: Just a typescript migration

app/views/Export/ExportSelectHA.js
```diff
@@ -18,24 +19,19 @@ import { Typography } from '../../components/Typography';
...
-  const [authorities, setAuthorities] = useState(null);
-  const [selectedAuthority, setSelectedAuthority] = useState(null);
+  const dispatch = useDispatch();
   useEffect(() => {
-    getParsedAuthorities().then(setAuthorities);
-  }, []);
+    dispatch(getHealthcareAuthorities());
+  }, [dispatch]);
+
+  const authorities = useSelector(healthcareAuthorityOptionsSelector);
```
johnschoeman, line 33: what do you think about inlining this selector?  I find it hard as a new reader to understand, at a glance, what data is relevant here. It would be nice to not need to have multiple files open to understand whats happening.

app/views/Export/ExportTemplate.js
```diff
@@ -12,6 +12,7 @@ import { Typography } from '../../components/Typography';
 import Colors from '../../constants/colors';
 import fontFamily from '../../constants/fonts';
 import { Theme } from '../../constants/themes';
+import { TouchableOpacity } from 'react-native-gesture-handler';
```
johnschoeman: Why are we using TouchableOpacity from rn-gesture-handler rather than from react-native?

app/store/types.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// From API
+export type HealthcareAuthority = {
+  name: string;
+  bounds: Record<string, unknown>;
+  ingest_url: string;
+  publish_url: string;
+};
+
+export type AppThunk<ReturnType = void> = ThunkAction<
```
johnschoeman, line 25: I like this pattern a lot!

app/views/Export/ExportCodeInput.js
```diff
@@ -121,12 +119,18 @@ export const ExportSelectHA = ({ route, navigation }) => {
   const [codeInvalid, setCodeInvalid] = useState(false);
 
   const { selectedAuthority } = route.params;
-
   const validateCode = async () => {
     setIsCheckingCode(true);
     setCodeInvalid(false);
     try {
-      const res = await fetch(`${MOCK_ENDPOINT}?access_code=${code}`);
+      const checkAccessCodeRoute = `${selectedAuthority.ingest_url}/access-code/valid`;
```
johnschoeman, line 127: So this was the pattern that was already here, so maybe it makes sense for another commit, but what do you thing about introducing an API module for some of these?  This file is pretty gnarly and it might be nice to start separating networking logic from the view logic.

app/store/reducers/healthcareAuthoritiesReducer.ts
```diff
@@ -0,0 +1,48 @@
...
+  request: {
+    status: ApiStatus.INITIAL,
+    errorMessage: null,
+  },
+};
+
+// The 'request' property of this reducer is helpful metadata for showing loading / error cases
+// for displaying data.
+const healthcareAuthoritiesReducer = createReducer(initialState, (builder) =>
+  builder
```
johnschoeman, line 29: nice to see a reducer without a switch statement!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Dispatch } from 'redux';
+
+import { AUTHORITIES_LIST_URL_MVP1 } from '../../../constants/authorities';
+import { AppThunk } from '../../types';
+
+const GET_HEALTHCARE_AUTHORITIES_STARTED = 'GET_HEALTHCARE_AUTHORITIES_STARTED';
+const GET_HEALTHCARE_AUTHORITIES_SUCCESS = 'GET_HEALTHCARE_AUTHORITIES_SUCCESS';
+const GET_HEALTHCARE_AUTHORITIES_FAILURE = 'GET_HEALTHCARE_AUTHORITIES_FAILURE';
+
+const getHealthcareAuthorities_started = createAction(
```
johnschoeman, line 12:
```suggestion
const getHealthcareAuthoritiesStarted = createAction(
```
What do you think about using camel case here?

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -0,0 +1,55 @@
...
+  ({ error }) => ({ payload: {}, meta: { error } }),
+);
+
+const getHealthcareAuthoritiesApi = async () => {
+  const yamlString = await fetch(AUTHORITIES_LIST_URL_MVP1).then((res) =>
+    res.text(),
+  );
+  const { authorities } = Yaml.safeLoad(yamlString);
+  if (!Array.isArray(authorities)) {
+    throw new Error('authorities yaml did not return an array of authorities');
```
johnschoeman: should we really be throwing a runtime error if we get a bad response? I see that we are catching this in the thunk below, but could we instead handle the case of not getting the exact data we need explicitly in the control flow? This would allow us to have some more flexibility on how we want to respond moving forward.  

---

additionally, if we add an either/result type for a network response we can have some type safety which is consistent across all layers of the app.  something like:
```javascript
interface NetworkSuccess<T> {
  kind: "success"
  body: T
}
interface NetworkFailure<U> {
  kind: "failure"
  error: U
}

export type NetworkResponse<T, U = {}> = NetworkSuccess<T> | NetworkFailure<U>
```

so the api fetch function might look something like: 

```javascript
const getHealthcareAuthoritiesApi = async (): NetworkResponse<HealthcareAuthority[], HealtcareAuthorityError> => {
  const yamlString = await fetch(AUTHORITIES_LIST_URL_MVP1).then((res) =>
    res.text(),
  );
  const { authorities } = Yaml.safeLoad(yamlString);
  if (!Array.isArray(authorities)) {
    return {
       kind: 'failure',
       error: 'authorities yaml did not return an array of authorities'
  } else {
  return {
      kind: 'success',
      body: authorities
    }
  }
};
```

this gives us a clear spot to handle in networking specific concerns, the api function itself, which is separated form our application code which handles how we want structure our application data based off of the request result.

---

finally I would consider introducing a dedicated `healthcareAuthorityAPI.ts` module which contains all the api calls to further encourage this logical separation of the networking layer and our application business logic. (This would have the added benefit of making tests easier to add!)

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: additionally its not clear to me what exactly 'started' means here.  What do you think about something like `userRequestedHealthcareAuthorities`?

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+  return authorities;
+};
+
+const getHealthcareAuthoritiesAction = (): AppThunk<void> => async (
+  dispatch: Dispatch,
+): Promise<void> => {
+  dispatch(getHealthcareAuthorities_started());
+  try {
```
johnschoeman, line 28: instead of splitting on network success or failure in the thunk what do you thing about dispatching a `healthcareAuthoritiesResponseReceived` action and handle the logic of success/failure in the reducer?

This will allow us to keep all of the relevant application logic co-located and follows more of an eventful style. It also better separates the updating of the application state based off a dispatched action and the dispatching an action based off of some event, which better matches the intent of the flux pattern.

---

if we introduce the NetworkResponse concept from the comment above, this might look something like:

```javascript
// actions.ts
const getHealthcareAuthoritiesAction = (): AppThunk<void> => async (
  dispatch: Dispatch,
): Promise<void> => {
  dispatch(getHealthcareAuthorities_started());
  const response = await getHealthcareAuthoritiesApi();
  dispatch(getHealthcareAuthoritiesResponseReceived({ response }));
};
```

```javascript
// reducer.ts

...
.addCase(
      getHealthcareAuthoritiesResposneRecieved,
      (state, { payload: { response } }) => {
          if (response.kind === "success") {
             return {
                state.request.status = ApiStatus.SUCCESS;
                state.request.errorMessage = null;
               state.availableAuthorities = healthcareAuthorities;
             }
          } else {
              return {
                 state.request.status = ApiStatus.FAILURE
                 state.request.errorMessage = response.error
          }
      }
)
```

we can probably find a better way to factor some of this, but this has the benefit of all the complexity and control flow being in one place.

app/store/types.ts
```diff
@@ -0,0 +1,36 @@
+import { store } from './index';
```
johnschoeman, line 1: Im less of a fan of dedicated `types.ts` files. I find that organizing by domain concept and having the types inline make it a bit easier to reason about the code.  What do you think about removing this file and inlining the types where they are used?

johnschoeman (approved):
Great work here, it is very clean and easy to follow!

I like the style introduced by @reduxjs/toolkit and typesafe-actions, its nice to see a lot of the boiler plate removed.  We are introducing a handful of dependencies though. How confident are you that these are well maintained? will there be any significant maintenance burden introduced by relying on these tools?

One thing that jumps out at me is that we might want to more distinctly separate the networking layer from the state management layer.  I see them as a bit conflated here and by decoupling them we would have some maintainability wins.  They ended up being a bit long, but I have a few comments thinking through what this might look like, let me know your thoughts!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: started the review before you merged some updates yesterday, so this maybe isnt relevant to this pr anymore, but still interested in your thoughts here!

JacobJaffe: IMO, the major con is that is requires having a dedicated type files to facilitate the export (which I set up here to allow for the explicit export type). 
It's kinda a pain on HOF exports like the thunks, but now that its setup it should be straightforward going forward.

app/views/Export/ExportSelectHA.js
JacobJaffe: I normally advocate for inline selectors when not using reselect, but with the `ts` introduction I'm a fan of this selector because it adds more safety across the views in `js` and redux in `ts` separation. The ts validation that we're selecting real & accessible redux state is a big win for stability I think.

(If this component were in `ts` I think inline would be better -- I agree this makes it less readable)

app/views/Export/ExportTemplate.js
JacobJaffe: Auto imported -- good catch. Will change

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
JacobJaffe: I'm not too opinionated about this so I'm down for changing it, but here's my rationale for this unconventional typing:

```
getHealthcareAuthoritiesStarted
getHealthcareAuthoritiesSuccess
getHealthcareAuthoritiesFailure
```
Is a lot of cognitive load to distinguish between the different actions, vs:

```
getHealthcareAuthorities_started
getHealthcareAuthorities_success
getHealthcareAuthorities_failure
```
 To me it's easier for a case like this where there are three variables intentionally similar to use `_` to denote where the distinction is. Kinda like using `_property` convention on a private property.

app/store/types.ts
JacobJaffe: See comment about the export modules. Having a central source becomes more necessary when we're explicitly defining exports & avoiding circular dependencies

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
JacobJaffe: imo its best to minimize paths occurring in reducers. Switching reducers on action types is pretty robust, when another switch is introduced (`response.kind` here), the resolution tree of the reducer becomes increasingly complex with each added condition to decipher the action.

As I understand that proposal, that would be saying 
`getHealthcareAuthoritiesResposneRecieved` with `response.kind === "success` would be equivalent to the `getHealthcareAuthorities_success` action. 

I prefer keeping actions as declarative as possible to their payloads and intent. This allows them to be agnostic to their reducers also -- If another reducer uses `getHealthcareAuthorities_success`, it can easily use that action, vs. _also_ needing to know how to infer that from a combination of action type & payload structure. 

This makes debugging / developing / middlewares around actions easier also -- when figuring out what is happening in the app, you just need to read the action stream, which is declarative of intent, vs. also knowing how the actions are meant to be interpreted 

Also, if it were to be inferred, I think that should occur in the `meta` field, so the action payloads are just the incoming data to the store state.

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
JacobJaffe: Totally agreeing on having a dedicated API folder for the app. Typically I like segmenting 
```
app/
  api/
  store/
  components/
  ...
```
I'm not a fan of how I set up the APIs contracts are being determined in the components right now, I'll make an API folder to improve that. 

I think creating an interface for error types is a good idea, but I disagree on extending that to encapsulate all responses -- I think the cleanest way for an async api abstraction to occur is for it to have the form of:

(very rough pseudo:)
```js
const exampleApi = async (): Foo => {
   try {
     const foo = await request(...); // Can throw, e.g. on 400
     if (!validateFoo(foo)) { // can throw on payload not matching expected form
       throw ValidationError 
     }
   catch (e) {
     throw createTypeSafeError(e);
}
```

Essentially, following the Axios model of either APIs resolve successfully or throw with an error. I like this because it makes chaining error handling APIs much easier (Like in the try in the thunk). When errors resolve into successful response bodies, though, I find it harder to have consumers implement error chaining.

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: > IMO, the major con is that is requires having a dedicated type files to facilitate the export.

Im not sure this is true? my understanding of the rule is that the any export in any ts file needs to be explicitly typed. The file doesnt need to have only types declared in it.

ref:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md

app/views/Export/ExportSelectHA.js
johnschoeman: Makes sense to me!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: Ah i see the reasoning now.  I too dont have a strong opinion, so if you prefer the getThing_started style i say go for it.

I guess im used to seeing actions named like:
```
userRequestedData
dataResponseReceived
```
so there's less duplication in the action names, but yeah if every network request had the three similarly named actions, I see the benefit of the unconventional casing.  (also what would you call this, camelSnake_case?)

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: > I think creating an interface for error types is a good idea, but I disagree on extending that to encapsulate all responses -- I think the cleanest way for an async api abstraction to occur is for it to have the form of:

To be clear, I wasnt suggesting creating an error type that covers **all** responses.  usually I see an error type for each response, or even no error type as the generic defaults to {}, but, imo, it depends on the api and what the intent of the code is to determine on what makes sense.

I do think it makes sense to have a generic type for network responses in general though.

> When errors resolve into successful response bodies, though, I find it harder to have consumers implement error chaining.

Hmm, I am confused on what you mean by errors resolving into successful response bodies here. do you mean network error? or the type of error when you get a successful response but the data contains an 'error' field? or the type of error where the data was in a form that wasn't expected? And how are they resolving into successes?

---

Oops, I see the source of the confusion now. totally my bad for being sloppy in the comment above. What i meant to say was that it strikes me as odd to be throwing errors for the data validations and that it would be nice to have a NetworkResponse type.

what i meant to write in the code snippet was something like:

```typescript
type HealthcareAuthorityError = 'network_error' | 'no_authorities_error'

const getHealthcareAuthoritiesApi = async (): Promise<NetworkResponse<HealthcareAuthority[], HealthcareAuthorityError>> => {
  try {
  const response = await fetch(url).then((res) =>
    res.text(),
  );
  const { authorities } = Yaml.safeLoad(response);
  if (!Array.isArray(authorities)) {
    return {
       kind: 'failure',
       error: 'no_authorities_error'
  } else {
  return {
      kind: 'success',
      body: authorities
    }
  }
  } catch (e) {
  return { kind: 'failure', error: 'network_error'}
}
};
```

I forgot the try/catch and the Promise in the type signature.

so you wouldnt be able to chain on with a .catch(), but you could still .then(), I guess I dont see validation errors as exceptions, so handling them in the normal control flow seems correct to me.

Either way, this thread is getting a bit long, and your implementation totally works; so if you prefer that, you should totally go for it!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
johnschoeman: you make a lot of good points! I guess i privilege having actions match as closely as possible to real world events, which in this case is that a network response was received. I see how that can be a bit more work to deal with in the dev tools though.


> when another switch is introduced (response.kind here), the resolution tree of the reducer becomes increasingly complex with each added condition to decipher the action.

true, but you still have the same amount of conditionals in the system, you just have it split across multiple files now. which is harder to reason about, imo.

Anyways i think we are getting into personal preference territory here. Definitely go with what you prefer!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
JacobJaffe: sorry, unclear explanation on my part -- 
I mean to avoid circular dependencies / keeping the types in a sensible place.

E.g., for something like `type HealthcareAuthority`,there are three locations that it would make sense to have the type definition: 
1. the API file which receives the payload from the API, and returns the `HealthcareAuthority` type
2. the action layer, which deals with ensuring correct types are sent to the store
3. the reducer, which deals with ensuring the store has a consistent type. (Declare the type alongside store shape)

(1) Is how I typically structure imports without types: API -> Action -> Reducer inheritance, because then the API is a leaf. However, in a type system then components who have methods on the type are importing from the API (which imo is ideally abstracted away if they interface via the actions). E.g. 
```js
// This component file is aware of both the API layer and the Redux layer
import { FooSelector } from "selectors";
import { FooType } from "API";

const localFooTransform = (foo: FooType): FooType => { ... };

const ExampleComponent = () => {
  const foo = select(FooSelector);
};
```
That said, my solution of the API taking the types from `../store/types` isn't much better, as it still becomes aware of redux. 

(2) & (3) bring circular dependencies (which aren't a huge deal, but imo reflects a confusing abstraction)

Without a root types file, where would you recommend defining types? I don't work with typescript much so I'm curious if there's a better pattern for share types between abstraction layers

johnschoeman: I think the first design consideration is whether or not the API is what is dictating the shape of the types.  If you dont control the api, or its a bit hairy, i usually see a distinction between the type that the api returns and the type that is used throughout the application.  and a function to convert between the two.

so you might have something like: 

```
/api
  /networkModel
     healthcareAuthority.ts
     foobar.ts
     index.ts
  /networkConverters
     healthcareAuthority.ts
     fooboar.ts
     index.ts
   /dataModel
     healthcareAuthority.ts
     foobar
     index.ts
api.ts
```

and you can use some module importing to access them pretty conveniently

```typescript
//api.ts
import { HealthcareAuthority } from './networkModel'

const fetchHealthAuthorities: = async (): Promise<NetworkResponse<HealthcareAuthority[]>> => {
   const reponse = fetch(url, ...
    const networkData = handle the response
   return networkData
...
```

```typscript
// reducer.ts or action.ts

import { HealthAuthority } from "../api/dataModel"
import { toHealthAuthority } from "../api/networkConverter"

const getHealthcareAuthorities_success = createAction(
  GET_HEALTHCARE_AUTHORITIES_SUCCESS,
  ({ data }) => ({ payload: { healthcareAuthorities: data.map(toHealthcareAuthority) } }),
);

```

so you can still pull out the types to a dedicated file, but its just grouped by feature/domain concept, rather than the fact it is a type.


And you if you are basing the type off of the api, you could simply have the dataModel folder or just have the type files at the same level as everything else.

JacobJaffe: Good points on the data model differences between the layers! that makes a lot of sense, thanks

PR #966, opened 2020-06-06
SAF-390: Almost all the issues are fixed in android

nextgenevent:
Android GPS Accuracy and reliability issues have been fixed.

### DIFF
--- a/scripts/replace.js
+++ b/scripts/replace.js
@@ -47,7 +47,7 @@ copyFile(
 
 copyFile(
   './scripts/node-modules/react-native-background-geolocation/BackgroundGeolocationFacade.java',
-  './node_modules/@mauron85/react-native-background-geolocation/android/common/src/main/java/com/tenforwardconsulting/bgloc/BackgroundGeolocationFacade.java',
+  './node_modules/@mauron85/react-native-background-geolocation/android/common/src/main/java/com/marianhello/bgloc/BackgroundGeolocationFacade.java',
   () => {},
 );
 copyFile(
--- a/scripts/replace.js
+++ b/scripts/replace.js
@@ -1,0 +1,57 @@
+const fs = require('fs');
+
+function copyFile(source, target, cb) {
+  let cbCalled = false;
+
+  const rd = fs.createReadStream(source);
+  rd.on('error', err => {
+    done(err);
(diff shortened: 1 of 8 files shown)


scripts/replace.js
```diff
@@ -0,0 +1,57 @@
...
+
+  function done(err) {
+    if (!cbCalled) {
+      cb(err);
+      cbCalled = true;
+    }
+  }
+}
+
+copyFile(
```
johnschoeman, line 27: This looks unfortunate and pretty brittle.  What led to needing to copy all these files?

johnschoeman (commented):
Overall looks good to me! but I dont have much context on the intent here. Could you provide a bit of explanation on what this commit is doing?

scripts/replace.js
nextgenevent: Hello @johnschoeman, 

This is for your information that we were working on the ticket SAF-390. Here we were fixing GPS accuracy and reliability issues. We had to change in some java files of lib react-native-background-geolocation (that we are currently using). You see I changed in a node-module lib that we don't push. So somehow I need to keep the changes what I did in the lib.

For the same I created a script that will run after yarn or npm install to copy some files (where we changed) with existing node-modules lib react-native-background-geolocation in the machine from where we are creating the builds.

Please let me know if you have some other ways to get our changes in the node-modules lib. 

However Diarmid has tested these changes and given good feedback regarding fixes of GPS accuracy and reliability in android.

johnschoeman: got it, so we needed to overwrite a bunch of the geolocations libraries files because the base library is broken for our needs?  is it possible to fork the library instead or is there a patch available?

If we do need to copy n paste node module files in apost script, your script here is fine, its just unfortunate that we have to do it in the first place.

nextgenevent: Hello @johnschoeman, yes we needed to overwrite some files of the geolocation lib. Yes we can fork it. But for the first MVP, could we continue with the postscript and later we will fork it?

johnschoeman (approved):
Doesnt look like anyone with familiarity with android is able to review this, so I dont have much to offer on the android side, but for sake of not having this open for weeks lets ship it!

PR #969, opened 2020-06-08
Exposure Notifications Debug Menu iOS

mattThousand:
## Why:
We would like to add a debug menu which allows developers running the app to interact with the native `Exposure Notifications` API through the app UI. This will enable them to recreate specific app states while debugging without modifying the codebase.

## This commit:
Adds components for the debug menu and also a detail page where the full list of `diagnosis keys` can be viewed. It also adds a native module in the iOS codebase to send commands to the `Exposure Notifications` API.

IMG_1819
IMG_1820


app/views/Settings/ENDebugMenu.js
```diff
@@ -0,0 +1,178 @@
...
+} from '../../exposureNotificationsNativeModule';
+
+export const EN_DEBUG_MENU_SCREEN_NAME = 'ENDebugMenu';
+export const EN_LOCAL_DIAGNOSIS_KEYS_SCREEN_NAME = 'ENLocalDiagnosisKeyScreen';
+
+export const ENDebugMenu = ({ navigation }) => {
+  useEffect(() => {
+    const handleBackPress = () => {
+      navigation.goBack();
+      return true;
```
johnschoeman: I think it better follows convention to return void here.

app/views/Settings/ENDebugMenu.js
```diff
@@ -0,0 +1,178 @@
...
+    );
+  };
+  return (
+    <NavigationBarWrapper title={'EN Debug Menu'} onBackPress={backToSettings}>
+      <ScrollView>
+        <Section>
+          <Item
+            label='Detect Exposures Now'
+            onPress={() => {
+              const cb = (errorString) => {
```
johnschoeman: theres a lot going on here, I might consider extracting a function something like: `handleOnPressDetectExposures`

you could even be fancy and do something like:
```typescript 

const handleOnPressSimulationButton(callSimulatedEvent, successMessage) => {
   return () => {
       const cb = (errorString) => {
                if (errorString) {
                  showErrorAlert(errorString);
                } else {
                  showSuccessAlert(successMessage)
                }
              };
              callSimulatedEvent(cb);
   }
}


... onPress={handleOnPressSimulationButton(detectExposures, "Exposure detection successful")}
```

app/views/Settings/Section.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
johnschoeman, line 1: This file looks a little complicated for what it does. You could likely inline a lot of this and handle the layout purely with styles.

app/views/Settings/ENDebugMenu.js
```diff
@@ -0,0 +1,178 @@
+import React, { useEffect } from 'react';
```
johnschoeman, line 1: what do you think about converting this to a ts file?

johnschoeman (approved):
Looking good so far!

app/ExposureNotificationContext.tsx
```diff
@@ -29,18 +37,32 @@ const ExposureNotificationsProvider = ({
...
+
   const requestExposureNotificationAuthorization = () => {
     const cb = (authorizationStatus: ENAuthorizationStatus) => {
       setExposureNotificationAuthorizationStatus(authorizationStatus);
     };
     ExposureNotifications.requestAuthorization(cb);
   };
 
+  const fetchDiagnosisKeys = () => {
+    const cb = (error: Error, diagnosisKeys: ENDiagnosisKey[]) => {
```
mattThousand: @johnschoeman is there a best practice for handling errors within a context? I'd like to be able to display an alert to the user, so somehow I'd like to pass the error to the caller I think. Related question: will this context have the same role now that `Redux` is gradually being worked in?

PR #973, opened 2020-06-09
iOS e2e se testing

nextgenevent:
testing


yonatanmk:
This change will be unnecessary after the new onboarding designs are implemented in https://github.com/Path-Check/covid-safe-paths/pull/963

yonatanmk (changes requested):
This change will be unnecessary after the new onboarding designs are implemented in #963

nextgenevent:
> This change will be unnecessary after the new onboarding designs are implemented in #963

Ok so e2e-iphone-se test will not run at the moment in any PR that are created with target branch develop, right?

PR #974, opened 2020-06-09
Update yarn run-ios-bte script ios target

johnschoeman:
Why:
Currently the yarn run-ios-bte script does not successfully build and
run the bte version of the app. The reason for this is because a recent
commit changed the bte ios target to 13.5, as it is required to use the
exposure notifications api, but the react-native cli does not correctly
find the ios 13.5 versions of the simulator to build to.

This commit:
Explicitly specifies iOS 13.5 in the build script.


tparesi (approved):
Awesome - thanks. I recognized this when xcode was using 13.5 and functioning properly.

JacobJaffe:
I ran this, and it looks like the simulator instruction was ignored now:
terminal

mattThousand (approved):

johnschoeman:
> I ran this, and it looks like the simulator instruction was ignored now:
> terminal

@JacobJaffe do you have the ios 13.5 simulators installed?

PR #975, opened 2020-06-09
Run validate script in ci actions

johnschoeman:
Why:
Currently ci is only running eslint as a static analysis tool. We would
like to be running prettier and tsc as well during ci.

This commit:
Changes the lint action to a validate action which will run prettier,
eslint, and tsc.


app/api/export/exportConsentApi.ts
```diff
@@ -1,4 +1,4 @@
-import type { HealthcareAuthority } from '../HealthcareAuthorities/getHealthcareAuthoritiesApi';
```
johnschoeman, line 1: These were failing the tsc check!

tstirrat (approved):

PR #976, opened 2020-06-09
Update logkitty dependency

johnschoeman:
Why:
Currently we have a security vulnerability in the logkitty dependency.

This commit:
Updates to the latest version. Not that the isnt a direct dependency,
but an indirect dependency of react-native itself. A better solution
would be to update react-native to 0.62 but we are just logkitty for
sake of time. A future commit should remove logkitty as a direct
dependency.


tparesi (approved):
LGTM

PR #977, opened 2020-06-09
Fix type declaration for @emotion/native

johnschoeman:
Why:
Currently there is a typo causing the tsc to fail compilation.


mattThousand:
@johnschoeman thanks for doing this!

mattThousand (approved):

PR #979, opened 2020-06-09
Convert Exposure Notifications Debug Screen to Typescript

mattThousand:
## Why
We'd like to convert the Exposure Notifications Debug menu screen to typescript to ensure better type safety and reduce the likelihood of runtime errors.

## This commit
Converts components related to the Exposure Notifications debug menu screen to typescript


app/components/Theme.ts
```diff
@@ -0,0 +1,11 @@
+export interface ThemeProps {
```
mattThousand: @johnschoeman this almost certainly is an odd place for these types to live. Let me know if you have a recommendation as to where they might go

johnschoeman: Personally, id keep this one inline. by principle of least surprise, i would expect these to be colocated with the component that uses them.

as an aside, geekier note:
I think theres a tendency to have smaller files in js land to provide some amount of boundaries on what is possible which is less necessary in ts land.
this is an elm talk, but i think it applies to types in frontend code in general: https://www.youtube.com/watch?v=XpDsk374LDE. (not relevant for this pr, but watch it if your interested!)

app/exposureNotificationsNativeModule.ts
```diff
@@ -19,55 +19,55 @@ export const fetchDiagnosisKeys = async (
 };
 
 export const detectExposuresNow = async (
-  cb: (error: Error) => void,
+  cb: (errorString: string) => void,
```
johnschoeman:
```suggestion
export type ENModuleError = string | null

export const detectExposuresNow = async (
  cb: (error: ENModuleError) => void,
```

my understanding is that this can be either a string or null, so we should probably capture that in the type system.  Additionally its best practice to only use the primitives if the actual intent is that primitive and provide a type alias if the variable represents a useful domain concept.  for one, it is more readable to those without context, and for two it is much easier to update if the ENModuleError ever needs to change in the future.

app/views/Settings/ENDebugMenu.tsx
```diff
@@ -15,14 +15,24 @@ import {
...
   getExposureConfiguration,
 } from '../../exposureNotificationsNativeModule';
+import {
+  NavigationParams,
+  NavigationScreenProp,
+  NavigationState,
+} from 'react-navigation';
+
+type ENDebugMenuProps = {
+  navigation: NavigationScreenProp<NavigationState, NavigationParams>;
```
johnschoeman, line 25: 💯 I admire your lack of laziness!

app/views/Settings/ENDebugMenu.tsx
```diff
@@ -36,11 +46,13 @@ export const ENDebugMenu = ({ navigation }) => {
     navigation.goBack();
   };
 
-  const showErrorAlert = (errorString) => {
-    Alert.alert('Error', errorString, [{ text: 'OK' }], { cancelable: false });
+  const showErrorAlert = (errorString: string) => {
```
johnschoeman, line 49: similar comment to the ENModule error from above.

johnschoeman (approved):
LGTM, a few small comments!

app/components/Theme.ts
johnschoeman: made a comment, but this a bit of a weird case as this really should be defined in the library.

I would duplicate the type in this case and define them in each component that uses it and not export it.  the reasoning being is this isnt an application concept to our app but a type we are putting in to over a third party library, we want to keep this at the boundary of our system and not encourage its use as we dont actually have control over that library.

another option would be to put in the declarations file.

app/exposureNotificationsNativeModule.ts
```diff
@@ -19,55 +19,58 @@ export const fetchDiagnosisKeys = async (
...
+
+export const detectExposuresNow = async (
+  cb: (error: ENModuleError) => void,
 ): Promise<void> => {
   debugModule.detectExposuresNow(cb);
 };
 
 export const simulateExposure = async (
-  cb: (error: Error) => void,
+  cb: (errorString: string) => void,
```
johnschoeman: to be clear, i would use `ENModuleError` for every callback,

PR #980, opened 2020-06-09
Fix bug in navigator stack for bt onboarding

johnschoeman:
Why:
Currently the EnableExposureNotificationsScreen is included in the wrong
stack.


tparesi (approved):
Thanks for fixing this

mattThousand (approved):

PR #982, opened 2020-06-10
Only run location background service in gps build

johnschoeman:
Why:
Currently both builds of the app are initiating the geolocation
background service. We would like to only run this on the gps version of
the app.


tparesi (approved):
LGTM

PR #983, opened 2020-06-10
Introduce common styles

johnschoeman:
Why:
Currently the application has styles hard coded throughout the app.
There are many instances of inconsistencies throughout the styling
implementation. We would like to introduce a styles module in which
components can derive common styles from, as this will allow us to
ensure better consistency throughout the app and make it easy to update
styles as designs change over time.

This commit:
Introduces an app/styles folder which contains a handful of styles
modules. Much of the included styles are not yet used but are left in to
show some utility function based style constant organization patterns.
Future commits should start to incorporate these styles into the app
components and update them to fit the design requirements.

An arbitrarily selected component was updated to use the style
modules as an example in usage.

### DIFF
--- a/app/styles/layout.ts
+++ b/app/styles/layout.ts
@@ -1,0 +1,25 @@
+import { Dimensions } from 'react-native';
+
+import * as Spacing from './spacing';
+
+export const screenWidth = Dimensions.get('window').width;
+export const screenHeight = Dimensions.get('window').height;
+
+export const baseScreenPadding = Spacing.xSmall;
+
+export const xxSmallWidth = 0.05 * screenWidth;
+export const mediumWidth = 0.4 * screenWidth;
+export const largeWidth = 0.6 * screenWidth;
+
+export const xxSmallHeight = 0.03 * screenHeight;
+export const smallHeight = 0.05 * screenHeight;
+export const mediumHeight = 0.45 * screenHeight;
+
+export const baseHorizontalMargin = Spacing.xSmall;
+export const horizontalMarginSmall = Spacing.xxxSmall;
+export const horizontalMarginMedium = Spacing.medium;
+
+export const baseFullWidthWithMargin = screenWidth - baseHorizontalMargin * 2;
+export const fullWidthWithSmallMargin = screenWidth - horizontalMarginSmall * 2;
+export const fullWidthWithMediumMargin =
+  screenWidth - horizontalMarginMedium * 2;
--- a/app/styles/outlines.ts
+++ b/app/styles/outlines.ts
@@ -1,0 +1,60 @@
+import { ViewStyle } from 'react-native';
+
+import * as Colors from './colors';
+
+export const baseBorderRadius = 4;
+export const borderRadiusLarge = 8;
+export const borderRadiusMax = 500;
+
+export const baseShadow: ViewStyle = {
+  shadowColor: Colors.black,
+  shadowOpacity: 0.2,
+  shadowRadius: 4,
+  elevation: 7,
+};
+
+export const buttonShadow: ViewStyle = {
+  ...baseShadow,
+  shadowOffset: {
+    width: 0,
+    height: 3,
+  },
+};
+
+export const upShadow: ViewStyle = {
+  ...baseShadow,
+  shadowOffset: {
+    width: 0,
+    height: -3,
+  },
+};
+
+export const leftShadow: ViewStyle = {
+  ...baseShadow,
+  shadowOffset: {
+    width: -3,
+    height: -3,
+  },
+};
+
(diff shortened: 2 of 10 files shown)


johnschoeman:
Some of the reasoning for introducing the styles in this way in particular can be found here: https://thoughtbot.com/blog/structure-for-styling-in-react-native

tparesi (approved):
Awesome

app/styles/layout.ts
```diff
@@ -0,0 +1,25 @@
+import { Dimensions } from 'react-native';
+
+import * as Spacing from './spacing';
+
+export const screenWidth = Dimensions.get('window').width;
```
JacobJaffe, line 5: I think we should move away from dimension percentage based styles, and stick just with flex.

Percentage based layouts force screens to be aware of their navigation context (e.g. a screen with a percentage height needs to know it is wrapped in a tab navigator for a bottom offset). IMO this mixing should be avoided unless explicitly needed, so I'm wary of it being encouraged by standardization. 

I think percentage based is fine for introducing paddings & spacings, but since there is also a `spacing` file, I don't think this is necessary.

app/styles/outlines.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+export const bottomHRule: ViewStyle = {
+  ...baseHRule,
+  borderBottomWidth: 1,
+};
+
+export const topHRule: ViewStyle = {
+  ...baseHRule,
+  borderTopWidth: 1,
```
JacobJaffe, line 59: For an outline rule, would `StyleSheet.hairlineWidth` make more sense?

app/styles/typography.ts
```diff
@@ -0,0 +1,101 @@
...
+export const borderlessButtonText: TextStyle = {
+  color: Colors.defaultBlue,
+  fontSize: medium,
+  fontWeight: heaviestWeight,
+};
+
+export const selectableButton: TextStyle = {
+  ...smallFont,
+  fontWeight: heaviestWeight,
+  lineHeight: 0,
```
JacobJaffe: What's with this lineheight?

app/styles/typography.ts
```diff
@@ -0,0 +1,101 @@
+import { TextStyle } from 'react-native';
+
```
JacobJaffe, line 2: What's the internal / external abstraction occurring in this file?
I'm all for having a standardized typography styling, but I think that's best by declaring our definitions of fonts (be it in a css Heading1, Body2 etc. system, or some other way), linking them with our design system, and ensuring parity. (rather than defining them in design, and having to construct them from here).

Maybe we should couple this strongly with the `Typograpy` component? Maybe we should remove the typography component, and just use declared styles? Introducing this as is will have two different styled text systems in the app.

JacobJaffe (commented):
This is awesome, thanks for putting this together. A few suggestions on some nits. 

Biggest comment here is on the `typography` setup. I'm wary of adding another text styling system without ensuring we can achieve design parity & consistency -- that is, ensuring that whatever system we set up for our typography is consistently in sync with design.

johnschoeman:
> Biggest comment here is on the `typography` setup. I'm wary of adding another text styling system without ensuring we can achieve design parity & consistency -- that is, ensuring that whatever system we set up for our typography is consistently in sync with design.

100% agreed with this concern.  I see the typography component as a component, which handles view logic, and the typography module as a place for common styles.  The naming is unfortunate but we can clean up and make consistent as we go.

I took a pass at adding these styles to the typography component, but it was taking a while so i just went with this.  The intent is to centralize all of the design knowledge.

And to be clear, none of these styles are set in stone, I really just want to get this process of not having hard coded values everywhere started.

app/styles/layout.ts
johnschoeman: > I think we should move away from dimension percentage based styles, and stick just with flex.

how do you define spacing orthogonal to the flex direction then?  But I do agree, flex should be the default, and fixed width based of the screen size for when flex doesnt work.

app/styles/typography.ts
johnschoeman: > I'm all for having a standardized typography styling, but I think that's best by declaring our definitions of fonts (be it in a css Heading1, Body2 etc. system, or some other way), linking them with our design system, and ensuring parity. (rather than defining them in design, and having to construct them from here).

People go bike shed *a lot* on this. While im a fan of utility classes, I'd personally defer to the designers on their preference.


> Maybe we should couple this strongly with the `Typograpy` component? Maybe we should remove the typography component, and just use declared styles? Introducing this as is will have two different styled text systems in the app.

agreed.

jeffstolz (commented):
Awesome - thanks for doing this

PR #985, opened 2020-06-10
Introduce tab navigation

johnschoeman:
Why:
We would like for users to be able to navigate the app via a bottom tab
bar.

This commit:
Switches the current main stack navigation with a tab naviation.

### GPS After

gps-after

### BTE After

bte-after
bte-after-8


JacobJaffe (commented):
Ship it, I'll follow up on fixing the back behaviors seperately

JacobJaffe (approved):
oops I didn't approve

PR #987, opened 2020-06-10
BTE iOS APIClient/Realm Implementation

mattThousand:
### Why
We would like to communicate with the GAEN server according to its api, and only store the bare minimum information necessary on device in order to preserve user privacy.

### This Commit
This commit modifies the `iOS` `APIClient` to request the collection of urls corresponding to the exposure keys it must download, then using those urls download the keys from the server as well as the server's exposure configuration, for use determining if the user has been exposed.

This commit also stores exposures and test results in `Realm`, and updates the debug menu in the Javascript layer accordingly.

IMG_1823

IMG_1822


johnschoeman (approved):
🥇 :shipit:

ios/BTE/ExposureManager.swift
```diff
@@ -136,60 +201,58 @@ final class ExposureManager: NSObject {
...
+          completion([NSNull(), "Exposure detection successfully executed."])
         } else {
           completion(["Exposure detection error.", NSNull()])
         }
       }
     case .simulateExposureDetectionError:
-      LocalStore.shared.exposureDetectionErrorLocalizedDescription = "Unable to connect to server."
-      completion([NSNull(), LocalStore.shared.exposureDetectionErrorLocalizedDescription])
+      BTESecureStorage.shared.exposureDetectionErrorLocalizedDescription = "Unable to connect to server."
+      completion([NSNull(), "Exposure deteaction error message: \(BTESecureStorage.shared.exposureDetectionErrorLocalizedDescription)"])
```
schmecs, line 211: small typo in "detection"

schmecs (approved):
👍  This is very helpful for next steps on configuring the Android side

PR #988, opened 2020-06-10
Convert helpers/general and storage.js to ts

johnschoeman:
Why:
We would like to use typescript on some files which import theses
functions and constants.

This commit:
Updates the files to ts.  Some rather permissive types are being
returned from the functions, which could likely be tightened up in a
future commit.


johnschoeman:
trying to ship smaller commits

app/helpers/General.ts
```diff
@@ -1,34 +1,30 @@
...
-    let data = await AsyncStorage.getItem(key);
+    const data = await AsyncStorage.getItem(key);
 
     if (isString) {
       return data;
     }
 
-    return JSON.parse(data);
+    if (data) {
+      return JSON.parse(data);
```
JacobJaffe, line 14: Doesn't this return an any type? 
Without explicit setters / getters per payload to the store I think this whole `GetStoreData` needs to be typed as an `any` return

johnschoeman: you are correct! good catch

app/TracingStrategyAssets.ts
```diff
@@ -0,0 +1,103 @@
+import { useTranslation } from 'react-i18next';
```
JacobJaffe: It looks like to me this file got added without removing the corresponding .js version?

JacobJaffe (changes requested):

JacobJaffe (approved):

tparesi (approved):
LGTM

PR #989, opened 2020-06-10
Tab Bar Fixes

JacobJaffe:
Follow ups to #984 to fix some styling and navigation issues:

- remove the safe area on the screen wrappers (No longer needed because of tab navigation). This fixes the screen mount jank issue. 
- align styling of buttons between Providers & Export so switching between the two has a consistent experieence.
- Puts the export at a modal level, so it's hoisted above the tab bar. 

image
image
image


johnschoeman (approved):
💯 🚢

PR #990, opened 2020-06-10
Convert Legal screen to typescript.

mattThousand:
### Why:
We would like to use typescript in general for better developer ergonomics

### This commit:
Updates files related to the `Legal` screen to `ts`

Simulator Screen Shot - iPhone 11 - 2020-06-10 at 18 43 19

### DIFF
--- a/app/TracingStrategyAssets.ts
+++ b/app/TracingStrategyAssets.ts
@@ -3,7 +3,28 @@ import { useTranslation } from 'react-i18next';
 import { Images } from './assets';
 import { isGPS } from './COVIDSafePathsConfig';
 
-export function useAssets() {
+export type TracingStrategyAssetMap = {
+  onboarding2Background: string;
+  onboarding2Header: string;
+  onboarding2Subheader: string;
+  onboarding3Background: string;
+  onboarding3Header: string;
+  onboarding3Subheader: string;
+  onboarding4Background: string;
+  onboarding4Header: string;
+  onboarding4Subheader: string;
+  onboarding4Button: string;
+  settingsLoggingActive: string;
+  settingsLoggingInactive: string;
+  aboutHeader: string;
+  legalHeader: string;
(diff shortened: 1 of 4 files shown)


app/TracingStrategyAssets.ts
```diff
@@ -3,7 +3,28 @@ import { useTranslation } from 'react-i18next';
 import { Images } from './assets';
 import { isGPS } from './COVIDSafePathsConfig';
 
-export function useAssets() {
+export type TracingStrategyAssetMap = {
+  onboarding2Background: string;
```
mattThousand: @johnschoeman not sure if this is remotely idiomatic; `onboarding2Background` for instance is an image - is it encoded as a string or is a different type needed here?

app/components/NavigationBarWrapper.tsx
```diff
@@ -20,10 +20,10 @@ import Colors from '../constants/colors';
 const widthScale = Math.min(Dimensions.get('window').width / 400, 1.0);
 
 interface NavigationBarWrapperProps {
-  children: JSX.Element;
+  children: React.ReactNode;
```
johnschoeman, line 23: why did this need to change?

app/TracingStrategyAssets.ts
johnschoeman: Yeah, noticed this too.  I think what you have here is fine.  but i think the ideal solution is to not have mixed value types, its a bit odd to have the strings and images mixed as such. so less of a typescript issue and more of the api could be changed to separate the text and the images.

a note, you could have done something like:
```
export function useAssets(): Record<string, string>
```

and ts should infer the correct names for you on the object

or even something like

```
export function useAssets(): Record<string, string | Image>
```
where `Image` is whatever the image is supposed to be.

johnschoeman (approved):

PR #991, opened 2020-06-11
Introduce exposure detail for demo

johnschoeman:
Why:
We would like to display exposure info to the user on the exposure
history screen for an upcoming demo.

This commit:
Adds an exposure detail component and wires up a way to toggle the
exposed and not exposed state in the bt app. We used fake data and
did not provide any translations for sake of time. This will need to be
updated in a future commit.

We also removed the Theme component as this is not a feature we will be
supporting moving forward.

### gif

has-exposure

### DIFF
--- a/app/views/ExposureHistory/DetailedHistory.js
+++ b/app/views/ExposureHistory/DetailedHistory.js
@@ -1,11 +1,16 @@
-import styled from '@emotion/native';
-import React from 'react';
+import React, { useContext } from 'react';
+import { View, StyleSheet } from 'react-native';
 
 import { Typography } from '../../components';
 import languages from '../../locales/languages';
 import { useAssets } from '../../TracingStrategyAssets';
 import { ExposureCalendarView } from './ExposureCalendarView';
 import { SingleExposureDetail } from './SingleExposureDetail';
+import { isGPS } from '../../TracingStrategyAssets';
+import ExposureNotificationContext from '../../ExposureNotificationContext';
+import HasBeenExposedDetail from './HasBeenExposedDetail';
+
+import { Spacing } from '../../styles';
 
 /**
  * Detailed info when there is some exposure found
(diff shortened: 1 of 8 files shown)


app/views/ExposureHistory/DetailedHistory.js
```diff
@@ -14,60 +19,94 @@ import { SingleExposureDetail } from './SingleExposureDetail';
...
 
-  return (
-    <>
-      <ExposureCalendarView weeks={3} history={history} />
+  const Divider = () => {
+    return <View style={styles.divider} />;
+  };
 
-      <Divider />
+  const gpsExposureInfo = () => {
```
johnschoeman, line 29: the existing view layout is tightly couple to gps logic, instead of untangling it all, i moved it over here for sake of time. a future pass should consolidate this. none of the functionality was changed, so for gps this is the same as it was before.

app/views/ExposureHistory/DetailedHistory.js
```diff
@@ -14,60 +19,94 @@ import { SingleExposureDetail } from './SingleExposureDetail';
...
-        <>
-          <Typography use='headline3'>
-            {languages.t('history.what_if_no_symptoms')}
-          </Typography>
-          <Typography use='body3'>
-            {languages.t('history.what_if_no_symptoms_para')}
-          </Typography>
-        </>
-      ) : null}
+  const bteExposureInfo = (hasBeenExposed) => {
```
johnschoeman, line 77: This is the new stuff added in this pr.

app/views/ExposureHistory/HasBeenExposedDetail.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {
+  Typography as TypographyStyles,
+  Outlines,
+  Colors,
+  Spacing,
+} from '../../styles';
+
+type Posix = number;
+
+interface Exposure {
```
johnschoeman, line 16: this should get pulled up in a future commit, leaving here until its clear what the data model wants to be.

app/views/ExposureHistory/HasBeenExposedDetail.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+interface HasBeenExposedDetailProps {
+  exposure: Exposure;
+}
+
+const HasBeenExposedDetail = ({
+  exposure: { date, possibleExposureTimeInMin, currentDailyReports },
+}: HasBeenExposedDetailProps): JSX.Element => {
+  const exposureDate = dayjs(date).format('dddd, MMM DD');
```
johnschoeman, line 29: all of this should be translated in a future commit, given the demo on is tomorrow, it makes sense to hard code the content.  We can take a pass at adding translations next week.

johnschoeman:
also note that i misunderstood the flow at first, i thought it would be a running list of exposures.  The actual intended behavior is that a user can tap a day on the calendar and see the info for that day. Going to add that functionality and remove the toggle exposure in a future commit.

mattThousand (approved):

app/views/ExposureHistory/DetailedHistory.js
```diff
@@ -14,60 +19,94 @@ import { SingleExposureDetail } from './SingleExposureDetail';
...
-      <ExposureCalendarView weeks={3} history={history} />
+  const Divider = () => {
+    return <View style={styles.divider} />;
+  };
 
-      <Divider />
+  const gpsExposureInfo = () => {
+    return (
+      <>
+        {exposedDays.map(({ exposureMinutes, date }) => (
```
mattThousand, line 32: 🍎 's `EN` api calls this property `duration` - should we do the same for consistency? 🤔

tparesi (approved):
LGTM

app/views/ExposureHistory/DetailedHistory.js
johnschoeman: @mattThousand , we definitely should, but lets grab that in a future commit.

PR #993, opened 2020-06-11
Allow users to select day on en calendar

johnschoeman:
Why:
We would like for users to be able to select various days on the
calendar and see any detail information about their exposure history on
that day.

This commit:
Introduces the above functionality. Currently much of the existing
calendar views are tightly coupled to the gps exposure calculations and
determinations. This commit introduces a parallel set of calendar
components to be rendered on the bt version of the app. The notion of an
ExposureHistory was introduced and the determination of data was kept
separate from the view rendering logic. Ideally we can introduce the
same separation of concerns on the gps side in a future commit and
remove redundant components. We did not do that in this pr for sake of
time and to keep the commits smaller.

Note that we introduced some fake data to mock out the exposure history
until we can get the actual data wired up.

### Gif

exposure-history


app/ExposureNotificationContext.tsx
```diff
@@ -14,9 +42,43 @@ interface ExposureNotificationsState {
 
 const initialStatus: ENAuthorizationStatus = 'notAuthorized';
 
+const now: Posix = Date.now();
+
+const daysAgo = [...Array(21)].map((_v, idx: number) => {
+  return 20 - idx;
+});
+
+const initialExposureHistory: ExposureHistory = daysAgo.map(
```
johnschoeman, line 51: @mattThousand, (cc: rebecca) this is all fake data for what it looks like we will need to populate the exposure notification history screen.

johnschoeman: lets find a time to discuss it would be good to get the schema well thought out, this is based off what makes sense for the requested views, not sure if its the best structure for what is going on on the native side.

johnschoeman:
ping @mattThousand , could i also get a thumbs up here?

app/ExposureNotificationContext.tsx
```diff
@@ -1,11 +1,39 @@
 /*global JSX*/
 import React, { createContext, useState } from 'react';
+import dayjs from 'dayjs';
 
 import * as ExposureNotifications from './exposureNotificationsNativeModule';
 
 export type ENAuthorizationStatus = 'authorized' | 'notAuthorized';
 
+type Posix = number;
```
tparesi, line 9: Is this used for the date? I thought typescript recognized a Date interface, but I could be wrong.

app/views/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -0,0 +1,131 @@
...
+      return <NoKnownExposureDetail datum={exposureDatum} />;
+    }
+  }
+};
+
+interface PossibleExposureDetailProps {
+  datum: Possible;
+}
+
+const PossibleExposureDetail = ({
```
tparesi, line 45: What are your thoughts about having these all in the same file? Are they three different components?

tparesi (approved):
Functionality is great. Comments were more for my knowledge. Also, feels like we are starting to diverge from GPS a bit more. Hopefully we can work to keep that at a minimum.

app/ExposureNotificationContext.tsx
johnschoeman: it is! using posix directly a thing a picked up from working with elm, just keeping things in posix(number of milliseconds since epoch) as to not need to deal with timezone stuff/minimize data payload size/no parsing issues/generally not need to care about all of the hard time stuff.  Once we figure out more about the en exposure data, it might make sense to choose a different format.

app/views/ExposureHistory/ExposureDatumDetail.tsx
johnschoeman: They are all quite similar at this point and functionally service the same purpose, so i left them in. But I could see a world where they each have their own file. I see it as style preference at this point.

johnschoeman:
> Functionality is great. Comments were more for my knowledge. Also, feels like we are starting to diverge from GPS a bit more. Hopefully we can work to keep that at a minimum.

totally agree, I plan to follow up with this consolidate the differences/bring the gps side to a similar format.

PR #994, opened 2020-06-11
Exposure Notification Home Status Screens

tparesi:
#### Description:

Allows user to see whether notifications are enabled/disabled in the bluetooth version of the app

Home Screen displays "exposure notifications are on" or "notifications disabled" 

Added a bluetooth tracking version of the notification service

Added shallow functionality to allow the notification service status to be toggled from the debug menu

#### Screenshots:

Screen Shot 2020-06-11 at 3 05 10 PM
Screen Shot 2020-06-11 at 3 05 17 PM
Screen Shot 2020-06-11 at 3 05 31 PM

#### How to test:

go into debug menu and use the "toggle exposure notifications" button


app/exposureNotificationsNativeModule.ts
```diff
@@ -48,13 +48,13 @@ export const simulatePositiveDiagnosis = async (
   debugModule.simulatePositiveDiagnosis(cb);
 };
 
-export const disableExposureNotifications = async (
+export const toggleExposureNotifications = async (
```
johnschoeman, line 51: is it necessary to need to go to the native layer for this? could we have some state on the ExposureNotificationsContext of a boolean? in fact i believe we already have this.

app/services/ExposureNotificationService.ts
```diff
@@ -0,0 +1,42 @@
+// eslint-disable-next-line
+declare const global: any;
+
+global.ExposureNotificationsOn = true;
```
johnschoeman, line 4: mmm, is there a way we could do this without using global?

app/services/ExposureNotificationService.ts
```diff
@@ -0,0 +1,42 @@
...
+  | 'DEVICE_EXPOSURE_NOTIFICATIONS_OFF'
+  | 'ALL_CONDITIONS_MET';
+
+interface ExposureNotificationsState {
+  canTrack: boolean;
+  reason: ENStatusReason;
+  hasPotentialExposure: boolean;
+}
+
+export default class ExposureNotificationService {
```
johnschoeman, line 16: What do you think about not using classes here?

It also looks like a lot of this might want to live in the context component.

app/views/Main.js
```diff
@@ -7,82 +7,78 @@ import Colors from '../constants/colors';
 import { isGPS } from '../COVIDSafePathsConfig';
 import { checkIntersect } from '../helpers/Intersect';
 import BackgroundTaskServices from '../services/BackgroundTaskService';
-import LocationServices, { Reason } from '../services/LocationService';
-import { ExposureNotificationNotAvailablePage } from './main/ExposureNotificationNotAvailablePage';
+import LocationServices from '../services/LocationService';
+import ExposureNotificationService from '../services/ExposureNotificationService';
```
johnschoeman, line 11: I wouldnt follow the LocationService as a guide here, i dont think they we ultimately be very similar work flows.  Its not clear to me what the ExposureNotificationService would even do as most of the EN interaction is happing in the native layer?

app/views/Main.js
```diff
@@ -7,82 +7,78 @@ import Colors from '../constants/colors';
...
+    setTrackingInfo(state);
+  }, [tracingService, setTrackingInfo]);
 
   const handleBackPress = () => {
     BackHandler.exitApp(); // works best when the goBack is async
     return true;
   };
 
   useEffect(() => {
-    if (isGPS) {
```
johnschoeman, line 49: should this be removed?

app/views/Settings/ENDebugMenu.tsx
```diff
@@ -22,6 +22,9 @@ import {
 } from '../../exposureNotificationsNativeModule';
 import ExposureNotificationContext from '../../ExposureNotificationContext';
 
+// eslint-disable-next-line
+declare const global: any;
```
johnschoeman, line 26: this strikes me as something we should find an alternative way to do.

johnschoeman (commented):
a few comments!

I think we could tackle this a bit more simply and have some local component state on the context!  And the main screen can read from that with a 'useContext' hook.

app/exposureNotificationsNativeModule.ts
tparesi: I think this will become toggle instead of disable anyways going forward, so not sure if it is a big deal if we keep for now.

app/services/ExposureNotificationService.ts
tparesi: Yeah - globals aren't great. What did you have in mind? This was a quick and dirty since we knew it was going to come out in a couple of days anyways.

app/services/ExposureNotificationService.ts
tparesi: We used the class to match the implementation details of the Location service. Trying to minimize the isGPS logic as much as possible in these shared components. I can look into refactoring the Main.js a bit to make this work with a context, but then this might cause us to but heads with the GPS implementation.

app/views/Main.js
tparesi: If we use the ExposureNotificationService, I think we can have the same implementation. Meaning that the ExposureNotificationService would be called to check and update the stateInfo similarly to GPS.

app/views/Main.js
tparesi: We should chat about how you want to use the ExposureNotificationContext instead. I was trying to minimize the isGPS logic and use dependency injection.

johnschoeman (approved):
Given the demo is tomorrow, lets ship this and do any required refactoring early next week!

app/views/Main.js
johnschoeman: got it, that makes sense to me as a strategy, I wonder if we could even do this at a higher level in the tree than the MainScreen, feels like this wants to be during the app initialization step.

app/services/ExposureNotificationService.ts
johnschoeman: thanks for the context, makes sense to leave as is and we can do any refactorings later.

app/exposureNotificationsNativeModule.ts
johnschoeman: probably nbd!

app/services/ExposureNotificationService.ts
johnschoeman: chatted irl, might be able to save the state in a context!

PR #999, opened 2020-06-11
Affected User Flow on Settings Screen

tparesi:
#### Description:

This commit adds the affected user flow to the settings screen for Bluetooth only.

#### Screenshots:

Screen Shot 2020-06-11 at 5 08 45 PM
Screen Shot 2020-06-11 at 5 08 55 PM
Screen Shot 2020-06-11 at 5 09 02 PM
Screen Shot 2020-06-11 at 5 09 09 PM
Screen Shot 2020-06-11 at 5 09 14 PM


jeffstolz (approved):

app/Entry.js
```diff
@@ -66,7 +67,9 @@ const ExportStack = () => (
       ...SCREEN_OPTIONS,
       cardStyleInterpolator: fade,
       gestureEnabled: false,
-    }}>
+    }}
+    initialRouteName={isGPS ? 'ExportSelectHA' : 'ExportIntro'}>
```
johnschoeman, line 75: For sake of the demo dont actually do this, but, do you think its possible to consolidate the flows between gps and bte such that they have the same number of screens?

app/TracingStrategyAssets.ts
```diff
@@ -1,9 +1,11 @@
 import { useTranslation } from 'react-i18next';
 
-import { Images } from './assets';
+import { Icons, Images } from './assets';
 import { isGPS } from './COVIDSafePathsConfig';
 
-export function useAssets(): Record<string, string> {
+type Asset = string | ((name: string) => string) | null;
```
johnschoeman, line 6: Thanks for making this type a bit more accurate! Though it does jump out at me as a bit board. What do you think about splitting up the various asset types and such that the functions return values a more explicit?
something like: 
```
export const useStrategyTranslation = (): Record<string, string> => { ...

export const useStragteyImage = (): Record<string, Image> => {

export const useStrategyContent = (): Record<string, (name: string) => string> => {
```
?

app/TracingStrategyAssets.ts
```diff
@@ -91,6 +93,53 @@ export function useAssets(): Record<string, string> {
     ? t('label.home_no_contact_subtext')
     : t('label.home_no_contact_subtext_bluetooth');
 
+  // Export Pages
+
+  const exportExitRoute = isGPS ? 'ExportStart' : 'SettingsScreen';
```
johnschoeman, line 98: What do you think about moving the screen names outside of the tracingStrategy assets ?

app/views/Export/ExportIntro.js
```diff
@@ -0,0 +1,40 @@
...
+    dispatch(getHealthcareAuthorities());
+  }, [dispatch]);
+  const authorities = useSelector(healthcareAuthorityOptionsSelector);
+  const selectedAuthorityDummy = authorities[0];
+
+  const onNext = () =>
+    navigation.navigate('ExportCodeInput', {
+      selectedAuthority: selectedAuthorityDummy,
+    });
+  const onClose = () => navigation.navigate('SettingsScreen');
```
johnschoeman:
```suggestion
  const onClose = () => navigation.goBack();
```
minor thing, but if this is the first screen its probably better to pop a screen off than to do a forward navigation.

johnschoeman (approved):
LGTM, a few comments, feel free to skip em for sake of time for getting stuff ready for the demo.

app/Entry.js
tparesi: I think this is a question for design, but I remember someone saying this is just a temporary flow for the demo. I didn't want to do too much here bc of that.

app/TracingStrategyAssets.ts
tparesi: I like this! Is it cool if I do this on a different PR?

app/TracingStrategyAssets.ts
tparesi: This was a bit weird bc of the flow that is in GPS, which I think is changing. On the GPS side, the feature flag puts this on the tab navigation. I am hoping that we can coordinate with design to have a more similar flow and not have to do this. We have put these routes in here in the past as an intermediate and we usually end up being able to get rid of them once design for GPS and BT catch up.

I can move them to a different hook or something if we see this pattern staying in place.

app/views/Export/ExportIntro.js
tparesi: Love it.

PR #1001, opened 2020-06-11
SAF-693 Fixed scrollable "About" page 

mikhaildmitrienko:
#### Description:

Prevents "About" page from scrolling even though there is no content to scroll to.

#### Linked issues:

Fixes https://pathcheck.atlassian.net/browse/SAF-693

#### Screenshots:

Screen Shot 2020-06-11 at 6 39 43 PM

#### How to test:

Go to "About" page and try to scroll.


app/views/About.js
```diff
@@ -114,6 +114,7 @@ export const AboutScreen = ({ navigation }) => {
         <View style={styles.spacer} />
         <View style={styles.spacer} />
       </ScrollView>
+
```
johnschoeman: you can run `yarn format` locally to run the prettier formatter! which you will need to do to pass the validation checks.

app/views/About.js
```diff
@@ -123,7 +124,7 @@ const styles = StyleSheet.create({
     flexDirection: 'column',
     width: '100%',
     backgroundColor: Colors.INTRO_WHITE_BG,
-    paddingHorizontal: 26,
+    paddingHorizontal: 40,
```
johnschoeman, line 126: its not clear to me how this keeps the page from scrolling?  If the intent is to remove the scrolling could we change the `<ScrollView>` to a `<View>`?

johnschoeman (approved):
thanks for the commit! a few comments.

app/views/About.js
mikhaildmitrienko: Hi John, thanks for your feedback. When I tried changing it to a View, the formatting* was distorted, which led to the more “hacky” solution of just increasing the padding slightly so that the scroll view doesn’t scroll anymore. Hope this isn’t too big of an issue.

mikhaildmitrienko: Stylesheet* not CSS

johnschoeman: Hmm, certainly wouldnt call it a big issue, but ideally we are laying out our views in such a way that we can be declarative without oddities in the look of it.  That said i think any clean up here would make sense in a future commit.

Once you get the formatting validation passing we can merge this in!

app/views/ChooseProvider.js
```diff
@@ -325,7 +325,7 @@ class ChooseProviderScreen extends Component {
               name='AuthoritiesMenu'
               renderer={SlideInMenu}
               style={{ padding: 20 }}>
-              <MenuTrigger>
+              <MenuTrigger disabled>
```
johnschoeman: Was this change intentional?

mikhaildmitrienko: That is a commit for a different PR, I must have pushed it to this branch on accident. I can go ahead and change it.

PR #1002, opened 2020-06-12
Partners screens, Redux State for Selected Authorities, new YAML updates

JacobJaffe:
Slough of changes to get selecting HAs working now with updated Yaml config. @vlasac This should enable you to connect to the selected health authority now via redux. 

This builds from the work done by @alpita-masurkar in a way to work with our redux store & our typescript changes around the api layers. 

Next steps are to introduce the manual input url for the HA selection to this flow.|

This also changes some of the export logic to improve error handling & make compatibility changes to the 1.0.1 yaml change (which introduces a configuration endpoint). 

image

image

image


app/api/healthcareAuthorities/getConfigurationApi.ts
```diff
@@ -0,0 +1,24 @@
+import type { HealthcareAuthority } from './getHealthcareAuthoritiesApi';
+
+export type Configuration = {
+  name: string;
+  infoWebsiteUrl: string;
+  privacyPolicyUrl: string;
+  referenceWebsiteUrl: string;
+  notificationThresholdPercent: number;
+  notificationThresholdCount: number;
```
vlasac, line 9: There was a recent [change in spec](https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/81231882/Hashing+of+GPS+points+-+Design#Notification-of-sensitivity-controls), the `count` will soon be replaced by `timeframe` because the `count` doesn't say anything about the period of time in which those matches occurred.

app/api/healthcareAuthorities/getCursorApi.ts
```diff
@@ -11,10 +11,6 @@ export type Cursor = {
   version: string;
   name: string;
   publish_date_utc: number;
-  info_website_url: string;
-  api_endpoint_url: string; // Where cursor is hosted
-  privacy_policy_url: string;
-  reference_website_url: string;
   notification_threshold_percent: number;
   notification_threshold_count: number;
```
vlasac, line 15: same as before, see this [change in spec](https://pathcheck.atlassian.net/wiki/spaces/TEST/pages/81231882/Hashing+of+GPS+points+-+Design#Notification-of-sensitivity-controls)

vlasac (changes requested):

app/api/healthcareAuthorities/getConfigurationApi.ts
JacobJaffe: Nice catch -- I left a comment on the spec to make sure the configuration endpoint gets updated appropriately

JacobJaffe:
@vlasac I think we should wait until those change are live, so that the mobile app is in sync with the current API?

app/api/export/exportConsentApi.ts
```diff
@@ -5,7 +5,7 @@ const exportConsentApi = async (
   consent: boolean,
   code: number,
 ): Promise<void> => {
-  const endpoint = `${authority.ingest_url}/consent`;
+  const endpoint = `${authority.public_api}/consent`;
```
johnschoeman, line 8: lt might make sense to group files by api rather than by endpoint, looks like theres some multiple places we need to updated the same thing.  This would also allow for pulling out helpers like defaultHeaders and baseUrl.

app/store/reducers/healthcareAuthoritiesReducer.ts
```diff
@@ -42,7 +45,20 @@ const healthcareAuthoritiesReducer = createReducer(initialState, (builder) =>
     .addCase(getHealthcareAuthorities_failure, (state, { meta: { error } }) => {
       state.request.status = ApiStatus.FAILURE;
       state.request.errorMessage = error.message;
-    }),
+    })
+    .addCase(
+      toggleSelectedHealthcareAuthorityAction,
+      (state, { payload: { authority, overrideValue } }) => {
```
johnschoeman, line 51: is this something that needs to be in the global state? could it stay local to the component?

app/views/Partners/PartnersEdit.tsx
```diff
@@ -0,0 +1,124 @@
...
+              paddingHorizontal: 24,
+            }}
+            onPress={() => toggleSelected(HA)}>
+            <View
+              style={{
+                justifyContent: 'space-between',
+                flexDirection: 'row',
+              }}>
+              <Typography
+                style={{ fontWeight: '500', paddingRight: 30 }}
```
johnschoeman, line 96: styles like this are pretty hard to read and understand the intent of. What do you think about introducing a stylesheet and given these names?

something like:
```
<Typography style={styles.titleText} ...
```

app/views/Partners/PartnersOverview.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const styles = StyleSheet.create({
+  backgroundWrapper: {
+    backgroundColor: Colors.INTRO_WHITE_BG,
+    flex: 1,
+  },
+  divider: {
+    height: StyleSheet.hairlineWidth,
+    width: '100%',
+    backgroundColor: Colors.VIOLET_BUTTON_LIGHT,
```
johnschoeman, line 110: Why are we using button colors for the background color?

app/views/Partners/PartnersOverview.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Typography
+          style={{
+            // Prevent from forcing overflow on parent
+            flex: 1,
+          }}>
+          {t('authorities.automatically_follow')}
+        </Typography>
+        <View style={{ width: 16 }} />
+        <Switch
+          trackColor={{ true: '#6979F8', false: '#6979F8' }}
```
johnschoeman: should we be using a common style here?

app/views/Partners/PartnersEdit.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const dispatch = useDispatch();
+
+  // Fetch list on screen mount
+  useEffect(() => {
+    dispatch(getHealthcareAuthoritiesAction());
+  }, [dispatch]);
+
+  const authorities = useSelector(healthcareAuthorityOptionsSelector);
+  const selectedAuthorities = useSelector(
+    selectedHealthcareAuthoritiesSelector,
```
johnschoeman, line 57: What do you think about inlining these selectors? As a reader of the file you need to have multiple other files open to be able to understand what is happening here.

johnschoeman (approved):
Overall looks good! some minor comments.

app/store/reducers/healthcareAuthoritiesReducer.ts
JacobJaffe: yeah, this will need to be consumed by the background running task to determine intersections

yonatanmk (approved):

app/views/Partners/PartnersOverview.tsx
JacobJaffe: Just a color match. More motivation to standardize our colors, the current color palettes are outdated

app/views/Partners/PartnersOverview.tsx
JacobJaffe: Good call, I'll ensure these are in Colors

vlasac (approved):

app/views/Partners/PartnersOverview.tsx
johnschoeman: as a heads up, this one consolidated the colors constants all into the styles module: https://github.com/Path-Check/covid-safe-paths/pull/1015

PR #1003, opened 2020-06-12
Refactor <Main /> to use updated service screens

Patrick-Erichsen:

#### Description:

Updates the screens in `<Main />` to reflect the most recent UI prototypes that contain the following states:

- Tracing off (Unique screens for GPS & BTE)
- Notifications off
- No health authorities in user's region
- No health authorities selected 
- All services on (Unique screens for GPS & BTE)

In addition to these larger updates, the following minor changes are included in this PR:

- Changes `paddingTop` on tab nav to better accommodate small screens
- Breaks down home screen language strings into `"gps"`, `"bluetooth"`, and `"shared"` keys
- Adds a new `hasAuthoritiesInBounds()` method to the HCA Service 

#### Linked issues:

- https://pathcheck.atlassian.net/browse/SAF-283

##### `<AllServicesOnScreen />`
Screen Shot 2020-06-11 at 11 39 05 PM

##### `<TracingOffScreen />`
Screen Shot 2020-06-11 at 11 41 14 PM

##### `<NotificationsOffScreen />`
Screen Shot 2020-06-11 at 11 40 19 PM

##### `<SelectAuthorityScreen />`
Screen Shot 2020-06-11 at 11 40 55 PM

##### `<NoAuthoritiesScreen />`
Screen Shot 2020-06-11 at 11 37 41 PM

#### How to test:

- `<TracingOffScreen />`
  - Navigate through onboarding in both GPS and BTE, and deny tracing permissions
- `<NotificationsOffScreen />`
  - Navigate through onboarding on iOS and deny notification permissions
- `<NoAuthoritiesScreen />`
  - Navigate through onboarding and accept all permissions. Unless you are in Haiti, you aren't in the bounds of an authority
- `<SelectAuthorityScreen />`
  - TODO: Not sure how to mock GPS location
- `<AllServicesOnScreen />`
  - TODO: Mock GPS location, and then subscribe to an authority in the mocked location.


app/views/main/ServiceOffScreens/ExposureNotificationNotAvailable.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { ServiceOffScreen } from './Base';
+
+export const ExposureNotificationNotAvailableScreen = (): JSX.Element => {
+  const {
+    exposureNotificationsNotAvailableHeader,
+    exposureNotificationsNotAvailableSubheader,
+  } = useAssets();
+
+  return (
+    <ServiceOffScreen
```
johnschoeman, line 12: All of these various flavors of a `ServiceOffScreen` which are only different by a some conditionally rendered content suggests to me that we really have a domain concept of one screen that has a few different internal states rather than six truely separate screens.

While your code is very readable and makes sense, for long term maintainability its generally preferable to prefer fewer components/files where possible as this reduces the surface area of things that need to be understood and maintained. Another way i like to think about it is, we should be preferring internally managed stated over passed in props when possible, which i believe is encouraged pretty heavily in the react docs as well.

What do you think about having one `HomeIndicatorScreen` which understands how to render the correct content based off the state of the application?  something like:

```typescript
const HomeIndicatorScreen = ({ navigation }: HomeIndicatorScreenProps) => {
    const { permissionState } = useContext(myPermissionContext)
    const [myOtherPermissionState, setMyOtherPermissionState] = useState<MyPermissionState>({ ... })
    .... other bits of application state that matter

   const determineScreenContent = () => {
       if (!trackingInfo.canTrack) {
           return [tracingOffHeaderText, tracingOffSubheaderText]
        } else if (other possible state) {
          return ...
   }
    const [headerText, subheaderText] = determineScreenContent()

   return (
      <View>
            <Text>{headerText}</Text>
        ....
```


you could even get fancy and use a typescript union or enum to ensure that impossible states remain impossible with something like:

```typescript
type HomeIndicatorState = TracingEnabled | TracingOff | NotificationsOff | NoAuthorities | ... | Unknown


const HomeIndicatorScreen = ({ navigation }: HomeIndicatorScreenProps) => {
   const [indicatorState, setIndicatorState] = useState<HomeIndicatorState>(Unknown)

   const determineIndicatorState = async (...maybeSomeArgs): HomeIndicatorState => {
           const state = await ....logic
           ...do other logic
           setIndicatorState(theIndicatorState)
        }
        
     useEffect(() => {
           determineIndicatorState()
        ...
        },[])

       const inidicatorContent = () => {
             switch (indicatorState) {
                  case (TracingEnabled): {
                       return [enabledHeader, enabledSubHeader]
                  }
                  case (TracingOff): {
                       return [offHeader, offSubHeader]
            ...
        }

      const [headerText, subHeaderText] = indicatorContent()

         return (
               <View>
                     <Text>{headerText}</Text>
           ...
```

which keeps all the business logic in one spot and separated completely from the view logic.

app/views/Main.js
```diff
@@ -66,26 +82,19 @@ export const Main = () => {
...
-  let page;
+  let screen;
 
-  if (trackingInfo.canTrack) {
-    if (trackingInfo.hasPotentialExposure) {
-      page = <ExposurePage />;
-    } else {
-      page = <NoKnownExposure />;
-    }
+  if (!trackingInfo.canTrack) {
```
johnschoeman, line 87: quite the improvement here!

johnschoeman (approved):
One rather lengthy thought!  Overall looks great to me though!

app/views/main/ServiceOffScreens/ExposureNotificationNotAvailable.tsx
Patrick-Erichsen: Lengthy but a great piece of feedback that is much appreciated! That all makes a lot of sense to me, and I can see the benefits of keeping all of that logic consolidated in a single logical view.

JacobJaffe (approved):
Thanks for adding those snapshots! 🚢

PR #1006, opened 2020-06-12
Replace BTE with BT within application

tparesi:
#### Description:

Initially, BTE was used even though it didn't stand for anything. I think this came from
BLE, but got lost in translation. Now the PathCheck apps are GPS and BT, so removing
use of BTE within the app will hopefully unify the language and remove the use of BTE
in conversations.


tparesi:
We should probably wait to get this in until after the demo, but I still see people using the BTE name, so I wanted to remove it from the codebase as an attempt to minimize this confusion.

android/fastlane/Fastfile
```diff
@@ -109,8 +109,8 @@ platform :android do
       track: 'beta',
       release_status: 'draft',
       skip_upload_apk: true,
-      package_name: 'org.pathcheck.covidsafepathsBte',
```
johnschoeman, line 112: This one i dont believe we can change at this point.

johnschoeman (commented):
But now the gps side and the bt side wont have the same number of letters so they will always be one character off!   (jk)

This looks fine to me. I would just doubly triple check the ios and android builds still work for dev to production schemes.  Additionally i dont think we are able to change package names for android.

Theres additionally a lot of config that needs to be verified for the build targets if we change those names. So I might consider only implementing this change in the rn layer.

android/fastlane/Fastfile
tparesi: 👍 I will revert this and let's take a closer look at anything else we may want to revert.

johnschoeman (approved):
Thanks for grabbing this!

package.json
```diff
@@ -10,9 +10,9 @@
     "format": "yarn prettier --write",
     "lint": "eslint ./",
     "validate": "yarn prettier --list-different && yarn tsc && yarn lint",
-    "run-android-bte": "react-native run-android --variant=bteDebug --appId=org.pathcheck.covidsafepathsBte",
+    "run-android-bt": "react-native run-android --variant=btDebug --appId=org.pathcheck.covidsafepathsBt",
```
mattThousand, line 13: I believe our app id on the play store is `org.pathcheck.covidsafepathsBte`, which means that changing to `org.pathcheck.covidsafepathsBt` will create a new store listing when it is uploaded. Since we haven't launched yet this is ok, just flagging it here

android/app/build.gradle
```diff
@@ -212,8 +212,8 @@ android {
             applicationId 'org.pathcheck.covidsafepaths'
             dimension "version"
         }
-        bte {
-            applicationId 'org.pathcheck.covidsafepathsBte'
+        bt {
+            applicationId 'org.pathcheck.covidsafepathsBt'
```
mattThousand, line 216: here too - we should keep the appId in `package.json` consistent with gradle

android/fastlane/Fastfile
mattThousand: as long as we keep `package.json` in sync with the appId we're using in the android build files, we're ok. If we use an appId other than `org.pathcheck.covidsafepathsBte` it will create a new listing on the store, but now would be a better time to make the switch than later, if we do indeed want to migrate 🤔

tparesi: What do we think? I can go either way on this one.

I will change the appId back to Bte for now to get this in.

package.json
tparesi: Will change it back - thanks!

android/app/build.gradle
tparesi: Changed - thanks!

android/fastlane/Fastfile
johnschoeman: if its fine to change, then i would prefer to use 'bt'

mattThousand: let's go with `bt`  - I'll make sure Dick Lucas and Sam know 👍

tparesi: Awesome - I changed everything back to bt. Should I merge this or would you, @johnschoeman and @mattThousand, want to take one more look?

octohub:
PR #1027 is a branch of the one in this PR. It fixes the `Build / build_staging_android_bt` build check. @tparesi @johnschoeman Feel free to merge the #1027 branch into this one or simply merge #1027.

PR #1007, opened 2020-06-12
Self Assessment Navigation Bugfix

mattThousand:
### Why
We'd like to return to the root of the self assessment stack when pressing the `X` button and when pressing `Done` at the end of the self assessment.

### This commit
This commit navigates the user to the root of the self assessment stack upon pressing the `X` button during the course of the self assessment and also when the `done` button is pressed at the conclusion of the self assessment. This commit also hides the `X` button on the self assessment start screen, as it is redundant there.

Jun-12-2020 15-31-37


app/views/assessment/index.js
```diff
@@ -87,8 +87,7 @@ const Assessment = ({ navigation }) => {
   const AssessmentCancel = () => (
     <TouchableOpacity
       onPress={() => {
-        navigation.goBack();
-        navigation.popToTop();
+        navigation.navigate('Start');
```
JacobJaffe: I realize this was already `Start`, but I'd recommend making route names more specific so there's no risk of duplicates in the tree.
e.g.
`AssessmentStart`

johnschoeman: perhaps `SelfAssessmentStart`

johnschoeman (approved):
thanks for grabbing this!

tparesi (approved):
LGTM - thanks!

JacobJaffe (approved):
🥳 Thanks for the naming changes

app/views/assessment/AssessmentStart.js
```diff
@@ -18,12 +18,12 @@ import { Colors } from '../../styles';
...
 
 /** @type {React.FunctionComponent<{}>} */
-export const Start = ({ navigation }) => {
+export const AssessmentStart = ({ navigation }) => {
   let { t } = useTranslation();
   return (
     <Info
       ctaAction={() => {
-        navigation.push('Question', {
+        navigation.push('AssessmentQuestion', {
```
johnschoeman, line 26: should we use the Screens from navigation here?
```suggestion
        navigation.push(Screesn.AssessmentQuestion, {
```

app/views/assessment/index.js
```diff
@@ -122,31 +120,32 @@ const Assessment = ({ navigation }) => {
...
               cardStyleInterpolator: CardStyleInterpolators.forHorizontalIOS,
               cardStyle: {
                 backgroundColor: 'transparent',
               },
             }}>
             <Stack.Screen
-              component={Start}
-              name='Start'
+              component={AssessmentStart}
+              name='AssessmentStart'
```
johnschoeman, line 132:
```suggestion
              name={Screens.AssessmentStart}
```

PR #926, opened 2020-05-28
[WIP - SAF 282] Implement self reporting screens

zhaccc:
#### Description:

This PR fixes comments on the existing PR that was open a while ago but abandoned. It also implements new design and does some tweaks to existing code.

#### Screenshots:

self-assessment

PR #1011, opened 2020-06-13
Remove react-native-screens navigator

johnschoeman:
Why:
A previous commit introduced a native navigator and an initialization
function for react-native-screens. When doing so it broke unit specs on
ci.

This commit:
Switches the native stack navigator in the self assessment flow to a
plain react-navigation navigator, similar to every other navigator in
the application.

Note that there was no explanation provided in the PR, #926, that
introduced this as to why it was being introduced. It is possible that
there was a good reason and we might need to add this back in, but as of
this commit is not clear why we need to be using this package.

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,6 +1,5 @@
 import React, { useEffect } from 'react';
 import { MenuProvider } from 'react-native-popup-menu';
-import { enableScreens } from 'react-native-screens';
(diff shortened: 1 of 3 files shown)


tparesi (approved):
LGTM - thanks. 

I think this was leftover from the older self assessment PR we were working off of. Good catch.

PR #1012, opened 2020-06-13
Refactor settings and debug screens

johnschoeman:
Why:
Currently the Settings and Debug Screen have a few extracted components
that have more functionality built in than is required. Additionally
there are few places where react components are being used to implement
style.

We would like to minimize the number of components that need to be
maintained and prefer to use style sheets to implement designs when
possible as this is likely to need to be updated quickly in future iterations.

This commit:
Consolidates extra components into the settings and debug screen and
replaces component props with styles where possible.


cwillett77 (approved):
LGTM 👍

johnschoeman:
closing and including changes in: https://github.com/Path-Check/covid-safe-paths/pull/1013

PR #1016, opened 2020-06-14
Introduce exposure history context

johnschoeman:
Why:
A previous commit added the ability to see exposure history data in the
exposure history screen for the bt build. Given that the gps and bt
versions of the app will eventually have the same notion of the exposure
history, we would like to also include this behavior in the gps build.

This commit:
Introduces an exposure history context which takes the logic for
exposing the exposure history from the ExpsoureNotificationsContext such
that we can reuse it in the gps side. We removed the gps specific code
from the ExposureHistory screen such that we no longer needed to do any
data manipulation in the user facing screen components, which allows us
to reuse the screen for both builds and in general provides for a better
separation of view and model logic.

Co-Authored-By: mattthousand <matt@nicethings.io>


tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
     "lib": [
-      "es2018"
+      "es2019"
```
johnschoeman, line 9: This change was to allow us to use `flatMap`

app/views/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -47,11 +48,14 @@ interface PossibleExposureDetailProps {
...
+  const exposureDurationText = TimeHelpers.durationMsToString(
+    possibleExposureTimeInMin * 60 * 1000,
+  );
   const navigation = useNavigation();
   const exposureDate = dayjs(date).format('dddd, MMM DD');
-  const exposureTime = `Possible Exposure Time: ${possibleExposureTimeInMin}`;
+  const exposureTime = `Possible Exposure Time: ${exposureDurationText}`;
   const dailyReports = `Current daily reports: ${currentDailyReports}`;
-  const explainationContent = `For ${possibleExposureTimeInMin} consecutive minutes, your phone was within 10 feet of someone who later received a confirmed positive COVID-19 diagnosis.`;
+  const explainationContent = `For ${exposureDurationText}, your phone was within 10 feet of someone who later received a confirmed positive COVID-19 diagnosis.`;
```
mattThousand, line 58: small typo fwiw, ~explaination~ explanation

mattThousand: also, I may be reading this incorrectly, but when the string is interpolated, won't this read: "for 40, your phone was within..." - i.e. missing "minutes"?

johnschoeman: the dayjs humanize method which is used for this provides minutes, hours, days, etc.

mattThousand (approved):

ArgoJRB (approved):
LGTM!

PR #1017, opened 2020-06-15
Adjust badge alignment.

mattThousand:
### Why
We'd like tab bar badges to be aligned relative to tab bar icons such that they reflect the designs in figma and also conform to platform-specific convention.

### This Commit
This commit modifies the positioning of the badge icon relative to the tab bar icons to more closely hew to the figma designs, and also ensures that, if the badge and the icon do overlap, the badge sits above the icon (per iOS/Android convention).

### Before:
Screen Shot 2020-06-14 at 9 36 35 PM

### After:
Screen Shot 2020-06-14 at 9 32 04 PM

### DIFF
--- a/app/Entry.js
+++ b/app/Entry.js
@@ -151,8 +151,8 @@ const MainAppTabs = () => {
   const applyBadge = (icon) => {
     return (
       <>
-        <View style={styles.iconBadge} />
         {icon}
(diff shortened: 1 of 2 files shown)


johnschoeman (approved):
thanks for grabbing this!

tparesi (approved):
Thanks!!

PR #1018, opened 2020-06-15
Replace HCAService usage in Main, fix routing for export & partners

JacobJaffe:
Various code cleanup + fixes surrounding the change to our `Partners` stack of HA selection screens. 

- Adds partners screen routes to type navigation type files.
- Replaces HCAService logic in `main.js` with redux selector (Part of HCAService deprecation)
- Render simplification on `main.js` - No wrapper view is needed.
- Removes export link in settings, this should just be in tabs.
- Resolves old links to `ChooseProvider` to new screen of `PartnersEdit`
- Deletes`ChooseProvider` screen


app/views/Settings/index.tsx
```diff
@@ -135,15 +135,6 @@ const SettingsScreen = ({ navigation }: SettingsScreenProps): JSX.Element => {
           </NativePicker>
         </View>
 
-        <View style={styles.section}>
```
johnschoeman, line 138: My understanding is that we wanted to allow users to get to the reporting a positive test flow from the settings screen, should we be removing this?

johnschoeman (approved):
Great clean up! LGTM! one question.

app/views/Settings/index.tsx
JacobJaffe: If that's the case that's a BT only functionality -- on GPS the export flow is linked to as a main tab bar option. I was under the impression export is _not_ in BT right now?

johnschoeman: I might be a bit confused here, but this is a link to the "Affected User" flow which is different from the "Export my location data Flow" ?  the figma here: https://www.figma.com/file/79OwaKqHvSeriExpYzAvVN/PathCheck-%2F-Exploration-%2F-April%2C-May%2C-June-2020?node-id=3601%3A26330 , implies the Affected User flow is used by the BT app as well.

@tparesi , do you have any insight here?

johnschoeman: But yes, if this is just for the bt flow, we can definitely put it behind a !isGPS.

tparesi: hmm this might be a question for design. This should definitely be on the BT app on the settings screen as per the mocks John linked above. We can put it behind an !isGPS check for now and follow up with design.

PR #1019, opened 2020-06-15
Fix reference to share test result translation

tparesi:
#### Description:

The Report a Positive Test Result item on the settings screen was using the wrong translation key. This fixes that.

#### Screenshots:

Screen Shot 2020-06-15 at 1 39 35 PM]


johnschoeman (approved):
LGTM, though it might be in conflict with #1018

PR #1020, opened 2020-06-15
Convert typography component to ts remove theme

johnschoeman:
Why:
Currently the Typography component is implemented with significant logic
surrounding being able to allow the user to update the theme.
Additionally the component defines a few typography style classes
inline. Given that we do not want to support dynamic themes and would
like to have all of the styles consistently handled by the design
system, we would like to refactor the Typography component such that it
is only responsible for handling right to left language adjustments.

We would also like to have this component be covered by the ts compiler.

This commit:
Updates the Typography component to use type script, removes unused
logic, and uses styles from the common project styles. While converting
to ts, a handful of call sites were discovered to have typos and type
errors, which were also fixed.


app/components/Typography.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Text, TextStyle } from 'react-native';
+
+import { useLanguageDirection } from '../locales/languages';
+
+import { Typography as TypographyStyles } from '../styles';
+
+type TypographicUse = 'headline1' | 'headline2' | 'body1' | 'body2' | 'body3';
+
+interface TypographyProps {
+  use?: TypographicUse;
```
johnschoeman, line 11: ideally we could remove this prop and just use 'style' everywhere as this is the convention in RN and would be one less thing to need to maintain be understood by onboarding developers. We didnt remove it here as to reduce the amount of git/merge conflict churn.

app/views/Licenses.tsx
```diff
@@ -33,6 +32,8 @@ export const LicensesScreen = ({
   const { t } = useTranslation();
   const { legalHeader } = useAssets();
 
+  const legalHeaderText: string = legalHeader as string;
```
johnschoeman, line 35: the `Asset` type is a bit wonky and was causing some issues.  Text content is a string and not an Asset so it seems like we could refactor useAssets() to return Assets only hand handle the translations is a different way than the assets as to avoid some of these oddities in they type system.

app/locales/languages.d.ts
```diff
@@ -2,10 +2,15 @@ import {
...
   supportedDeviceLanguageOrEnglish as supportedDevice,
+  useLanguageDirection as languageDirection,
 } from './languages';
 
 export const LOCALE_LIST: Record<unknown, unknown> = localList;
 export const setUserLocaleOverride: (
   locale: string,
 ) => Promise<void> = setOverride;
 export const supportedDeviceLanguageOrEnglish: () => string = supportedDevice;
+export const useLanguageDirection: () =>
```
johnschoeman, line 13: ideally we update languages.js to ts as well at some point,

i initially thought to inline this function to the component as well, since it was a bit odd to have it pulled out imo, and wed get the type definitions as well, but it turns out the test suite (and the whole app) is depending on some setup that runs when this is imported.

It might makes sense to pull that initialization logic into a more obvious initialization place. This would be less prone to be forgotten about and we'd be able to mock it tests. Probably not a priority, but wanted to mention it as a thing.

tparesi (approved):
LGTM - thanks for taking the TS initiative

PR #1021, opened 2020-06-15
Add back Theme to EnableExposureNotifications

tparesi:
#### Description:

Shared Button Component still uses the parent's theme, so cannot remove until shared components are refactored.

#### Screenshots:

Screen Shot 2020-06-15 at 2 24 48 PM


johnschoeman (approved):
Looks good!, thanks for catching this.

PR #1026, opened 2020-06-15
Use same styles for buttons

johnschoeman:
Why:
Currently a few of the buttons on various screens are not using the same
button and text label styles.

This commit:
Updates the buttons to use the same styles.

Co-Authored-By: achen178 <achen178@gmail.com>
Co-Authored-By: ericwbailey <eric.w.bailey@gmail.com>

### Before

buttons-before

### **After**

buttons-after


tparesi (approved):
LGTM

PR #1028, opened 2020-06-15
Update entitlements for bte target

johnschoeman:
Why:
Currently we have the same entitlements file for both the BTE and GPS
targets in ios.

This commit:
Creates a new entitlements file for both versions of the app.


johnschoeman:
Outdated, will open a new pr

PR #1029, opened 2020-06-16
Custom URL entry for HA List YAML

JacobJaffe:
Allows using a custom URL for HA selection. This should be gated behind a feature flag. Feature flags are wonky right now & we have a fair amount of experimental demo code, so that'll be a follow up of mine to get feature flags working robustly.

This is meant to unblock QA for being able to test against a dynamic yaml list. 

Can test with: 
`https://raw.githubusercontent.com/Path-Check/trusted-authorities/jj/custom-url-authorities/testing/authorities.1.0.1.yaml` as the input url.


app/views/Partners/PartnersCustomUrlScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+      </View>
+    </NavigationBarWrapper>
+  );
+};
+export default PartnersScreen;
+
+const styles = StyleSheet.create({
+  wrapper: { padding: 24 },
+  textInput: {
+    fontSize: 20,
```
johnschoeman:
```suggestion
import { Colors, Forms, Typography as TypographyStyles } from '../../styles';

...
textInput: {
    ...TypographyStyles.formInput,
    ...Forms.textInput
```

We probably should be using common styles such that we can deliver on the whitelabeling product need without having to revisit every file.

app/views/Partners/PartnersEdit.tsx
```diff
@@ -118,6 +126,12 @@ const PartnersScreen = ({
         }
         ItemSeparatorComponent={() => <Separator />}
       />
+      <View style={{ padding: 24 }}>
+        <Button
+          label={t('authorities.custom_url')}
+          onPress={() => navigation.navigate('PartnersCustomUrl')}
```
johnschoeman:
```suggestion
          onPress={() => navigation.navigate(Screens.PartnersCustomUrl)}
```

app/views/Partners/PartnersCustomUrlScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+type PartnersEditScreenProps = {
+  navigation: NavigationScreenProp<NavigationState, NavigationParams>;
+};
+
+const PartnersScreen = ({
+  navigation,
+}: PartnersEditScreenProps): JSX.Element => {
+  const { t } = useTranslation();
+  const [url, setUrl] = useState('');
+  const didMountRef = useRef(false);
```
johnschoeman, line 22: why do we need a reference here? is there a way to do this declaratively?

app/views/Partners/PartnersCustomUrlScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const fetchCustomStatus = useSelector(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (state: any) => state.healthcareAuthorities.request.status,
+  );
+
+  const onRequestComplete = useCallback(() => {
+    // We only care if this happened from a trigger on this screen,
+    // not if we enter this screen on this state. Hack to achieve this, not
+    // super robust against parallel events, but good for this screen's testing purposes.
+    if (didMountRef.current) {
```
johnschoeman, line 35: ah, i see the reason for the ref here.  does this feature want to live in a parent component then?  Whats the need to be using memoization at all? could we get away with the same functionality without introducing a ref/callback hook if we dont have these side effects coupled to the screen components render cycle?

app/views/Partners/PartnersCustomUrlScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import { useDispatch, useSelector } from 'react-redux';
+import getHealthcareAuthoritiesAction from '../../store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction';
+import { Typography } from '../../components/Typography';
+
+import { Button } from '../../components/Button';
+import { Colors } from '../../styles';
+
+type PartnersEditScreenProps = {
+  navigation: NavigationScreenProp<NavigationState, NavigationParams>;
```
johnschoeman:
```suggestion
import { Screens, NavigationProp } from "../../navigation"

type PartnersEditScreenProps = {
  navigation: NavigationProp
```

app/store/reducers/healthcareAuthoritiesReducer.ts
```diff
@@ -36,10 +38,28 @@ const healthcareAuthoritiesReducer = createReducer(initialState, (builder) =>
...
+        // using a testing url for the yaml instead of the environment one
+        if (usesCustomUrl) {
+          state.availableCustomAuthorities = healthcareAuthorities;
+        } else {
+          state.availableAuthorities = healthcareAuthorities;
+        }
+
+        // Ensure that we don't subscribe to non-existent HAs, if the yaml updates
+        const validIds = new Set();
+        state.availableAuthorities.forEach(({ internal_id }) => {
```
johnschoeman, line 54:
```suggestion
        state.availableAuthorities.forEach(({ internalId }) => {
```
whats the reason for mixing cases here?

app/views/Partners/PartnersCustomUrlScreen.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const { t } = useTranslation();
+  const [url, setUrl] = useState('');
+  const didMountRef = useRef(false);
+
+  const dispatch = useDispatch();
+
+  // This is testing code only,we omit selector:
+  const fetchCustomStatus = useSelector(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (state: any) => state.healthcareAuthorities.request.status,
```
johnschoeman: bit of a code smell here? should we be maintaining an AppState type?

johnschoeman (approved):
Nice work! a few questions/suggestions.

vlasac (approved):

app/views/Partners/PartnersCustomUrlScreen.tsx
JacobJaffe: The goal here is achieve a `useEffect_exceptOnFirstMount` behavior. It needs to be at the same level as the redux selector logic. 
An alternative approach would be a `usePrev` state hook for mocking a more "componentDidUpdate" lifecycle, or building a explicit `useEffect_exceptOnFirstMount` hook. 

I went with this for the most lightweight solution here, since this screen is just for testing

app/store/reducers/healthcareAuthoritiesReducer.ts
JacobJaffe: This is part of the `HealthAuthority` data schema. The incoming data is all snake case, and the `internal_id` is set as part of the API layer, so from a consumer standpoint matches the payload. 
this is work revisiting for creating an API schema vs. app schema data validation / transformation step

app/views/Partners/PartnersCustomUrlScreen.tsx
JacobJaffe: we have a `RootState` type that the individual selectors use, I can add that here

app/views/Partners/PartnersCustomUrlScreen.tsx
johnschoeman: got it yeah, i wouldnt spend too much time on it, but it is quite a complicated thing which jumps out to me as a place for confusion/bugs.

PR #1030, opened 2020-06-16
Make health authority name generic

johnschoeman:
Why:
Currently we have `The Boston Health Department` hardcoded on the next
steps screen. Eventually we would like this to say the appropriate
health authority, but until we have that worked out, we will prefer to
use a generic placeholder for the Health authority name.

### DIFF
--- a/app/views/NextSteps/index.tsx
+++ b/app/views/NextSteps/index.tsx
@@ -16,8 +16,9 @@ const NextSteps = ({ navigation }: NextStepsProps): JSX.Element => {
     navigation.goBack();
   };
 
-  const headerText =
-    'The Boston Public Health Department recommends you take a self-assessment';
+  const healthAuthority = 'The PathCheck Health Department';
+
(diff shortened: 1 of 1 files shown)


tparesi (approved):
LGTM.

On the Export screens, we add the HA name from the dummy HA that is used. Although this is generic currently on BT and the name is used only for GPS, so they are not a big deal for any BT demo purposes. Just wanted to let you know.

johnschoeman:
> LGTM.
> 
> On the Export screens, we add the HA name from the dummy HA that is used. Although this is generic currently on BT and the name is used only for GPS, so they are not a big deal for any BT demo purposes. Just wanted to let you know.

I was torn in introducing some stuff to be able to share this between screens. I think it makes sense to have a `HealthAuthority` type that includes the name which we could share via a context maybe? I dont have enough context on whats going on there and Im fairly certain that @JacobJaffe is actively working on HA stuff.  Jacob, do you have any insight/thoughts on this?

PR #1031, opened 2020-06-16
Fix typography regressions

johnschoeman:
Why:
A previous commit updated the Typography component such that the styles
were handled by stylesheets, rather than react component logic, but the
onboarding screens were not updated to use the correct styles so some
visual regressions were introduced.

This commit:
Fixes the regressions by using the common styles.

### Before

onboarding-typography-before

### After

onboarding-typography-after

### DIFF
--- a/app/views/onboarding/EnableExposureNotifications.tsx
+++ b/app/views/onboarding/EnableExposureNotifications.tsx
@@ -1,17 +1,27 @@
 import React, { useContext } from 'react';
+import {
+  TouchableOpacity,
+  ImageBackground,
+  StatusBar,
+  StyleSheet,
+  View,
+} from 'react-native';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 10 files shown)


mattThousand (approved):

app/views/onboarding/EnableExposureNotifications.tsx
```diff
@@ -58,17 +72,23 @@ export const EnableExposureNotifications = (): JSX.Element => {
           </View>
 
           <View style={styles.footerContainer}>
-            <Button
```
tparesi, line 61: Does removing this allow you to remove the Theme above?

tparesi (approved):
LGTM - thanks

app/views/onboarding/EnableExposureNotifications.tsx
johnschoeman: probably! will leave for another commit though.

PR #1032, opened 2020-06-16
Use redux for selected HAs in intersection logic 

vlasac:
#### Description:
This PR updates the intersection logic to load the list of selected health authorities from redux instead of fetching the full list of available HAs on every intersection check cycle. 

#### How to test:
- Select one of available HAs from the list
- Home screen should perform the intersection check on the data from the newly selected HA


JacobJaffe (approved):
LGTM

JacobJaffe:
You might need to mock the store for the unit tests

PR #1034, opened 2020-06-16
Simulate/Reset Exposures

mattThousand:
### Why
We would like to show real exposure data from the `Exposure Notifications` API to the user.

### This Commit
This commit creates an event emitter that updates the JS layer when exposures are added or removed in the native iOS layer.

IMG_1841

Screen Shot 2020-06-16 at 1 49 23 PM


app/ExposureHistoryContext.tsx
```diff
@@ -94,6 +95,25 @@ const ExposureHistoryProvider = ({
   const toggleHasExposure = () => {
     setHasBeenExposed(!hasBeenExposed);
   };
+  const eventEmitter = new NativeEventEmitter(
+    NativeModules.ExposureEventEmitter,
+  );
+
+  useEffect(() => {
+    NativeModules.ExposureEventEmitter.startListening();
```
johnschoeman:
```suggestion
    const unsubscribe = NativeModules.ExposureEventEmitter.startListening();
    
    return unsubscribe
```

johnschoeman (approved):
:SHIPEITE:

app/exposureNotificationsNativeModule.ts
```diff
@@ -125,20 +126,23 @@ export const resetExposures = async (
...
   cb: exposureEventCallback,
-): EventSubscription => {
-  return ExposureEvents.addListener(
-    'EXPOSURES_CHANGED',
-    (jsonString: string) => {
-      const exposures: Possible[] = JSON.parse(jsonString);
-      cb(exposures);
-    },
-  );
+): EventSubscription | null => {
```
mattThousand, line 133: @johnschoeman this feels somewhat grotesque, but allocating `ExposureEvents` in the root scope caused an invariant error when running `GPS`. Would welcome an alternative suggestion, I'm not sure how idiomatic optionals are in `typescript`

johnschoeman: If you have a Maybe in your domain, then you have a Maybe in your domain. The best practice of trying to reduce and move Maybes to the edges of your system still applies, but ts doesnt have super nice of ergonomics around Maybe types so you do see a lot of the `MyThing | null` syntax that you have here.

johnschoeman: that said i would try to pull the conditional up as much as possible in the flow.

PR #1035, opened 2020-06-16
Instead of turning the date object into a string, use toDateString

tparesi:
#### Description:

The current implementation led to a flakey test that failed when the seconds were off by 1. Using the toDateString function removes the timestamp, which is not needed in this particular
instance (date only).


JacobJaffe (approved):
Huzzah! thanks for looking into this

johnschoeman (approved):
💯

PR #1037, opened 2020-06-16
Update copy for BT Tracing Off

tparesi:
#### Description:

Small copy change for the BT exposure notifications off.
Also, added in a change to hide the Select HA indicator for BT only.

#### Screenshots:

Screen Shot 2020-06-16 at 2 06 59 PM


johnschoeman (approved):
🚢

PR #1039, opened 2020-06-16
Tp temporary export code solution

tparesi:
#### Description:

1. Temporarily hard code an export code for the BT app
2. Fix close navigation on BT app Export Publish Consent


johnschoeman (approved):
:ship:

PR #1040, opened 2020-06-17
Build exposure history from raw exposures

johnschoeman:
Why:
We would like to be creating an exposure history for the user based off
of data received from the Exposure Notification Server data which is
passed to the JavaScript layer from the native layer.

This commit:
Introduces the logic for converting a list of raw Exposures sent from
the native side to an ExposureHistory. Additionally we refactored the
NativeModule a bit and exposed the debug screen to the production build
so that users can test in test flight.

Currently we are not receiving data from the GAEN server, but we do have
mocked data coming from the native layer. A user can tap 'simulate
exposure' on the debug screen to create a new simulated exposure.

#### Gif

faked-exposures

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -1,82 +1,21 @@
 import React, { createContext, useState, useEffect } from 'react';
-import dayjs from 'dayjs';
 
-import { startListening } from './exposureNotificationsNativeModule';
 import { isGPS } from './COVIDSafePathsConfig';
-
-type Posix = number;
-
-export interface Possible {
-  kind: 'Possible';
-  id: string;
-  date: Posix;
-  possibleExposureTimeInMin: number;
-  currentDailyReports: number;
-}
-
-export interface NoKnown {
-  kind: 'NoKnown';
-  id: string;
-  date: Posix;
-}
-
-export interface Unknown {
-  kind: 'Unknown';
-  id: string;
-  date: Posix;
-}
-
-export type ExposureDatum = Possible | NoKnown | Unknown;
-
-export type ExposureHistory = ExposureDatum[];
+import { BTNativeModule } from './bt';
+import { ExposureHistory, blankHistory } from './exposureHistory';
 
 interface ExposureHistoryState {
   exposureHistory: ExposureHistory;
   hasBeenExposed: boolean;
   userHasNewExposure: boolean;
-  toggleHasExposure: () => void;
   observeExposures: () => void;
   resetExposures: () => void;
 }
 
-const now: Posix = Date.now();
-
-const daysAgo = [...Array(21)].map((_v, idx: number) => {
-  return 20 - idx;
-});
-
-const fakeExposureHistory: ExposureHistory = daysAgo.map(
-  (daysAgo: number): ExposureDatum => {
-    const date = dayjs(now).subtract(daysAgo, 'day').valueOf();
-    if (daysAgo > 10) {
-      return {
-        kind: 'Unknown',
-        id: daysAgo.toString(),
-        date,
-      };
-    } else if (daysAgo === 3) {
-      return {
-        kind: 'Possible',
-        id: daysAgo.toString(),
-        date,
-        possibleExposureTimeInMin: 124,
-        currentDailyReports: 1,
-      };
-    } else {
-      return {
-        kind: 'NoKnown',
-        id: daysAgo.toString(),
-        date,
-      };
-    }
-  },
-);
-
 const initialState = {
-  exposureHistory: fakeExposureHistory,
+  exposureHistory: [],
   hasBeenExposed: false,
   userHasNewExposure: true,
-  toggleHasExposure: () => {},
   observeExposures: () => {},
   resetExposures: () => {},
 };
@@ -92,17 +31,18 @@ interface ExposureHistoryProps {
 const ExposureHistoryProvider = ({
   children,
 }: ExposureHistoryProps): JSX.Element => {
-  const [hasBeenExposed, setHasBeenExposed] = useState<boolean>(false);
+  const [exposureHistory, setExposureHistory] = useState<ExposureHistory>(
+    blankHistory(),
+  );
   const [userHasNewExposure, setUserHasNewExposure] = useState<boolean>(true);
-  const toggleHasExposure = () => {
-    setHasBeenExposed(!hasBeenExposed);
-  };
 
   useEffect(() => {
     if (!isGPS) {
-      const subscription = startListening((exposures: Possible[]) => {
-        console.log(exposures);
-      });
+      const subscription = BTNativeModule.subscribeToExposureEvents(
+        (exposureHistory: ExposureHistory) => {
+          setExposureHistory(exposureHistory);
+        },
+      );
 
       return () => {
         subscription?.remove();
@@ -119,13 +59,14 @@ const ExposureHistoryProvider = ({
   const resetExposures = () => {
     setUserHasNewExposure(true);
   };
+
+  const hasBeenExposed = false;
   return (
     <ExposureHistoryContext.Provider
       value={{
-        exposureHistory: fakeExposureHistory,
+        exposureHistory,
         hasBeenExposed,
         userHasNewExposure,
-        toggleHasExposure,
         observeExposures,
         resetExposures,
       }}>
--- a/app/ExposureNotificationContext.tsx
+++ b/app/ExposureNotificationContext.tsx
@@ -1,6 +1,6 @@
 import React, { createContext, useState } from 'react';
 
-import * as ExposureNotifications from './exposureNotificationsNativeModule';
+import { BTNativeModule } from './bt';
 
 export type ENAuthorizationStatus = 'authorized' | 'notAuthorized';
 
@@ -35,7 +35,7 @@ const ExposureNotificationsProvider = ({
     const cb = (authorizationStatus: ENAuthorizationStatus) => {
       setAuthorizationStatus(authorizationStatus);
     };
-    ExposureNotifications.requestAuthorization(cb);
+    BTNativeModule.requestAuthorization(cb);
   };
 
   return (
--- a/app/bt/exposureNotifications.spec.ts
+++ b/app/bt/exposureNotifications.spec.ts
@@ -1,0 +1,98 @@
+import dayjs from 'dayjs';
+
+import { DateTimeUtils } from '../helpers';
+import { Possible, ExposureDatum } from '../exposureHistory';
+import { toExposureHistory, RawExposure } from './exposureNotifications';
+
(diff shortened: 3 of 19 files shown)


app/ExposureHistoryContext.tsx
```diff
@@ -92,17 +31,18 @@ interface ExposureHistoryProps {
...
-  const toggleHasExposure = () => {
-    setHasBeenExposed(!hasBeenExposed);
-  };
 
   useEffect(() => {
     if (!isGPS) {
-      const subscription = startListening((exposures: Possible[]) => {
-        console.log(exposures);
-      });
+      const subscription = BTNativeModule.subscribeToExposureEvents(
```
johnschoeman, line 41: Idk if it makes sense to do something similar for the GPS side of things, but if so we could just pass this subscription handler in on the mounting of the component and pull out the `!isGPS`

app/bt/exposureNotifications.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      };
+
+      const result = toExposureHistory(rawExposures);
+
+      expect(result[result.length - 3]).toEqual(expected);
+    });
+  });
+
+  describe('when there are multiple raw exposures on the same day', () => {
+    it('combines the raw exposure into a single possible exposure with the sum of duration and max of risk', () => {
```
johnschoeman, line 56: The current strategy is to just add all the total durations and take the max of riskLevels, a future iteration will/should do something more sophisticated at this point to determine the total risk level for the day.

app/exposureHistory.ts
```diff
@@ -0,0 +1,37 @@
+import dayjs from 'dayjs';
+
+export type Posix = number;
+
+export interface Possible {
```
johnschoeman, line 5: idk if the GPS side will have the same data requirements for an `ExposureDatum` but if not we could introduce a generic over the history to account for this.  something like:
```
export type ExposureHistory<T> =T[];
```

app/views/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -108,28 +96,6 @@ const NoKnownExposureDetail = ({
   );
 };
 
-interface UnknownExposureDetailProps {
```
johnschoeman, line 111: It turns out that we dont get info back to determine if the user was or was not tracking exposure notifications on a day. So we cant make the determination that we have an Unknown ExposureDatum.  We just have, Possible and NoKnown.

ios/BT/bridge/ExposureEventEmitter.m
```diff
@@ -2,13 +2,18 @@
...
 // Notification/Event Names
-NSString *const onExposuresChanged = @"EXPOSURES_CHANGED";
+NSString *const onExposuresChanged = @"onExposureRecordUpdated";
 
 @interface ExposureEventEmitter : RCTEventEmitter <RCTBridgeModule>
 @end
 
 @implementation ExposureEventEmitter
 
++ (BOOL)requiresMainQueueSetup
```
johnschoeman, line 12: @mattThousand , wanted to point this out in particular, is this okay to be running on the main thread?

tparesi (approved):
This looks and works good.

app/ExposureHistoryContext.tsx
```diff
@@ -92,17 +31,18 @@ interface ExposureHistoryProps {
...
+        (exposureHistory: ExposureHistory) => {
+          setExposureHistory(exposureHistory);
```
JacobJaffe, line 43: Could you clarify what the intended abstraction is supposed to be for a subscriber here? Would the subscriber be responsible for keeping track of all exposure history, and sending the full history, vs. emitting a new history case? 

I believe it's the first one, in which case reconciliation of history is delegated to the service?

app/ExposureNotificationContext.tsx
```diff
@@ -35,7 +35,7 @@ const ExposureNotificationsProvider = ({
     const cb = (authorizationStatus: ENAuthorizationStatus) => {
       setAuthorizationStatus(authorizationStatus);
     };
-    ExposureNotifications.requestAuthorization(cb);
+    BTNativeModule.requestAuthorization(cb);
```
JacobJaffe, line 38: This looks breaking for GPS? Should this be a switch on the flavor?

app/exposureHistory.ts
```diff
@@ -0,0 +1,37 @@
...
+  date: Posix;
+}
+
+export type ExposureDatum = Possible | NoKnown;
+
+export type ExposureHistory = ExposureDatum[];
+
+export const blankHistory = (): NoKnown[] => {
+  const now = Date.now();
+  const daysAgo = [...Array(21)].map((_v, idx: number) => {
```
JacobJaffe: can we use some defined variables for our retention length?
```js
const RETENTION_DAYS = 21
```

app/helpers/dateTimeUtils.ts
```diff
@@ -3,7 +3,10 @@ import dayjs from 'dayjs';
 type Posix = number;
 
 export const isToday = (date: Posix): boolean => {
-  const beginningOfDay = dayjs(Date.now()).startOf('day').valueOf();
   const endOfDay = dayjs(Date.now()).endOf('day').valueOf();
-  return beginningOfDay <= date && endOfDay >= date;
+  return beginningOfDay(date) <= date && endOfDay >= date;
```
JacobJaffe, line 7: 🎉 Thanks for splitting up these helpers

app/views/ExposureHistory/ExposureDatumDetail.tsx
JacobJaffe: > we dont get info back to determine if the user was or was not tracking exposure notifications on a day

My understanding is this is inferred from being in an app state of not being active? e.g. on GPS if location services were off, or if locations are missing for a day, that's different than NoKnown

app/exposureHistory.ts
JacobJaffe: I think it will need to end up being a little different, but that's not well defined right now. From what I see, that would need only a corresponding change at the exposureHistory details component level?

johnschoeman: yup totally.  which given that that component is a leaf node i dont see any risks there.

app/views/ExposureHistory/ExposureDatumDetail.tsx
johnschoeman: yeah for GPS, it might make sense to have this concept, so we could leave this in, if you think itll be useful, but for BT we simply dont have that data. All we have is if there was an exposure or not.  At least at this point we may have a way to get that info down the line.

app/ExposureNotificationContext.tsx
johnschoeman: so long as the function isnt called for the gps build it wont throw, as RN handles the subscription across the bridge at run time and we are instantiating the emitter on invocation of the function.  

Though perhaps thats not super clear by reading the function itself, so we could put a redundant !isGPS in here, but id really like to get away from all that checking and it would be ideal to have each tracing strategy not need to be aware of each of the others.

app/exposureHistory.ts
johnschoeman: great idea!

app/ExposureHistoryContext.tsx
johnschoeman: > I believe it's the first one, in which case reconciliation of history is delegated to the service?

Thats how it is for the BT side, the native layer sends whole data set of exposures on each update. which is great as it means we dont have to keep track of anything, we just need to present the data to the user. 

---
 if we can keep the signature for all of the tracing strategies to something like:

```typescript
(callback: (exposureHistory: ExposureHistory) => void) => { remove: () => void }
```

(or perhaps

```typescript
(callback: (exposureHistory: ExposureHistory) => void) => () => void

// const unsubscribe = subscribeToGPSUpdates((history) => setHistory(history))
// return unsubscribe
``` 
which i believe is the pattern that redux uses under the hood.)

Then the ExposureHistoryContext really doesnt need to care if the data is coming from the gps side or the bt side or the whatever-side

So to make this idea work, i suppose the gps service would process the whole history and pass that in every time the history needs to update.

And whether or not this makes sense,  really depends on how the data is stored and passed along on in the gps flow, if its more of a "heres a new data point, please append onto the end of the list" , then we have a pretty different situation and would need to think through how to reconcile the differences if we want to be able to keep this context agnostic to the tracing strategy.

My guess is that it would similar to the BT pattern though, given the nature of task, we arent doing real time stuff at this level. We are listening for updates from the HA, and when we get one we check the list of data against the locally stored date gps data, if we get any hits we pass it along in a batch. But i dont have too much context on that side so I'd defer to you on if this a viable approach.

app/ExposureNotificationContext.tsx
JacobJaffe: Ah, I see. If GPS also needs a step in the flow here, it might make sense to have this decide the request that's made, so that the provided function is only the relevant one. I'm a little wary of providing a function to components that they shouldn't call. 
This sounds good for now then as we don't know on the GPS side what that would look like

app/views/ExposureHistory/ExposureDatumDetail.tsx
JacobJaffe: Do you know where this copy originally came from? If this was just added specifically for BT I'm good with removing it

side note: the strings in this file should be localized? out of scope for this PR but in general

johnschoeman: I initially put it in as it was in the figma, I did chat with Emma about it a bit but we werent sure. Ali would have more context on the why. My guess is that it was assumed to be an easy thing to do, and now that we've worked through it, its actually a bigger ask, so it would be worth revisiting if its a useful enough feature to the user. Either way, i think it belongs in a separate commit.

app/ExposureNotificationContext.tsx
johnschoeman: > I'm a little wary of providing a function to components that they shouldn't call.

same here! I think we could take a pass at make this a little harder to get at for a random component on the tree.

like if we case on the app mounting and conditionally include these there. it wouldnt be fool proof but it would be pretty easy to implement and youd have to go out of your way to pull this in.

app/views/ExposureHistory/ExposureDatumDetail.tsx
johnschoeman: > side note: the strings in this file should be localized? out of scope for this PR but in general

yes totally, was skipped for sake of the demo, we cut a card to capture the work.

ios/BT/bridge/ExposureEventEmitter.m
johnschoeman: chatted irl, theres no need fo this to be on the main thread since its not updating ui, but it is fine to keep on as well.

PR #1043, opened 2020-06-17
Hotfix: Migrate Redux Store for HA Custom URL

JacobJaffe:
Adds a migration step to add an initial empty array for the custom urls. This should fix any issues of a migrated app crashing on the URL entry screen

Also converts the migrations from js to ts, with a declarations file.
Typing migrations correctly is inherently problematic, as it requires storing previous state shapes, which gets very messy. Conditional safe edits and type assertion on the declaration is a better move imo. If this causes issues with improper state shapes in the future, we can look into migration unit tests.

We were previously not running any transformative migrations, which is why this is coming up now.

### DIFF
--- a/app/store/migrations/migrations.js
+++ b/app/store/migrations/migrations.js
@@ -1,0 +1,13 @@
+// Intentionally in JS to allow ensuring we can directly edit state between migrations.
+const migrations = {
+  0: (prevState) => prevState,
+  // Migration 1: We introduce the "availableCustomAuthorities" property to the "healthcareAuthorities" slice
+  1: (prevState) => {
+    if (prevState && prevState.healthcareAuthorities) {
+      prevState.healthcareAuthorities.availableCustomAuthorities = [];
+    }
+    return prevState;
+  },
+};
+
+export default migrations;
--- a/app/store/index.ts
+++ b/app/store/index.ts
@@ -12,7 +12,7 @@ const enhancers = composeWithDevTools(
   applyMiddleware(thunk, createImmutableStateInvariantMiddleware()),
 );
 
-export const STORE_VERSION = 0;
+export const STORE_VERSION = 1;
 const persistConfig = {
   key: 'root',
   storage: AsyncStorage,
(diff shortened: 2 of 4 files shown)


app/store/migrations/migrations.js
```diff
@@ -0,0 +1,13 @@
+// Intentionally in JS to allow ensuring we can directly edit state between migrations.
```
johnschoeman, line 1: not sure you need this comment, as itd be a bit odd to expect to also version the typesystem. That said, not sure its related to being able to mutate the state? ts doesnt preclude you from mutations?

johnschoeman (approved):
LGTM!

for context, why do we need to be handling migrations in js at all? we have some significant native experience on the team and realm is set up on the native layer i believe, is this something we can delegate to the ios/android side?

JacobJaffe:
@johnschoeman To clarify what's going on here -- this is a redux state migration. 
We persist redux state, for retention between app sessions (e.g. for on-boarding completed status).

On redux state shape changes, occasionally migrations are needed. Because they are migrating from different state shapes, they are really tricky to type because it requires holding a type from a previous version of the app (But we're constructing our type from reducer shapes).

vlasac (approved):

johnschoeman:
> @johnschoeman To clarify what's going on here -- this is a redux state migration.
> We persist redux state, for retention between app sessions (e.g. for on-boarding completed status).

Yup totally makes sense, my question is do we need to be doing this in the JS layer specifically? or could we save what we need to save on the native side.  Though, even if the answer is 'yes' I wouldnt take the trouble to change it up as this works, mostly just asking for the sake of understanding.

JacobJaffe:
> > @johnschoeman To clarify what's going on here -- this is a redux state migration.
> > We persist redux state, for retention between app sessions (e.g. for on-boarding completed status).
> 
> Yup totally makes sense, my question is do we need to be doing this in the JS layer specifically? or could we save what we need to save on the native side. Though, even if the answer is 'yes' I wouldnt take the trouble to change it up as this works, mostly just asking for the sake of understanding.

Gotchya. This is saved via RN async store. Theoretically we could boot this to the native side.
TBH I think the amount of persisted redux data we will need  in the app is super minimal (saved authorities, onboarding complete).
I'm leaning towards ripping out redux persist soon, and building our own super light weight persist layer (that'll play nicer from a security perspective too).

PR #1044, opened 2020-06-17
Changing and enabling update alert

Faustrox:
#### Screenshots: 
Screenshot_2020-06-17-16-46-12-652_com optic covdr


johnschoeman (changes requested):
Was this PR intentional?

It's significantly behind the latest development and changes the layout and design such that we cant include this without abandoning the existing product design. It is also too large to be reviewable by team in a meaningful or timely way.

Could you provide some context on your goals here?

PR #1045, opened 2020-06-18
Create EN Event Emitter on iOS, Update Debug Menu

mattThousand:
### Why
We'd like to manage exposure notification `enabled` state using an even emitter instead of querying it directly from the `JS` layer. This way, the UI can listen for updates to the native state, and update accordingly as needed.

### This Commit
This commit creates an event emitter in the iOS codebase that notifies the `JS` layer of any changes to the `Exposure Notifications` enabled/disabled state. It also updates the debug menu to bring it in line with the available native functionality (implementing the ability to toggle `EN` enabled/disabled state, removing the ability to simulate a positive test result because that concept does not exist in this app).

Note @johnschoeman @tparesi there is a distinction between `EN` "authorization" status and "enabled" state (per 🍎 ). The "authorization" status is for the `OS` (i.e. not per app), whereas the `enabled` state is per app. Thus, during onboarding, we request authorization and then enable `EN`, but thereafter only enable/disable. 

Jun-17-2020 21-27-42


app/ExposureNotificationContext.tsx
```diff
@@ -1,15 +1,16 @@
-import React, { createContext, useState } from 'react';
+import React, { createContext, useState, useEffect } from 'react';
 
+import { isGPS } from './COVIDSafePathsConfig';
 import { BTNativeModule } from './bt';
 
-export type ENAuthorizationStatus = 'authorized' | 'notAuthorized';
+export type ENAuthorizationStatus = `NOT_AUTHORIZED` | `AUTHORIZED`;
```
johnschoeman: what do you think about `AuthorizationStatus` here an drop the `EN`, we are in the ExposureNotificationContext, so it is perhaps redundant.

app/views/Settings/ENDebugMenu.tsx
```diff
@@ -80,8 +78,20 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
...
   const handleOnPressToggleExposureNotifications = () => {
-    handleOnPressSimulationButton(BTNativeModule.toggleExposureNotifications)();
-    global.ExposureNotificationsOn = !global.ExposureNotificationsOn;
+    handleOnPressSimulationButton(toggleExposureNotifications)();
+  };
+
+  const showDebugVerificationCode = () => {
+    Alert.alert(
+      'Debug Verification Code:',
+      '111111',
```
johnschoeman: is this `111111` actually hooked up to anything or is it dummy data? Either way does it want to live in a constant?

ios/BT/ExposureManager.swift
```diff
@@ -32,37 +32,41 @@ final class ExposureManager: NSObject {
...
-      } else if let error = error {
-        callback([error.localizedDescription])
-      } else {
-        callback([String.authorized])
+  @objc func requestExposureNotificationAuthorization(authorized: Bool, callback: @escaping RCTResponseSenderBlock) {
+    // Ensure exposure notifications are enabled if the app is authorized. The app
+    // could get into a state where it is authorized, but exposure
+    // notifications are not enabled,  if the user initially denied Exposure Notifications
+    // during onboarding, but then flipped on the "COVID-19 Exposure Notifications" switch
+    // in Settings.
```
johnschoeman, line 40: This explains the four states in the enum then?

ios/BT/ExposureManager.swift
```diff
@@ -250,18 +234,14 @@ final class ExposureManager: NSObject {
...
-          completion([NSNull(), "Exposure Configuration: \(configuration)"])
-        case .failure(let error):
-          completion([error.localizedDescription, NSNull()])
-        }
-      }
     case .resetExposures:
       BTSecureStorage.shared.exposures = List<Exposure>()
       completion([NSNull(), "Exposures: \(BTSecureStorage.shared.exposures)"])
+    case .toggleENAuthorization:
+      let authorized = manager.exposureNotificationEnabled ? false : true
```
johnschoeman:
```suggestion
      let nextAuthorizationStatus = manager.exposureNotificationEnabled ? false : true
```

johnschoeman (approved):
LGTM, some minor questions/suggestions

app/ExposureNotificationContext.tsx
mattThousand: Sure - and now that you mention it, this should be `EnabledStatus` (since `AuthorizationStatus` belongs to the OS)

mattThousand: `export type EnabledStatus = `DISABLED` | `ENABLED`;`

app/views/Settings/ENDebugMenu.tsx
mattThousand: It's not hooked up atm. But good thinking, it should be stored in a constant. Would putting a `const DEBUG_VERIFICATION_CODE = '111111';` at the top of the file be the idiomatic way to do this (or is it more idiomatic to store constants in a different centralized place)?

ios/BT/ExposureManager.swift
mattThousand:
```
public enum ENAuthorizationStatus : Int {

    
    /// Authorization status has not yet been determined.
    	/// This status means the user has not been prompted yet. Using the API in this state may prompt the user.
    case unknown = 0

    
    /// This app is not authorized to use Exposure Notification. The user cannot change this app's authorization status.
    	/// This status may be due to active restrictions, such as parental controls being in place.
    case restricted = 1

    
    /// The user denied authorization for this app.
    case notAuthorized = 2

    
    /// The user has authorized this app to use Exposure Notification.
    case authorized = 3
}
```
🧐

app/views/Settings/ENDebugMenu.tsx
johnschoeman: I would say yes.

constants that are configuration-ish in intent are usually are SCREAMING_SNAKE_CASE and are at the top of the file.

sometimes folk pull constants out to separate files, which i generally avoid as it makes it farther away from the thing that it is relevant to and is generally harder to maintain.

app/ExposureNotificationContext.tsx
johnschoeman: maybe even: `export type Status = DISABLED | ENABLED` but that might not be descriptive enough.

`export type Authorization = DISABLED | ENABLED`

is another one that might be good.

mattThousand: Hm, I'd steer clear of using "authorization", only because it has meaning in the `EN` domain that is distinct from "enabled". `Status` sounds good to me 👍

johnschoeman: Ah, I see the disconnect here. this is the `ExposureNotifications` Context so i see it as EN only. ideally wed never have any gps stuff here. OR are you thinking that this component will eventually become something generic over the tracing strategy?  if so should we change the name of the component?

mattThousand: ah - `EN` needs to be both "authorized" for the OS and "enabled" for the app; it's possible for it to be `authorized` but not `enabled` for the app. The user will see the confirmation dialog only if `EN` have not yet been "authorized" for the OS. From then on, the user must enable/disable `EN` for specific apps from the settings detail for the specific app. They can also "authorize"/"unauthorize" `EN` from global settings

johnschoeman: got it, so we have the 4 possible states as mentioned above.  As far as the type we want here, i could see it going either way depending on how we want to present this info to the user. If we want a specific message for each of the 4 cases it might make sense to have a type along the lines of

```typescript
type Authorization = Authorized | Unauthorized
type Enablement = Enabled | Disabled

type DeviceStatus = [Authorization, Enablement]
```

This fully captures the possible states and makes all others impossible. and makes the somewhat confusing situation mapped out for future readers.

and then in the view layer we'll have a pretty easy time pattern matching to show the correct message to the user.

johnschoeman: and the device status is what would be exposed to the view layer. not the authorization or enablement separately.

johnschoeman: or is it such that we cant be enabled if its unauthorized?

so we might want

```
type DeviceStatus = Enabled | Disabled | Unauthorized.
```

mattThousand: ^^ 💯

PR #1047, opened 2020-06-18
Add safe area view for navigation

johnschoeman:
Why:
Currently the navigation ui is not set up to respect safe area for large
format iPhones. Additionally the status bar barStyle does not update for
dark and light backgrounds declaratively. We would like for the
navigation screens to look consistent and professional on all phone
formats.

This commit:
Introduces a safe area view that is compatible with react navigation.
Given the JS > native approach that the react-navigation team takes, the
safe area view that is provided directly by react-native does not
integrate with the navigator update cycle as one happens on the native
side and the other on the js side.

A similar js implementation is provided for setting the status bar, its
unfortunately tied to a side effect, but it seems to be the only
available solution compatible with react-navigation at this point.

The various navigation screens were updated to use these new components
and side effect via a `useStatusBarEffect` custom hook. The navigation
bar wrapper was also updated to use the correct styles per the current
design.

## Before

safearea-8-before-gps

safearea-11-before-bt

## After

safearea-8-after-gps

safearea-11-after


app/navigation/index.ts
```diff
@@ -90,3 +95,15 @@ export const Stacks: { [key in Stack]: Stack } = {
   More: 'More',
   Partners: 'Partners',
 };
+
+type BarStyle = 'dark-content' | 'light-content';
+
+export const useStatusBarEffect = (barStyle: BarStyle): void => {
+  useFocusEffect(
```
johnschoeman, line 102: The fact that we have to do this makes me infinitely sad. If anyone is aware of a better solution I would love to hear it!

JacobJaffe: AFAIK _some_ kind of focus listener (at component or at navigator level) is required because of componentMounts in the navigation lifecycle, but this could also be as a declarative component on each screen if that's more readable (like how RN has both options, we can do that for a focus effect version)

```js
const ExampleScreen = () => {
...
return (
...
<FocusStatusBar barStyle={'light-content'}/>
...
)
/>
```

Where `FocusStatusBar` is simply a component with this hook

app/components/NavigationBarWrapper.tsx
```diff
@@ -1,31 +1,21 @@
...
- */
-const widthScale = Math.min(Dimensions.get('window').width / 400, 1.0);
+import { Icons } from '../assets';
+import { Spacing, Colors, Typography as TypographyStyles } from '../styles';
 
 interface NavigationBarWrapperProps {
   children: React.ReactNode;
   title: string;
-  onBackPress: () => void;
+  onBackPress?: () => void;
```
JacobJaffe, line 14: IMO this would be a good time to only have `onBackPress`, and remove `includeBackButton`, since this is now optional

App.js
```diff
@@ -21,6 +22,7 @@ export const UnconnectedApp = () => (
         <PermissionsProvider>
           <ExposureNotificationsProvider>
             <ExposureHistoryProvider>
+              <StatusBar barStyle={'dark-content'} />
```
JacobJaffe: Thoughts on this in the unconnected vs. the connected? also using the hook / the base status bar?
This is currently gated by the connection setup (slight delay on app mount for status bar being changed from splash screen)

JacobJaffe (approved):
🙏 Thank you for fixing this. A few suggestions

App.js
johnschoeman: makes sense to pull it out all together i think in that case as we are mostly overwriting the status style on a per screen basis.

app/navigation/index.ts
johnschoeman: Hmm, i see this as a style preference, the real pain point is the async complexity rather than being handled by a component directly, which i dont think you can get around for the reasons you mentioned. It just makes me a bit sad.

PR #1053, opened 2020-06-18
Download Diagnosis Keys from GAEN Server

mattThousand:
### Why
We'd like the iOS client to download diagnosis keys from the GAEN server in order to submit them to the `Exposure Notifications` library on device and determine whether the user has been exposed to COVID-19.

### This Commit
This commit updates the `ExposureManager` class to request an index file from the GAEN server containing a list of URLs for a given set of regions, and after parsing that file, download each diagnosis key.


ios/BT/API/APIClient.swift
```diff
@@ -1,24 +1,28 @@
...
 
-  let postUrl: URL
-  let pullUrl: URL
+  let postKeysUrl: URL
+  let downloadKeyFile: URL
+  let indexFileUrl: URL
   static let shared = APIClient(
-    postUrl: URL(string: "https://exposure-2kabcv6c4a-uc.a.run.app")!,
-    pullUrl: URL(string: "https://federationout-2kabcv6c4a-uc.a.run.app")!
+    postKeysUrl: URL(string: "https://exposure-2kabcv6c4a-uc.a.run.app")!,
```
johnschoeman: would these strings make sense in an env file or something?

ios/BT/API/APIClient.swift
```diff
@@ -81,14 +108,35 @@ private extension APIClient {
     static let errorMessage = "error_description"
   }
 
+  func downloadRequest<T: APIRequest>(for request: T, fileName: String) -> DownloadRequest {
```
johnschoeman, line 111: Not sure if this is a swift thing, but should there be 2 `downloadRequest` functions?

ios/Podfile.lock
```diff
@@ -199,7 +199,7 @@ PODS:
     - React
   - react-native-geolocation (2.0.2):
     - React
-  - react-native-safe-area-context (3.0.5):
```
johnschoeman: not sure this change is necessary to commit.

johnschoeman (approved):
LGTM! This is a crazy amount of work!

ios/Podfile.lock
mattThousand: ah - good catch, this one snuck in unintentionally

ios/BT/API/APIClient.swift
mattThousand: good question - yes, it's idiomatic (depending on who you ask) to have 2 functions with identical names but different signatures. The compiler (autocomplete) helps you out during development to ensure that you're calling the intended one

ios/BT/API/APIClient.swift
mattThousand: It would indeed. Will make a new commit.

ios/BT/API/APIClient.swift
johnschoeman: pretty neat! elixir has a similar feature and its awesome!

PR #1056, opened 2020-06-18
Remove Consistently Failing Tests

octohub:
As per our discussion in Slack, this PR comments out tests that are consistently failing on develop. We want to get to a place where we trust our testing strategy. That starts by removing tests that always fail on develop. Removing these will make a failed check on develop mean something.


johnschoeman (approved):
Thanks for grabbing this!  What do you think about deleting it altogether? itll still be available in the git history if/when we need to the code back.

octohub:
> Thanks for grabbing this! What do you think about deleting it altogether? itll still be available in the git history if/when we need to the code back.

Great call. Updating the branch now

.github/workflows/lokalise.yml
```diff
@@ -0,0 +1,24 @@
+name: i18n
```
johnschoeman, line 1: what drove pulling this one out to its own file?

octohub: As discussed offline, I want to avoid any skipped checks on PRs. Green checks only. As this check only runs when there is a merge to develop, I pulled it out into its own file so it would have its own `on` rules.

PR #1057, opened 2020-06-18
UI Regression fixes

JacobJaffe:
Various bug fixes, catching some regressions happened in the style standardizations. 
- A few components were still using themed typography, so this fixes those on the export screens. 
- Localizes the name of the settings screen
- adds the status bar hook to the main screen(s)
- adds a space in the about screen between copy & link (side note: we should consider using a transformative component here to accommodate Right - Left languages)
- Changes padding to margins where appropriate for the safe area view changes (RN safe area view has a slightly different approach to how it passes down margins & pad dings, resulting in some funkiness)
- Gives buttons a contrasting text for their violet version
- makes header back buttons white, removes back button settings & partners screens
- fixes underlay color on export selection of ha
- remove white line above tab bar(that was only there on contrasting dark screens)

Buttons now:
image
image


app/views/Partners/PartnersOverview.tsx
```diff
@@ -21,7 +21,6 @@ import { Typography } from '../../components/Typography';
 import Colors from '../../constants/colors';
 
 import { SvgXml } from 'react-native-svg';
-import colors from '../../constants/colors';
```
JacobJaffe, line 24: this had two imports of `colors` / `Colors` from same source

app/views/Export/ExportTemplate.js
```diff
@@ -102,11 +102,21 @@ export const ExportTemplate = ({
...
+                lightTheme && { color: Colors.primaryHeaderText },
+              ]}>
               {headline}
             </Typography>
             <View style={{ height: 8 }} />
-            <Typography style={styles.contentText}>{body}</Typography>
+            <Typography
+              style={[
+                styles.contentText,
+                lightTheme && { color: Colors.primaryText },
```
JacobJaffe, line 116: These `lightTheme &&` are gross, but this component needs refactoring overall now to accommodate the new typography system, so this is an immediate fix.

app/Entry.js
```diff
@@ -178,6 +178,7 @@ const MainAppTabs = () => {
           backgroundColor: Colors.navBar,
           paddingTop: Spacing.xSmall,
           height: Layout.navBar,
+          borderTopColor: Colors.navBar,
```
JacobJaffe, line 181: Removes the white line above the tab bar

app/views/About.js
```diff
@@ -55,6 +56,8 @@ export const AboutScreen = ({ navigation }) => {
         </View>
         <Typography style={styles.aboutSectionPara}>
           {t('label.about_para')}
+          {/* Space between the copy & link*/}
```
johnschoeman, line 59: not sure this comment is necessary, your code speaks for itself!

app/views/Export/ExportSelectHA.js
```diff
@@ -74,7 +75,7 @@ export const ExportSelectHA = ({ navigation }) => {
             keyExtractor={({ name }, i) => `${name}:${i}`}
             renderItem={({ item: HA }) => (
               <TouchableHighlight
-                underlayColor={Colors.moonRaker}
+                underlayColor={OldColors.UNDERLAY}
```
johnschoeman: what do you think about adding this color to the styles/colors file?

app/views/Partners/PartnersOverview.tsx
```diff
@@ -111,14 +113,14 @@ const styles = StyleSheet.create({
     paddingHorizontal: 24,
   },
   bottomSheet: {
-    backgroundColor: colors.WHITE,
+    backgroundColor: Colors.WHITE,
```
johnschoeman:
```suggestion
    backgroundColor: Colors.primaryBackground,
```
What do you think about using the design system based colors?  What happens when we reskin the app in the future? if we establish the pattern of named by intent rather than detail, we will save ourselves a lot of time.

app/views/Partners/PartnersOverview.tsx
```diff
@@ -111,14 +113,14 @@ const styles = StyleSheet.create({
...
     padding: 24,
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
     // TODO: Standardize bottom sheets. We mix shadows & borders.
     // Since this is on a main tab, borders are consistent.
     borderTopWidth: StyleSheet.hairlineWidth,
-    borderTopColor: colors.VIOLET_BUTTON_LIGHT,
+    borderTopColor: Colors.VIOLET_BUTTON_LIGHT,
```
johnschoeman:
```suggestion
    borderTopColor: Colors.primaryViolet,
```

johnschoeman (approved):
LGTM, thanks for grabbing these.

app/views/Partners/PartnersOverview.tsx
JacobJaffe: 💯  big agree

vlasac (approved):

PR #1058, opened 2020-06-18
Fix Logging Bug

octohub:
Fix logging bug associated with [SAF-718](https://pathcheck.atlassian.net/browse/SAF-718).


johnschoeman (approved):
🥇

PR #1059, opened 2020-06-18
Loading Indicator for Debug Screen

mattThousand:
### Why
We'd like to provide developers and testers with a visual indication that their intended action is being carried out on the native layer, so that they don't continue tapping the button and thus executing multiple actions unintentionally.

### This Commit
This commit displays an activity indicator when any button on the debug menu is pressed, and hides it when the relevant callback returns.

Jun-18-2020 19-19-38

### DIFF
--- a/app/views/Settings/ENDebugMenu.tsx
+++ b/app/views/Settings/ENDebugMenu.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   View,
   ViewStyle,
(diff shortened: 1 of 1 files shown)


app/views/Settings/ENDebugMenu.tsx
```diff
@@ -66,7 +68,13 @@ const ENDebugMenu = ({ navigation }: ENDebugMenuProps): JSX.Element => {
     ) => void,
   ) => {
     return () => {
+      // Update loading state
```
johnschoeman, line 71: These comments arent necessary imo, your code speaks for itself!

johnschoeman (approved):
LGTM!

PR #1060, opened 2020-06-18
Introduce more info screen

johnschoeman:
Why:
We would like to show the user a screen that provides more information
about the exposure history screen

This commit:
Introduces the more info screen and moves the NextSteps screen into the
exposure history feature folder.

more-info

### DIFF
--- a/app/views/ExposureHistory/MoreInfo.tsx
+++ b/app/views/ExposureHistory/MoreInfo.tsx
@@ -1,0 +1,66 @@
+import React from 'react';
+import { View, ScrollView, StyleSheet } from 'react-native';
+import { useNavigation } from '@react-navigation/native';
+import { useTranslation } from 'react-i18next';
+
+import { Typography } from '../../components/Typography';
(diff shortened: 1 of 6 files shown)


app/views/ExposureHistory/MoreInfo.tsx
```diff
@@ -0,0 +1,66 @@
...
+const MoreInfo = (): JSX.Element => {
+  const { t } = useTranslation();
+  const navigation = useNavigation();
+  useStatusBarEffect('light-content');
+
+  const handleOnBackPress = () => {
+    navigation.goBack();
+  };
+
+  const headerTextOne = t('exposure_history.why_did_i_get_an_en');
```
johnschoeman, line 21: @JacobJaffe , this is one where there might be some differences between gps and bt, what are your thoughts on this? is the copy generic enough or should we rope in product?

JacobJaffe (approved):
LGTM -- need a seperate followup on GPS specific copy

PR #1061, opened 2020-06-19
Show future days on calendar

johnschoeman:
Why:
Currently we would like to show days in the future on the calendar such
that the day labels line up with the correct day number.

This commit:
Introduces the logic to build the correct days in the ExposureHistory
list, that is, we determine when the next saturday is and pad out the
history such that the last day in the history is the up coming
staturday.

We separated out the concept of building an ExposureHistory as a list,
and building out the ExpousreHistory as a Map of days to exposures,
which we are calling an ExposureInfo. This will allow us flexibility in
presenting the exposure history in any calendar format a bit more easily
moving forward, for example other communities may not use the gregorian
calendar or we may discover that users would benefit from not seeing
this info in a calendar view at all.

This has a consequence to any tracing strategy that wants to be consumed
by the calendar view ui, as it is expected to now provide an
ExposureInfo rather than an ExposureHistory.

### Before

Screen Shot 2020-06-19 at 6 46 59 AM

### After

Screen Shot 2020-06-19 at 6 45 08 AM

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -2,7 +2,12 @@ import React, { createContext, useState, useEffect } from 'react';
 
 import { isGPS } from './COVIDSafePathsConfig';
 import { BTNativeModule } from './bt';
-import { ExposureHistory, blankHistory } from './exposureHistory';
+import {
+  ExposureInfo,
+  ExposureHistory,
+  calendarDays,
+  toExposureHistory,
+} from './exposureHistory';
 
 interface ExposureHistoryState {
   exposureHistory: ExposureHistory;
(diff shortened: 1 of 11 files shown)


app/exposureHistory.ts
```diff
@@ -17,23 +17,43 @@ export interface NoKnown {
...
+    } else {
       return {
         kind: 'NoKnown',
         date,
       };
+    }
+  });
+};
+
+export const calendarDays = (today: Posix, totalDays: number): Posix[] => {
```
johnschoeman, line 40: this one might want to live in another file as the concept is somewhat independent from the exposure history, though not sure if itll ever be used elsewhere so i think its fine to leave in here for now.

app/helpers/dateTimeUtils.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { isToday } from './dateTimeUtils';
+
+describe('isToday', () => {
```
johnschoeman, line 3: Noticed a bug in `isToday` which is why these specs where added. tbh im not sure how much value these specs in particular add, but it would be nice to encourage more specs to be written in general and this would be a place for folk to build on top of later

app/ExposureHistoryContext.tsx
```diff
@@ -28,18 +33,22 @@ interface ExposureHistoryProps {
   children: JSX.Element;
 }
 
+const blankHistory = toExposureHistory({}, calendarDays(Date.now(), 21));
```
JacobJaffe, line 36: Can we reference the 21 days as a shared constant?
AFAIK GPS is going 14 days, so this dynamic is useful

JacobJaffe (approved):

PR #1062, opened 2020-06-19
Fix selected language from onboarding in the Settings screen

vlasac:
#### Description:
This PR fixes the state initialisation for user selected locale in the Settings screen. 

Previously, the Settings screen would be in the correct language, but the menu for selecting the language would show that Englis is selected, and would prevent selecting English without previously selecting something else.

#### How to test:
Testing on Android and iOS:
- go through the onboarding process in any language different than English
- open the dashboard (Settings screen)
- the selected language should be the one that is selected in onboarding


johnschoeman (approved):
LGTM, thanks for grabbing this.

Is there a place for some test coverage around this functionality? That is, would it makes sense to add a spec to protect against this feature against regressions? If so, it probably would make sense to bring in in a future commit.

PR #1063, opened 2020-06-19
Migrate Realm, remove unused notification extension.

mattThousand:
### Why
We'd like the app not to crash due to a mismatched `Realm` schema

### This Commit
This commit migrates realm on iOS and also removes the `NextDiagnosisKeyFileIndexDidChange` extension on `Notification.Name`, which is no longer used


johnschoeman (approved):
Migration! :100:

PR #1067, opened 2020-06-19
Tp refactor en permissions

tparesi:
#### Description:

Connect the EN permissions on the Main page. 

Unfortunately, this does not seem to play well with the simulator, but does work on device.

Also, due to how the EN permissions works on the apple side, the onboarding flow does not show a dialogue asking for permissions. We are investigating this a bit more to figure out how this works.

### GIF

en-permissions


mattThousand (approved):

App.js
```diff
@@ -18,13 +19,19 @@ export const UnconnectedApp = () => (
   <FlagsProvider>
     <MenuProvider>
       <Theme use='default'>
-        <PermissionsProvider>
+        {isGPS ? (
```
johnschoeman: I like pulling the conditional up the tree a lot! though what do you think about passing the permission provider the tracing strategy itself? I think that help with future iteration were we introduce additional tracing strategies.

something like:
```
          <PermissionsProvider tracingStrategy={gps}>
            <ExposureHistoryProvider>
              <Entry />
            </ExposureHistoryProvider>
          </PermissionsProvider>
```

though this might be a good refactor for later.

app/ExposureNotificationContext.tsx
```diff
@@ -31,38 +32,38 @@ interface ExposureNotificationProviderProps {
...
-    const cb = (authorizationStatus: DeviceStatus) => {
-      setAuthorizationStatus(authorizationStatus);
+    const cb = (deviceStatus: DeviceStatus) => {
+      setDeviceStatus(deviceStatus);
     };
     BTNativeModule.requestAuthorization(cb);
   };
 
+  const getIsENAuthorizedAndEnabled = () => {
+    console.log(deviceStatus);
```
johnschoeman: extra console.log?

app/ExposureNotificationContext.tsx
```diff
@@ -31,38 +32,38 @@ interface ExposureNotificationProviderProps {
...
+    return deviceStatus[0] === 'AUTHORIZED' && deviceStatus[1] === 'ENABLED';
```
johnschoeman: i really wish js had pattern matching in elm you could say something like:
```elm
case deviceStatus of
  (Authorized, Enabled) -> true
  _ -> false
```

app/ExposureNotificationContext.tsx
```diff
@@ -31,38 +32,38 @@ interface ExposureNotificationProviderProps {
...
+  };
+
   return (
     <ExposureNotificationsContext.Provider
       value={{
-        authorizationStatus,
+        deviceStatus,
         requestENAuthorization,
+        getIsENAuthorizedAndEnabled,
```
johnschoeman:
```suggestion
        isENAuthorizedAndEnabled,
```

app/views/Main.js
```diff
@@ -1,52 +1,65 @@
...
+  );
+  const {
+    getIsENAuthorizedAndEnabled,
+    requestENAuthorization,
+    deviceStatus,
+  } = useContext(ExposureNotificationsContext);
+
+  useEffect(() => {
+    console.log('STATUS: ', deviceStatus);
+    setIsENAuthorizedAndEnabled(getIsENAuthorizedAndEnabled());
```
johnschoeman: this seems a little redundant to have the state in both the context and the component. could we keep it as a single source of truth and determine the `isEnabledAndAuthorized` on the fly?
something like:
```typescript
const isEnabled = getIsEnabledAndAuthorzed()

return isEnabled ? <AllServicesOnScreen /> : <AllServicesOnScreen />
```

ios/BT/ExposureManager.swift
```diff
@@ -5,242 +5,254 @@ import UserNotifications
...
-            // could get into a state where it is authorized, but exposure
-            // notifications are not enabled,  if the user initially denied Exposure Notifications
-            // during onboarding, but then flipped on the "COVID-19 Exposure Notifications" switch
-            // in Settings.
-            if ENManager.authorizationStatus == .authorized && !self.manager.exposureNotificationEnabled {
-                self.manager.setExposureNotificationEnabled(true) { _ in
-                    // No error handling for attempts to enable on launch
-                }
-            }
+
```
johnschoeman: are these changes from a rebase? or are did we also need to update the native layer for this functionality?

johnschoeman (approved):
LGTM! a few thoughts.

ios/BT/ExposureManager.swift
tparesi: We did need to change this, specifically requestExposureNotificationAuthorization.

I think all these changes had something to do with autoformatting. Right @mattThousand??

app/views/Main.js
tparesi: We tried to do this without local state, but the changes within the context were not triggering a re-render when navigating between tabs. I would love to have you look more closely at that to help me understand why.

app/ExposureNotificationContext.tsx
tparesi: Yeah - that is better. Thanks.

app/ExposureNotificationContext.tsx
tparesi: Me too - I didn't love this piece either. If we come up with something better, I am all ears

App.js
tparesi: I am going to keep this as is for now I think. I want to see how the exposure history provider works out too. But I think it will be an easy refactor in the future.

app/views/Main.js
johnschoeman: ah yeah, this has to do with the way react-navigation does things. basically screen components wont go through their update cycle in this situation.  Another way to get around this without introducing redundant local state and forcing a re-render with a `useEffect` is to pass the data in as a prop to a new component, so when the prop value updates, there will be re-render of the component.

so something like:
```typescript

const MyScreen = () => {
  const { myState } = useContext(MyContext)
  
  return (
    <View>
        <MyComponentThatIWantToUpdate myProp={myState} />
    </View>
}
```

so the parent screen component wont go through an update cycle on myState updating, but the child component `MyComponentThatIWantToUpdate` will.

Kinda annoying but it leads to some separation from the component that is responsible for screen transitions and the component that is responsible for displaying content on a screen, which is helpful i suppose.

app/ExposureNotificationContext.tsx
johnschoeman: totally fine for now! but it might makes sense to wait to the last moment to make the decision and use a switch statement for the various code paths.  so we would skip the need for a function that goes from DeviceStatus to boolean altogether.

it also suggests to me that we might have the wrong data model here.  if from the UI perspective we only have the two states of `CanGetENData` and `CannotGetENData`, we might want to provide a type to represent this directly

```typescript
type CanCollectENData = boolean
type ENStatus = CanCollectENData
```

and provide a way to convert from the native data representation of `type DeviceStatus = [Authorization, Enablement]` to `ENStatus`

with something like:

```typescript
const toENStatus = (deviceStatus: DeviceStatus): ENStatus => {
  ...
```

though this might be getting to the touch pedantic side and im not sure its necessary.

app/views/Main.js
johnschoeman: Its also possible that the context is not the correct fit for this use case, though i wouldnt be surprised if other components on the tree than this screen will need access to this.

app/bt/Home/Home.tsx
```diff
@@ -0,0 +1,94 @@
...
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: Spacing.medium,
+  },
+  contentContainer: {
+    position: 'absolute',
```
johnschoeman, line 69: This absolute positioning is a bit wonky, but is a consequence of the using an icon as the background image with a fixed position. A future commit might want to introduce a png, and or revisit the design altogether.

app/TracingStrategyContext.tsx
```diff
@@ -4,10 +4,16 @@ import { ExposureHistoryProvider } from './ExposureHistoryContext';
...
+  homeScreenComponent: () => JSX.Element;
 }
 
+const DefaultHomeScreenComponent = () => {
+  return <></>;
+};
+
 const defaultState = {
   name: 'initializingTracingStrategy',
+  homeScreenComponent: DefaultHomeScreenComponent,
```
johnschoeman, line 16: Not a fan of having the homescreen component on the tracing strategy. Ideally a future commit could separate the data required by the tracing strategy and the application view logic.

app/bt/Home/Home.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
```
tparesi, line 1: I am a bit worried about separating for the Home screens for GPS and BT. What are the ramifications around design inconsistencies and such?

app/views/Main.js
```diff
@@ -1,12 +1,11 @@
-import { useNavigation } from '@react-navigation/native';
```
tparesi, line 1: It might be confusing to have Main.js at the root of the views directory while it is only used for GPS.

app/views/Main.js
```diff
@@ -68,11 +57,16 @@ export const Main = () => {
...
   }, [navigation, updateStateInfo]);
 
+  const handleBackPress = () => {
+    BackHandler.exitApp(); // works best when the goBack is async
+    return true;
+  };
+
   if (!trackingInfo.canTrack) {
-    return <TracingOffScreen />;
+    return <TracingOffScreen onPress={openSettings} />;
```
tparesi: If this is only being used for GPS, we no longer need to pass an onPress to the TracingOffScreen. We can import this directly in that screen.

tparesi (commented):
I can't approve as it is my PR haha, but I added some comments.

app/bt/Home/Home.tsx
johnschoeman: totally agree, but its a bit hard to consolidate them quickly as the underlying data model is currently very strategy dependent and it would take a lot of effort to massage it out of the main screen.  I think it would great to move towards consolidating the screens over time though.

app/views/Main.js
johnschoeman: agree, though im concerned with merge conflicts on the gps team.  We can move it and see what happens i suppose.

tparesi:
@johnschoeman I say we merge it as is for now and keep it moving. That sound good? We will be working in this area a lot in the coming days and weeks, so we can tackle anything in future PR's

PR #1068, opened 2020-06-19
Release Screens Cleanup

JacobJaffe:
A lot of coupled styling changes, so a few different cosmetic updates are bundled here. 
Major changes:
- adds GPS flavors for the 'More info' screen for the calendar / exposure diary 
- Changes TouchableOpacities to TouchableHighlights where they were used on white background. This is a more accessible contrast on our listview buttons, and is now consistent across settings & ha selection pages.
- Changes the exposure diary stack to have a slide animation, for consistency on these flows. LMK if this was intentionally a modal fade style. The code this was based off of (export flow) used the modal / fade flow specifically for a 1 directional page transition flow / combining screens which are conceptually the _same_ screen states. Other stacks that allow going back in the app are using standard stack behaviors, so this is more consistent imo. 
- Standardizes About / Legal pages to our typography system, matches them to figma, and fixes linkable text to use onPress text, vs. wrapping in a toucahble. 
- Adds a specific GPS release state for the main page with no HAs selected, as no HAs will be available immediately. 

image
image
image
image
image

### DIFF
--- a/app/components/NavigationBarWrapper.tsx
+++ b/app/components/NavigationBarWrapper.tsx
@@ -12,7 +12,6 @@ interface NavigationBarWrapperProps {
   children: React.ReactNode;
   title: string;
   onBackPress?: () => void;
-  includeBottomNav?: boolean;
   includeBackButton?: boolean;
 }
 
--- a/app/components/NavigationBarWrapper.tsx
+++ b/app/components/NavigationBarWrapper.tsx
@@ -12,7 +12,6 @@ interface NavigationBarWrapperProps {
   children: React.ReactNode;
   title: string;
   onBackPress?: () => void;
-  includeBottomNav?: boolean;
   includeBackButton?: boolean;
 }
 
@@ -43,19 +42,19 @@ export const NavigationBarWrapper = ({
 
   return (
     <SafeAreaView style={styles.container}>
-      <View style={styles.headerContainer}>
-        <View style={styles.leftContent}>
-          {includeBackButton ? (
+      {includeBackButton ? (
+        <View style={styles.headerContainer}>
+          <View style={styles.leftContent}>
             <TouchableOpacity onPress={handleOnPressBack}>
               <SvgXml xml={Icons.BackArrow} color={Colors.white} />
             </TouchableOpacity>
-          ) : null}
+          </View>
+          <View style={styles.middleContent}>
+            <Typography style={styles.headerText}>{title}</Typography>
+          </View>
+          <View style={styles.rightContent} />
         </View>
-        <View style={styles.middleContent}>
-          <Typography style={styles.headerText}>{title}</Typography>
-        </View>
-        <View style={styles.rightContent} />
-      </View>
+      ) : null}
       <View style={styles.contentContainer}>{children}</View>
     </SafeAreaView>
   );
--- a/app/components/NavigationBarWrapper.tsx
+++ b/app/components/NavigationBarWrapper.tsx
@@ -12,6 +12,7 @@ interface NavigationBarWrapperProps {
   children: React.ReactNode;
   title: string;
   onBackPress?: () => void;
+  includeBottomNav?: boolean;
   includeBackButton?: boolean;
 }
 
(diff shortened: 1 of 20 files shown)


app/views/Settings/index.tsx
```diff
@@ -158,7 +168,8 @@ const SettingsScreen = ({ navigation }: SettingsScreenProps): JSX.Element => {
 
         {isGPS ? (
           <FeatureFlag name={'google_import'}>
-            <View style={styles.section}>
+            <View
+              style={[styles.section, { paddingHorizontal: Spacing.small }]}>
```
johnschoeman: why was this moved to an array? could we keep it in the styles.section?

app/views/Settings/index.tsx
```diff
@@ -169,6 +180,7 @@ const SettingsScreen = ({ navigation }: SettingsScreenProps): JSX.Element => {
             label={t('screen_titles.about')}
             onPress={navigateTo(Screens.About)}
           />
+          <Divider />
```
johnschoeman, line 184: what are your thoughts on not using `Divider` components? could we handle the layout using flex and padding or margin?

app/components/NavigationBarWrapper.tsx
```diff
@@ -43,19 +42,19 @@ export const NavigationBarWrapper = ({
 
   return (
     <SafeAreaView style={styles.container}>
-      <View style={styles.headerContainer}>
-        <View style={styles.leftContent}>
-          {includeBackButton ? (
+      {includeBackButton ? (
```
JacobJaffe: This is a mistake, undoing shortly

app/components/Typography.tsx
```diff
@@ -12,13 +12,15 @@ interface TypographyProps {
   style?: TextStyle;
   testID?: string;
   children: JSX.Element | string;
+  onPress?: () => void;
```
johnschoeman, line 15: What do you think about not adding this prop and using a touchable opacity at the component itself?  I worry about this component becoming gradually a kitchen drawer of a maintainability headache.

Additionally it would be really great to fully implement RTL language layout eventually and keeping this super lean will help with that.

app/views/About.js
```diff
@@ -125,38 +121,11 @@ export const AboutScreen = ({ navigation }) => {
...
-    marginTop: 14,
-    lineHeight: 32,
-  },
-  aboutSectionPara: {
-    color: Colors.primaryText,
-    fontSize: 16,
-    lineHeight: 22.5,
-    marginTop: 12,
-    fontFamily: fontFamily.primaryRegular,
+    paddingHorizontal: 24,
```
johnschoeman:
```suggestion
    paddingHorizontal: Spacing.medium,
```

Minor nitpick, but mostly to encourage future contributors to staying within the design system!

app/views/ExposureHistory/index.tsx
```diff
@@ -61,20 +61,16 @@ const ExposureHistoryScreen = (): JSX.Element => {
     selectedDatum && !DateTimeUtils.isInFuture(selectedDatum.date);
 
   return (
-    <SafeAreaView>
-      <ScrollView
-        style={styles.container}
-        contentInset={{ top: 0, bottom: 140 }}>
+    <SafeAreaView style={{ flex: 1 }}>
```
johnschoeman, line 64: Ah I get what you were explaining earlier now, I thought you wanted to put the flex on the scroll view directly.  This makes a lot of sense. 👍

app/views/Licenses.tsx
```diff
@@ -61,28 +60,35 @@ export const LicensesScreen = ({
...
-          <TouchableOpacity
-            onPress={handleOnPressOpenUrl('mailto:info@pathcheck.org')}>
-            <Typography style={styles.hyperlink}>{infoAddress}</Typography>
-          </TouchableOpacity>
-          <TouchableOpacity
-            onPress={handleOnPressOpenUrl('covidsafepaths.org')}>
-            <Typography style={styles.hyperlink}>{pathCheckAddress}</Typography>
-          </TouchableOpacity>
+          <Typography use='headline2'>{legalHeaderText}</Typography>
+          <View style={{ paddingTop: 10, paddingLeft: 24 }}>
```
johnschoeman:
```suggestion
          <View style={{ paddingTop: 12, paddingLeft: 24 }}>
```
or better yet

```suggestion
          <View style={{ paddingTop: Spacing.xSmall, paddingLeft: Spacing.medium }}>
```

In general its best to stick to multiples of 4 for px, and to try and limit the number of spacing values used as it leads to a less visually noisy look and feel.

johnschoeman (approved):
LGTM, some minor nitpicks.

app/views/Settings/index.tsx
JacobJaffe: This is specifically for drawing the dividing line (really a `SeperatorComponeent` in FlatList terminology I guess?)
We were previously doing this with implicit border colors, but that relies on the margins of the border being the same as intended divider, which isn't true here. 

IMO the real solution is that the `<Section>` component maps elements with diving lines between them, but that feels overkill considering this happens in just this 1 place right now.

app/components/Typography.tsx
JacobJaffe: The behavior is improved with using the direct onPress IMO -- it delegates to the react native `Text` link behavior, which gives us immediate wins on accessibility and this way the topographies are better set up to be composed as Text elements.

Maybe we can have it directly extend a Text component instead, from a props perspective?

I'm unfamiliar with RTL language layout -- what would that entail?

app/views/About.js
JacobJaffe: Good call, I've been lazy on these

app/views/Licenses.tsx
JacobJaffe: 💯

app/views/Settings/index.tsx
JacobJaffe: This was to correct behavior of the GoogleButton component. I updated this to just fix that component + wrappers

PR #1071, opened 2020-06-19
Ignore env Files

mattThousand:
### Why
We'd like to ensure that sensitive information is not exposed

### This Commit
This commit removes sensitive information from all `.env` files. Env information is now pulled in from a private submodule managed by the org. When setting up the repo, developers should now run:

```
$ git submodule update --init
$ sh env_setup.sh
```

The first command pulls in the submodule containing the populated .env files. The second copies them to the project root and ensures that changes to them are not tracked.


tparesi (approved):
Perfect - thanks

JacobJaffe (approved):
Might be good to document the need for this in the readme now so folks know how to set it up

mattThousand:
> Might be good to document the need for this in the readme now so folks know how to set it up

great point 👍

johnschoeman (approved):
My understanding is that .gitignore will keep things from being tracked, but since these where previously being tracked wont we need to also explicitly untrack the files?

It also would be nice to include a `.env.sample` file which is in the git history.

README.md
```diff
@@ -117,6 +117,15 @@ dev_setup.sh
 dev_setup.bat
```
johnschoeman, line 117: if we do add the `.env.sample`, it might be nice to include a step in the setup scripts that creates all of the `.env.*` for the developer.

mattThousand:
> My understanding is that .gitignore will keep things from being tracked, but since these where previously being tracked wont we need to also explicitly untrack the files?
> 
> It also would be nice to include a `.env.sample` file which is in the git history.

I *believe* that this commit will have the effect of keeping the existing files (as changed in this PR) in the repo, but will ignore changes going forward. I'm not sure that there is a way to get around the need for a `.env.gps` with a `gps` tracing strategy in order for the e2e tests to pass. I'd like to store a set of populated `.env` files in a private repo on the org, and then per your recommendation add a step to `dev_setup.sh` to pull them down. Very open to alternative approaches if any come to mind for you 🤔

johnschoeman:
> I _believe_ that this commit will have the effect of keeping the existing files (as changed in this PR) in the repo, but will ignore changes going forward. I'm not sure that there is a way to get around the need for a `.env.gps` with a `gps` tracing strategy in order for the e2e tests to pass. I'd like to store a set of populated `.env` files in a private repo on the org, and then per your recommendation add a step to `dev_setup.sh` to pull them down. Very open to alternative approaches if any come to mind for you 🤔

If we need info from the env files for CI, we should be storing those as secrets in the github org. I dont see a reason to do the extra work of pulling down files.

I was mostly thinking that during dev setup we would create the .env files for the dev locally as a convince. They would still need to ask another developer for the secret keys and stuff if they want to deploy.

.github/workflows/deploy.yml
```diff
@@ -11,22 +11,22 @@ on:
 
 jobs:
   version_bump:
-    name: "Bump version"
+    name: 'Bump version'
```
johnschoeman, line 14: Not sure we even need this action anymore? I dont think we are going to try to automate the version numbers at this point, given the multiplicity of apple orgs and build flavors.

README.md
```diff
@@ -117,6 +117,29 @@ dev_setup.sh
...
+Then, add your own values for each environment variable in the following 6 `.env` files:
+
+- .env.bt
+- .env.bt.staging
+- .env.bt.release
+- .env.gps
+- .env.gps.staging
+- .env.gps.release
+
+### For Safe Paths Developers
```
johnschoeman:
```suggestion
### For PathCheck Developers
```

PR #1072, opened 2020-06-19
HA List Empty State

JacobJaffe:
Empty States for HA List & Health Authorities -> Department copy updates.


app/components/NoAuthoritiesMessage.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
johnschoeman, line 1: This this is only being used in one place, what do you think about colocating the file with the PartnersEdit screen?

johnschoeman (approved):

app/components/NoAuthoritiesMessage.tsx
JacobJaffe: Two places -- Export Flow & Partners Edit

Maybe that HA List should have an abstracted component...

PR #1076, opened 2020-06-20
iOS Background Task for Exposure Detection

mattThousand:
### Why
We'd like `Exposure Detection` to be carried out at regular intervals in the background, so that the user remains current on their possible exposures.

### This Commit
This commit registers and schedules a background task to run on iOS that periodically executes exposure detection in the background. It also implements a local notification to encourage the user to enable bluetooth if exposure notifications have been authorized but bluetooth has not be enabled.

** Note: the `BT` app does not currently request notification permissions. This is something we should request for the `BT` app before ship in order for notifications to appear for the user

Jun-19-2020 17-18-20


ios/BT/ExposureManager.swift
```diff
@@ -2,245 +2,228 @@ import Foundation
...
+  
+  enum EnabledState: String {
+    case ENABLED, DISABLED
+  }
+  
+  enum AuthorizationState: String {
+    case AUTHORIZED, UNAUTHORIZED
+  }
+  
+  typealias DeviceState = (AuthorizationState, EnabledState)
```
johnschoeman, line 26: 💯

ios/BT/ExposureManager.swift
```diff
@@ -2,245 +2,228 @@ import Foundation
...
+    do {
+      try BGTaskScheduler.shared.submit(taskRequest)
+    } catch {
+      print("Unable to schedule background task: \(error)")
+    }
+  }
+
+  func notifyUserBlueToothOffIfNeeded() {
+    let identifier = String.bluetoothNotificationIdentifier
+    if ENManager.authorizationStatus == .authorized && manager.exposureNotificationStatus == .bluetoothOff {
```
johnschoeman, line 239: so does this mean that if the user has bluetooth off, but en permissions on we dont have en notifications coming through?  or does turning bluetooth off also turn en permission off automatically?

johnschoeman (approved):
LGTM! a question.

It was also a bit hard to distinguish between the diff that was due to the files being moved around and the actual new code.  If you think its useful, it might be worth pairing on this review!

johnschoeman:
and a card has been cut for the push notification work: https://trello.com/c/UyWrnN0E/66-as-a-user-when-i-am-onboarding-the-bt-app-i-am-asked-to-provide-push-notification-permissions

ios/BT/ExposureManager.swift
mattThousand: good question - the latter; I'll add a comment 👍

PR #1080, opened 2020-06-21
Download and unarchive key files (iOS only)

mattThousand:
### Why
We'd like to download and unarchive exposure keys from the `GAEN` server

### This Commit
This commit creates infrastructure to fetch archives from the `GAEN` server, and unarchive each one in order to submit them to the `Exposure Notifications` framework 

** Note: the `Exposure Notifications` framework will error out because the keys are currently unsigned. In order to complete the end-to-end flow, the `GAEN` server must return signed keys

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -79,18 +79,20 @@ final class ExposureManager: NSObject {
     }
     detectingExposures = true
     
-    var localURLs = [URL]()
-    
+    var localCompressedURLs = [URL]()
+    var localUncompressedURLs = [URL]()
+
     func finish(_ result: Result<[Exposure]>) {
-      for localURL in localURLs {
-        // Delete downloaded file from file system
-        do {
-          try FileManager.default.removeItem(at: localURL)
-        } catch {
-          completionHandler?(false)
-          return
-        }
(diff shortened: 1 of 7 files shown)


ios/BT/ExposureManager.swift
```diff
@@ -135,37 +137,40 @@ final class ExposureManager: NSObject {
...
                 return
               }
-              let newExposures = exposures!.map { exposure in
-                Exposure(id: UUID().uuidString,
-                         date: exposure.date.posixRepresentation,
-                         duration: exposure.duration,
-                         totalRiskScore: exposure.totalRiskScore,
-                         transmissionRiskLevel: exposure.transmissionRiskLevel)
+              let userExplanation = NSLocalizedString("USER_NOTIFICATION_EXPLANATION", comment: "User notification")
+              ExposureManager.shared.manager.getExposureInfo(summary: summary!, userExplanation: userExplanation) { exposures, error in
```
johnschoeman, line 160: we ended up introducing a concept of `ExposureInfo` on the client side as well, to name the map of date to aggregated exposures for a user.  What does ExposureInfo mean here? If it means something different here, should we update the name on the client side?

johnschoeman (approved):
LGTM, one question.

PR #1081, opened 2020-06-22
Introduce tracing strategy provider

johnschoeman:
Why:
Currently we do not have much protection against one version of the app
accessing functionality intended to only be used another version. For
example, in the BT build of the app, any component could request
location permissions, which would cause the app to error. Given that
situations like these could be grounds for violating compliance, we
would like to introduce a better strategy for providing tracing strategy
specific logic and functionality.

Additionally, the application has many conditional checks of the tracing
strategy scattered throughout the code base (`isGPS` and `!isGPS`). This
is prone to human error, is likely to be forgotten, and introduces
significant cognitive noise while reading the implementation of a
feature. Given that near term product goals include introducing more
tracing strategies, this is also prime for a combinatoric explosion of
complexity. We would like to reduce the number of sites in which we need
to be conditionally checking the tracing strategy, ideally to only one.

This commit:
Introduces the concept of a TracingStrategy as an interface, which can
be used as the common schema to which all tracing strategies can be
built against. And which UI application code can rely on to provide a
functional user interface for that tracing strategy.

We check the tracing strategy in the root App component and provide the
strategy specific functionality to the rest of the application via a
context. This removes the need of child components to need to check the
current strategy and makes it inconvenient and unlikely for a child
component to access functionality belonging to a different strategy.

Note that this commit does not remove every instance of isGPS from the
code base to reduce the size of this commit. Future commits should
include introducing the strategy specific content, assets, and routes to
the tracing strategy interface.

Finally by having a notion of a TracingStrategy in the type system it
becomes significantly easier to reason about the portions of the which
are different between the strategies and ideally we will be able to keep
the interface as small as possible to maximize the amount of common code
between the strategies.

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -1,7 +1,5 @@
 import React, { createContext, useState, useEffect } from 'react';
 
-import { isGPS } from './COVIDSafePathsConfig';
-import { BTNativeModule } from './bt';
 import {
   ExposureInfo,
   ExposureHistory,
@@ -29,14 +27,25 @@ const ExposureHistoryContext = createContext<ExposureHistoryState>(
   initialState,
 );
 
+export type ExposureInfoSubscription = (
+  cb: (exposureInfo: ExposureInfo) => void,
+) => { remove: () => void };
+
 interface ExposureHistoryProps {
   children: JSX.Element;
+  exposureInfoSubscription: ExposureInfoSubscription;
 }
 
-const blankHistory = toExposureHistory({}, calendarDays(Date.now(), 21));
+const CALENDAR_DAY_COUNT = 21;
+
+const blankHistory = toExposureHistory(
+  {},
+  calendarDays(Date.now(), CALENDAR_DAY_COUNT),
+);
 
 const ExposureHistoryProvider = ({
   children,
+  exposureInfoSubscription,
 }: ExposureHistoryProps): JSX.Element => {
   const [exposureHistory, setExposureHistory] = useState<ExposureHistory>(
     blankHistory,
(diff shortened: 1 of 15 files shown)


app/TracingStrategyContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface TracingStrategyProps {
+  children: JSX.Element;
+  strategy: TracingStrategy;
+}
+
+const TracingStrategyProvider = ({
+  children,
+  strategy,
+}: TracingStrategyProps): JSX.Element => {
+  const StrategyPermissionsProvider = strategy.permissionsProvider;
```
johnschoeman, line 26: Im less of a fan of passing in a whole react component here. Ideally we keep a better separation from the data model layer and the ui application layer.  I think we could instead provide a list of Permissions which provide was to interact with the device and provide that to a generic Permissions Provider.  It would make sense to do a pass at this in a future commit though.

app/TracingStrategyContext.spec.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useContext } from 'react';
+import { View, Text } from 'react-native';
+import { render, cleanup } from '@testing-library/react-native';
+import '@testing-library/jest-native/extend-expect';
```
johnschoeman, line 4: introduced jest-native as it makes working with react-testing-library __a lot__ easier in react-native land.

app/TracingStrategyContext.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { createContext } from 'react';
+import { TracingStrategy } from './tracingStrategy';
+import { ExposureHistoryProvider } from './ExposureHistoryContext';
+
+interface TracingStrategyContextState {
+  name: string;
```
johnschoeman, line 6: right now, name is acting as a bit of a place-holder. Future commits would like put the strategy specific copy and assets in here.

app/ExposureHistoryContext.tsx
```diff
@@ -29,37 +27,42 @@ const ExposureHistoryContext = createContext<ExposureHistoryState>(
...
+  cb: (exposureInfo: ExposureInfo) => void,
+) => { remove: () => void };
+
 interface ExposureHistoryProps {
   children: JSX.Element;
+  exposureInfoSubscription: ExposureInfoSubscription;
 }
 
-const blankHistory = toExposureHistory({}, calendarDays(Date.now(), 21));
+const CALENDAR_DAY_COUNT = 21;
```
johnschoeman, line 39: this change should have been included in a previous commit.

tparesi (approved):
LGTM

johnschoeman:
@JacobJaffe, gonna merge this in, as it's blocking some other prs that we want to ship, if you have any thoughts, we can revisit this stuff in a future pr/commit.

PR #1082, opened 2020-06-22
Include strategy specific assets in tracing strategy 

johnschoeman:
Why:
We would like to include all of the strategy specific content assets in
the tracingStrategy object provided to the app on app load. This will
allow us to not need to check the strategy for each individual piece of
copy or asset as we simply provide all of the strategy dependent content
in the root component.

This commit:
Removes the useAssets hook and replaces its usage with a
useStrategyContent hook which gets is data from a prop passed to the
Tracing Strategy provider on mounting of the root component.
Additionally we add the test factories library, fishery, to provide test
factories for the specs related to this behavior.

Note that a better strategy to handle the translated content may be to
have separation at the translation json files, rather than at the
javascript object level. A future commit may want to introduce this to
replace the `StrategyCopy` portion of this workflow.


app/TracingStrategyAssets.ts
```diff
@@ -1,214 +0,0 @@
...
-import { Icons, Images } from './assets';
-import { isGPS } from './COVIDSafePathsConfig';
-
-export type Asset = string | ((name: string) => string) | null;
-
-export function useAssets(): Record<string, Asset> {
-  const { t } = useTranslation();
-
-  // Onboarding2
-  const onboarding2Background = isGPS
```
johnschoeman, line 12: so the main goal is to remove as many instances of `isGPS` as possible, as this a challenge to maintain and will only get worse as we add more tracing strategies.  So instead of checking at each instance of some copy, we provide an interface of whole set of strategy specific copy/assets/whatever, expect the tracing strategy to provide that wholesale, and then we only need to check `isGPS` once and provide the all of content as a single group of data.

app/views/About.js
```diff
@@ -11,16 +11,17 @@ import {
...
 import fontFamily from './../constants/fonts';
 import { NavigationBarWrapper, Typography } from '../components';
-import { useAssets } from '../TracingStrategyAssets';
 
 import { Colors, Spacing } from '../styles';
 
 export const AboutScreen = ({ navigation }) => {
   const { t } = useTranslation();
-  const { aboutHeader } = useAssets();
+  const { useCopy } = useContext(TracingStrategyContext);
```
johnschoeman: this we could probably consolidate into one custom hook: something like: `useStrategyContent`

JacobJaffe:
Throwing an alternative approach out there: for copy assets, I've been thinking it might be cleaner to make the abstraction at the i18next namespace level. So that at the language config, BT vs. GPS namespace is selected.

johnschoeman:
> Throwing an alternative approach out there: for copy assets, I've been thinking it might be cleaner to make the abstraction at the i18next namespace level. So that at the language config, BT vs. GPS namespace is selected.

Totally agree, that struck me as a bigger change and i mostly wanted to start with a clean interface on the react side.  This feels like a meaningful first step towards that, or do you think that we can skip this and go straight to doing this at the i18next level?

tparesi (commented):
o0o0o I like this!

tparesi (approved):
Looks great

PR #1083, opened 2020-06-22
Export Template | StatusBar Hook

JacobJaffe:
image

`useStatusBarEffect` for export template, so that switching tabs doesn't overwrite status bar state (now that the first screen is a tab this matters).

### DIFF
--- a/app/views/Export/ExportTemplate.js
+++ b/app/views/Export/ExportTemplate.js
@@ -1,7 +1,6 @@
 import React, { useEffect } from 'react';
 import {
   ScrollView,
-  StatusBar,
   StyleSheet,
   View,
   TouchableOpacity,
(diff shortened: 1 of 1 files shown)


app/views/Export/ExportTemplate.js
```diff
@@ -65,6 +51,7 @@ export const ExportTemplate = ({
   bodyLinkOnPress,
   ignoreModalStyling, // So first screen can be slightly different in tabs
 }) => {
+  useStatusBarEffect(lightTheme ? 'dark-content' : 'light-content');
```
johnschoeman, line 54: My understanding is that we are moving away from the in app Theme stuff.  should this just be
```suggestion
  useStatusBarEffect('dark-content');
```
?

johnschoeman (approved):
LGTM

app/views/Export/ExportTemplate.js
JacobJaffe: Ideally yeah. There are a few places in the app still using the legacy system, mostly because the buttons were built around theming.

PR #1084, opened 2020-06-22
Feature Flag Custom URL

JacobJaffe:
- Adds prop typing for `FeatureFlag` 
- Adds a feature flag for the custom url input. @mattThousand I recognize you're removing the .env, so awareness depending order these merge in.


johnschoeman (approved):
💯

PR #1086, opened 2020-06-22
Onboarding Template Overhaul

JacobJaffe:
After the QA report, there were lots of issues around on-boarding flow screens. Because the files are copied from each other, I transitioned the onboarding pages to use an `OnboardingTemplate`, and turned the permission screens into two separate screens. That should help us in debugging the permission logic. 

image
image
image
image
image
Scrolling now:
image
image
image

iphoneX:
image
image


app/views/onboarding/OnboardingTemplate.tsx
```diff
@@ -0,0 +1,122 @@
...
+  body,
+  primaryButtonLabel,
+  primaryButtonOnPress,
+  secondaryButtonLabel,
+  secondaryButtonOnPress,
+  invertIcon,
+  theme,
+}: OnboardingTemplateProps): JSX.Element => {
+  return (
+    <Theme use={theme === 'light' ? 'default' : 'violet'}>
```
JacobJaffe, line 43: @johnschoeman I'm sorry for digging us deeper in this theming hole, this is only being used for light / dark themed buttons. I'd like to refactor the button component soon and remove these themes here & in the export template

app/views/onboarding/OnboardingLocations.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useContext } from 'react';
```
JacobJaffe, line 1: @yonatanmk Check this file out for potential location services issues

yonatanmk (approved):
LGTM

app/views/onboarding/OnboardingNotifications.tsx
```diff
@@ -0,0 +1,50 @@
...
+  navigation: NavigationScreenProp<NavigationState, NavigationParams>;
+};
+
+const OnboardingNotifications = ({
+  navigation,
+}: OnboardingNotificationsProps): JSX.Element => {
+  const { authSubscription, location } = useContext(PermissionsContext);
+
+  const requestNotifications = async () => {
+    await location.request();
```
yonatanmk: This should definitely be `notification` not `location`

yonatanmk (changes requested):
Need to update `OnboardingNotifications` per my comment

app/views/onboarding/OnboardingNotifications.tsx
JacobJaffe: good catch yeah

app/views/onboarding/OnboardingLocations.tsx
```diff
@@ -0,0 +1,47 @@
...
+    await location.request();
+    await authSubscription.request();
+    completeOnboarding();
+  };
+
+  const dispatch = useDispatch();
+  const completeOnboarding = () => {
+    // TODO: investigate if this is broken. This is directly copied from our old OnboardingPermissions, but the types look wrong.
+    const storeData = ((location.status ===
+      PermissionStatus.GRANTED) as unknown) as string;
```
yonatanmk:
```suggestion
    const storeData = JSON.stringify(location.status === PermissionStatus.GRANTED);
```
I think the odd types were just added during the conversion to TypeScript. The `SetStoreData` method specifies that its parameters must be strings though it also calls `JSON.stringify` on any arguments that aren't strings

yonatanmk (approved):
Permissions are working for me once again

app/views/onboarding/OnboardingTemplate.tsx
johnschoeman: makes sense to me!

johnschoeman (approved):
Overall looks great and is quite the improvement!

As a side note, I am less of a fan of the ComponentTemplate pattern as it strikes me as being too dry in the view layer, where you typically want to have more flexibility, but nbd tbh.

PR #1090, opened 2020-06-22
Fix Enable Location Popup Bug

yonatanmk:
#### Description:
Fix a bug on IOS where the user if constantly prompted to enable their location if they set their Location Permission to `Ask Next Time`. 

#### Linked issues:
https://pathcheck.atlassian.net/browse/SAF-731

#### How to test:

1. Install the app from scratch and go through onboarding. Select `Maybe Later` on the Location Permissions screen. You should not see the Enable Location popup on the next screen.

1. With the app already installed, set the Location Permissions to `Ask Next Time`. Close and reopen the app. You should not You should not see the Enable Location popup on the next screen.


app/services/LocationService.js
```diff
@@ -191,24 +192,27 @@ export default class LocationServices {
...
-        reason: this.DEVICE_LOCATION_OFF,
+        reason: DEVICE_LOCATION_OFF,
         hasPotentialExposure,
         isRunning,
       };
     }
-
-    if (!isAppGpsEnabled) {
+    /* 
+      On Android: isAppGpsEnabled = 0 means Never Use Location
```
johnschoeman: should these values be set to constants?
something like
```
IOS_NEVER_USE_LOCATION = 0
```

johnschoeman: Should these values be set as constants?
something like:
```
IOS_NEVER_USE_LOCATION = 0
```

johnschoeman (approved):

app/services/LocationService.js
yonatanmk: 👍

PR #1092, opened 2020-06-22
Main Screen | Small Screen Fixes

JacobJaffe:
### These changes are 99% snapshots de-nesting one component. Disregard the large amount of dif changes

This changes the view layout on the main pages so that the `BackgroundImage`'s layout is positioned separately from the content, allowing us to position the animation / background higher on small screens by a minimal amount of conditional screen size logic.
Now:
image
image

(same on larger:)
image

Before:
image
image


johnschoeman (approved):
Oh man those snapshots.

This highlights how we probably dont want a base component here though.

PR #1093, opened 2020-06-22
Polyfill FlatMap

JacobJaffe:
TIL different iOS versions might need polyfills. 
https://github.com/facebook/react-native/issues/26283

Before:
image

After:
image


johnschoeman (approved):
TIL.
👍

PR #1094, opened 2020-06-23
Skip Notification Onboarding for Android

JacobJaffe:
Regression introduced by #1086 

Android should only enable location permissions.

### DIFF
--- a/app/views/onboarding/Onboarding4.js
+++ b/app/views/onboarding/Onboarding4.js
@@ -15,15 +15,18 @@ const Onboarding4 = (props) => {
   } = useAssets();
   const { t } = useTranslation();
 
-  const handleOnPressNext = () =>
+  const gpsNext = () =>
     props.navigation.replace(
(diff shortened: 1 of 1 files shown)


app/views/onboarding/Onboarding4.js
```diff
@@ -17,7 +18,10 @@ const Onboarding4 = (props) => {
   const handleOnPressNext = () =>
```
johnschoeman: could we avoid the nested ternaries with something like
```suggestion
  const handleOnPressNext = () => {
     const nextScreen = () => {
       if (isGPS && isPlatformiOS()) {
         return Screens.OnboardingNotificationPermissions
       } else if (isGPS) {
         return Screens.OnboardingLocationPermission
        } else {
        Screens.EnableExposureNotifications
        }
        props.navigation.replace(nextScreen()
        }
     
```

johnschoeman (approved):

PR #1095, opened 2020-06-23
Android Backhandling Crash Fix

JacobJaffe:
We have been incorrectly using `useEffect` to add backhandlers. https://reactnavigation.org/docs/custom-android-back-button-handling/

This has been causing multiple triggered backhandled functions, causing issues (crashes!) on android devices when stacks have been navigated to. 

These extra backhandles are largely unnecessary, so I've removed them, except on the specific screens they're intended for:
- `Export` flow: to prevent navigating out of without a warning
- `ENDebugMenu`, `EnLocalDiagnosisKeyScreen`: because I haven't tested this on BT. 

Instead, we can delegate behavior React Navigation's opinions, and build our navigation hierarchy in a compatible manner (which it currently is).

A few minor fixes that stood out on android also:
- vertically centers back button on navigation top bar
- adds top padding to navigation top bar on android
- centers globe on settings screen for languages
- convert Util file to typescript

image
image


johnschoeman (approved):
Awesome, thanks for grabbing this.

PR #1098, opened 2020-06-23
Create Separate MainActivity for BT Development

octohub:
Pull out `MainActivity` from the main folder to allow one for the `GPS` build and the `BT` build. I want @JacobJaffe to smoke test the GPS build from this PR to make sure this doesn't introduce regressions there. The goal is to give BT development it's own MainActivity to work in. One that would not introduce changes to the GPS build.


JacobJaffe (approved):
Tested building a signed local bundle, and turning that bundle into an APK and running on device. 
All good there!

Will test the e2e google test as part of a later deployment today, I don't think that should block this getting in

PR #1103, opened 2020-06-24
Fix Request Location Permissions Bug

yonatanmk:
#### Description:
Fix a bug in which you cannot enable location permissions follow certain onboarding selections. 

#### Linked Tickets:
https://pathcheck.atlassian.net/browse/SAF-759

#### How to test:
Do through the onboarding flow and select `Maybe Later`. On the main screen pressing `Allow Location Access` should open the enable location popup. If the pop has already been opened, then pressing the button should navigate to the ios app settings.

### DIFF
--- a/app/gps/PermissionsContext.tsx
+++ b/app/gps/PermissionsContext.tsx
@@ -54,6 +54,7 @@ interface PermissionContextState {
     check: () => void;
     request: () => void;
   };
+  requestLocationSettings: () => void;
   requestNotificationSettings: () => void;
 }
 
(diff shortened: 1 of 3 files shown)


app/gps/PermissionsContext.tsx
```diff
@@ -171,6 +178,13 @@ const PermissionsProvider = ({
     }
   };
 
+  const requestLocationSettings = async () => {
+    const status = await requestLocationPermission();
+    if (statusToEnum(status) === PermissionStatus.DENIED) {
```
JacobJaffe, line 183: To double check, this is different than the status code if the user declines in _this_ session?
/ What's that status code?

JacobJaffe: I see 
"BLOCKED" is an option, when does that get triggered?

JacobJaffe (approved):

PR #1104, opened 2020-06-24
iPhone 8 / Nexus 4 UI improvements, larger accessibility text sizes improvements

jeffstolz:
#### Description:

This commit resolves UI issues present on the on iPhone 8 and with larger accessibility text sizes. Includes:

- Update enable exposure notification screen to match mockups and resolve small screen padding issues, add scrollview.
- fix overflow issue with tab bar on exposure history
- Shift position of home svg to accommodate larger text sizes
- Conditionally hide onboarding bg image for iphone 8 or smaller phones
- add flex wrap to calendar header
- Updates snapshots

#### Linked issues:

https://trello.com/c/WTz6cti5/74-as-nexus-4-and-iphone-8-users-we-want-an-optimal-user-experience-that-is-synonomous-with-those-with-larger-phones

#### Screenshots:

Screen Shot 2020-06-24 at 2 53 50 PM
Screen Shot 2020-06-24 at 2 53 44 PM
Screen Shot 2020-06-24 at 2 53 35 PM
Screen Shot 2020-06-24 at 2 53 26 PM
Screen Shot 2020-06-24 at 2 53 16 PM

### DIFF
--- a/app/bt/Home/index.tsx
+++ b/app/bt/Home/index.tsx
@@ -45,7 +45,7 @@ const styles = StyleSheet.create({
   iconContainer: {
     position: 'absolute',
     left: -(Layout.screenWidth / 2),
-    bottom: -(Layout.screenHeight / 2),
+    bottom: -(Layout.screenHeight / 3),
   },
   homeContainer: {
     flex: 1,
--- a/app/views/onboarding/OnboardingTemplate.tsx
+++ b/app/views/onboarding/OnboardingTemplate.tsx
@@ -11,7 +11,7 @@ import { SvgXml } from 'react-native-svg';
 import { Button } from '../../components/Button';
 import { Typography } from '../../components/Typography';
 
-import { Colors, Spacing } from '../../styles';
+import { Colors, Layout, Spacing } from '../../styles';
 import { Theme } from '../../constants/themes';
 
 type OnboardingTemplateProps = {
@@ -47,7 +47,9 @@ const OnboardingTemplate = ({
           backgroundColor='transparent'
           translucent
         />
-        <ImageBackground source={background} style={styles.backgroundImage} />
+        {Layout.screenWidth <= Layout.smallScreenWidth ? null : (
+          <ImageBackground source={background} style={styles.backgroundImage} />
+        )}
         <View style={styles.content}>
           <ScrollView
             alwaysBounceVertical={false}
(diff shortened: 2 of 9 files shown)


app/views/onboarding/OnboardingTemplate.tsx
```diff
@@ -47,7 +47,9 @@ const OnboardingTemplate = ({
           backgroundColor='transparent'
           translucent
         />
-        <ImageBackground source={background} style={styles.backgroundImage} />
+        {Layout.screenWidth <= Layout.smallScreenWidth ? null : (
```
johnschoeman, line 50:
```suggestion
isLargeScreen = Layout.screenWidth > Layout.smallScreenWidth
...
return (
...
        {isLargeScreen ?  (<ImageBackground source={background} style={styles.backgroundImage} />) : null
```

app/bt/Home/index.tsx
```diff
@@ -45,7 +45,7 @@ const styles = StyleSheet.create({
   iconContainer: {
     position: 'absolute',
     left: -(Layout.screenWidth / 2),
-    bottom: -(Layout.screenHeight / 2),
+    bottom: -(Layout.screenHeight / 3),
```
johnschoeman, line 48: This whole icon as a background image is a bit wonky.  What do you think about using an actual background image for the concentric circles?

johnschoeman (approved):
A few thoughts.

app/views/onboarding/OnboardingTemplate.tsx
johnschoeman: This also strikes me as a bit of a hack, is there a way we can do this without querying the screen width?

JacobJaffe: Maybe introducing "@ 2", "@ 3" size images? I wonder how much runway we can get with that

jeffstolz: Jacob, would that resize the bg image based on device? Since the copy is going to take up the entire screen for this size phone, and the bg image is mainly texture, so I still think removing the bg for small phones is best. Any suggestions on a less hacky way? I couldn't come up with anything better here.

app/bt/Home/index.tsx
jeffstolz: I agree that it's wonky - the only benefit is that it ensures alignment with the bg rings and the icon centered on them. I think we can prevent any issues if we separate the rings in a bg image and position that and the icon absolute to the top? It's also implemented like this throughout, do you think that change should be tackled in a separate PR (or separate commit in this PR) so it can be rolled back individually?

achen178 (approved):
lgtm, we might want to revisit a cleaner way of responsiveness

PR #1107, opened 2020-06-25
Handle "Dark Mode" in iOS

mattThousand:
### Why
We'd like the app to reflect our designs. The designs do not handle multiple user interface styles (i.e. "dark mode"), so in order to avoid, for example, white text on a white background while in dark mode, we set the user interface style to always be `Light`.

### This Commit
This commit sets the user interface style to `Light` for both the `BT` and `GPS` iOS targets

### Before
Jun-24-2020 21-08-48

### After
IMG_D9F3605D3F8A-1


JacobJaffe (approved):
Terrific catch
Also, that screen recording is a very soothing gif to watch

johnschoeman (approved):
🥇

PR #1109, opened 2020-06-25
Minor updates for demo

johnschoeman:
Why:
We would like to polish a few loose ends for upcoming demos

This commit:
- Removes 'Current daily reports' from the exposure info detail as its no
longer in the designs.
- Updates the HA name on the next steps screen to make it consistent
with other parts of the app
- Turns off the badge for new exposure on the exposure history icon


tparesi (approved):
LGTM

PR #1110, opened 2020-06-25
Fix start affected user flow crash

johnschoeman:
Why:
Currently the app will crash if a user tries to start the affected user
flow while there are not health authorities in the store.

This commit:
Provides a fake health authority if there are none such that the app
will not crash. A future commit should properly handle this null state
and we should remove this fake authority.

### DIFF
--- a/app/views/Export/ExportIntro.js
+++ b/app/views/Export/ExportIntro.js
@@ -20,11 +20,36 @@ export const ExportIntro = ({ navigation }) => {
   const authorities = useSelector(healthcareAuthorityOptionsSelector);
   const selectedAuthorityDummy = authorities[0];
 
-  const onNext = () =>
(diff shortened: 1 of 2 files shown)


app/views/Export/ExportIntro.js
```diff
@@ -20,11 +20,36 @@ export const ExportIntro = ({ navigation }) => {
...
+            latitude: 43.0,
+            longitude: 72.0,
+          },
+        },
+        org_id: '1234',
+        cursor_url: '',
+        public_api: '',
+        internal_id: '',
+      };
+      navigation.navigate(Screens.ExportCodeInput, {
```
tparesi, line 46: The dummy HA might cause on issue on GPS with entering a valid code on the ExportCodeInput screen, since there is no api to validate against.

johnschoeman:
closing as the issue is related to .env.

PR #1113, opened 2020-06-25
Dont post export code for bt

johnschoeman:
Why:
Currently for the BT app we are not setup for posting export codes on
the affected user flow, but would like to fake it for demo purposes.

This commit:
Removes the request for the BT build.


JacobJaffe (approved):
LGTM. Might be nice to have a comment in the code if this is intended to be temporary

tparesi (approved):
LGTM

PR #1115, opened 2020-06-25
Submit Positive Diagnosis (BT)

mattThousand:
### Why
We'd like to submit diagnosis keys to the GAEN server when going through the "submit positive diagnosis" flow

### This Commit
- Cleans up code related to unarchiving key archives fetched from GAEN server on iOS  
- Adds the HMAC key (public key used to sign keys from the GAEN server) to the request that submits keys to the GAEN server
- Removes `git` submodules (also updates the README accordingly)
- Converts `js` files related to publishing exposure keys to typescript

** Note - to run this it's necessary to have the latest .env.bt file from the `pathcheck-mobile-resources` private repo, which contains the HMAC key


johnschoeman (approved):
LGTM!

PR #1120, opened 2020-06-25
Feature Flag, Secret Developer Setting

JacobJaffe:
Changes feature flags to be in redux, and to be accessed by tapping the About screen header 20 times.
This removes the need to have them in .env variables, and allows us to distribute this to testers.

Part of the effort to move away from needing a staging version of the app

image


app/store/reducers/rootReducer.ts
```diff
@@ -2,10 +2,12 @@ import { combineReducers } from 'redux';
 
 import onboardingReducer from './onboardingReducer';
 import healthcareAuthoritiesReducer from './healthcareAuthoritiesReducer';
+import featureeFlagsReducer from './featureFlagsReducer';
```
yonatanmk, line 5: typo?

app/views/About.js
```diff
@@ -13,9 +13,24 @@ import { useStrategyContent } from '../TracingStrategyContext';
...
+import toggleAllowFeatureFlagsAction from '../store/actions/featureFlags/toggleAllowFeatureFlagsEnabledAction';
+import { TouchableWithoutFeedback } from 'react-native-gesture-handler';
 import { Colors, Spacing, Typography as TypographyStyles } from '../styles';
 
 export const AboutScreen = ({ navigation }) => {
+  const dispatch = useDispatch();
+
+  const [clickCount, setClickCount] = useState(0);
+  useEffect(() => {
+    console.log(clickCount);
```
yonatanmk: Do we want to remove this?

yonatanmk (approved):
Works well for me

app/views/About.js
JacobJaffe: yeah, will do. good catch

app/store/reducers/featureFlagsReducer.ts
```diff
@@ -0,0 +1,34 @@
+import { createReducer } from '@reduxjs/toolkit';
+import { FeatureFlagOption } from '../types';
+
+import toggleAllowFeatureFlagsEnabledAction from '../actions/featureFlags/toggleAllowFeatureFlagsEnabledAction';
+import toggleFeatureFlagAction from '../actions/featureFlags/toggleFeatureFlagAction';
+
+const AllFlagsOff: Record<FeatureFlagOption, boolean> = {
+  [FeatureFlagOption.GOOGLE_IMPORT]: false,
```
johnschoeman, line 8: This strikes me as a bit odd, if we are using types on the record here, why do we need have these dynamic keys?

app/store/types.ts
```diff
@@ -28,3 +28,9 @@ export type AppThunk<ReturnType = void> = ThunkAction<
   unknown,
   Action<string>
 >;
+
+export enum FeatureFlagOption {
```
johnschoeman, line 32: What do you think about a union type here rather than the enum?

app/store/reducers/featureFlagsReducer.ts
johnschoeman: I might consider an interface here rather than a record as a record without all of the keys would be considered valid by the complier but is likely not what we want to be representing.

app/views/About.js
```diff
@@ -13,9 +13,24 @@ import { useStrategyContent } from '../TracingStrategyContext';
...
+    if (clickCount === 20) {
```
johnschoeman: what do you think about using a constant for this number?

Also 20 strikes me as a bit annoyingly high of a number, do you think something like 5 would be sufficient?

johnschoeman (approved):
Code looks good, a few minor type thoughts.

This is also something that worries me a bit to be adding given the public scrutiny that this app undergoes.  Is there a world where this is behind a __DEV__ flag? or do we need these options on production builds?  And if it does need to be in production builds I might make sure to get a review from Sam as to make sure there's no PR concerns to be had with Dev things being in release builds.

app/store/types.ts
JacobJaffe: The values themselves are used, e.g. as reducer keys, or for flag to name mapping, so I think an enum makes sense? Not sure how you're envisioning a union type would work

app/store/reducers/featureFlagsReducer.ts
JacobJaffe: omitting a key throws this error:
image

Is there an additional check that the compiler is missing?

app/views/About.js
JacobJaffe: Updated to a constant, good point. 
How about we compromise at 10 😄

app/store/reducers/featureFlagsReducer.ts
johnschoeman: TIL, that ts enums are exhaustively checked as keys on records, pretty neat.  I still find this a bit counter intuitive though.

app/store/types.ts
johnschoeman: to be clear, i think what you have here is totally reasonable. I just tend to prefer avoiding enums when possible though, cause numeric enums have some pitfalls, and I find the enum { KEY = 'KEY' } syntax of string enums a bit redundant, so this is purely stylistic I wouldnt take the time to change your commit unless you feel like it.

that said, I would consider something like:

```typescript
type FeatureFlag = 'GoogleImport' | 'CustomUrl' | 'DownloadLocally'

const FeatureConfig: { [key in FeatureFlag]: boolean } = {
  GoogleImport: boolean
  CustomUrl: boolean
  DownloadLocally: boolean
}
```

vlasac (approved):

PR #1121, opened 2020-06-26
A Few Minor Updates

mattThousand:
### This Commit
- Adds the correct mapping between the `ExposureConfiguration` type and the `ENExposureConfiguration` type
- Adds a message body for the "new exposure" notification push UI
- Uses the development entitlement for debug mode on iOS, so that we can detect exposures when running the app locally
- Fixes a validation issue in `app/api/healthcareAuthorities/getHealthcareAuthoritiesApi`


ios/BT/Extensions/Foundation/String+Extensions.swift
```diff
@@ -16,6 +16,7 @@ extension String {
   static let bluetoothNotificationIdentifier = "bluetooth-off"
   static let exposureDetectionErrorNotificationTitle = "Error Detecting Exposures"
   static let exposureDetectionErrorNotificationBody = "An error occurred while attempting to detect exposures."
+  static let newExposureNotificationBody = "Someone you were near recently has been diagnosed with COVID-19. Tap for more details."
```
tparesi, line 19: Curious - do we have a way to translate the strings that are in native code?

mattThousand: @tparesi line 149 in `ExposureManager` feeds the string into apple's localization API

tparesi (approved):
LGTM

app/api/healthcareAuthorities/getHealthcareAuthoritiesApi.ts
```diff
@@ -25,7 +25,8 @@ const getHealthcareAuthoritiesApi = async (
   yamlUrl: string = AUTHORITIES_YAML_ROUTE,
 ): Promise<HealthcareAuthority[]> => {
   const yamlString = await fetch(yamlUrl).then((res) => res.text());
-  let { authorities } = Yaml.safeLoad(yamlString);
+  const record = Yaml.safeLoad(yamlString) as Record<string, any>;
```
johnschoeman: just a note, we really want to be avoiding type casting, as its basically removing all of the benefit of ts. Is there away we can express this without the `as`?

ios/BT/API/Model/ExposureConfiguration.swift
```diff
@@ -25,11 +25,11 @@ extension ExposureConfiguration {
 
   var asENExposureConfiguration: ENExposureConfiguration {
     let config = ENExposureConfiguration()
-    config.metadata = ["attenuationDurationThresholds": [50, 70]]
-    config.attenuationLevelValues = [1, 2, 3, 4, 5, 6, 7, 8]
-    config.daysSinceLastExposureLevelValues = [1, 2, 3, 4, 5, 6, 7, 8]
-    config.durationLevelValues = [1, 2, 3, 4, 5, 6, 7, 8]
-    config.transmissionRiskLevelValues = [1, 2, 3, 4, 5, 6, 7, 8]
+    config.metadata = ["attenuationDurationThresholds": attenuationDurationThresholds]
```
johnschoeman, line 28: 💯

johnschoeman (approved):
dope.  One ts comment, which might be fun to pair on if you are interested!

app/api/healthcareAuthorities/getHealthcareAuthoritiesApi.ts
mattThousand: `const record: Record<string, any> = Yaml.safeLoad(yamlString)` <- is this any better? Also unsure if there's a way to get around using `any` as the record value type here (`[String: Any]` is often how I unpack JSON in swift, but using `Any` always feels a little suspect) 🤔

johnschoeman: yup! the `const thing: Type = foo` syntax will fail tsc if foo is not of the type of Type as were `const thing = foo as Type` will not.

anys are also good to avoid, looks like something like `Record<string, string>` might work?

mattThousand: ah, I think we might be stuck with `any`; there are a few different types in there. But good to know - thank you for unpacking things conceptually for me!

PR #1122, opened 2020-06-26
Grab the HA to use in copy on app

tparesi:
#### Description:

In order to easily sub in the appropriate HA, different AUTHORITIES_YAML_ROUTE's can be used in the env files. This is a simple solution to allow for customization inside the app.

#### Linked issues:

[Customize HA in app](https://trello.com/c/L4pNhsN4/71-as-a-dev-team-we-have-a-way-to-customize-the-webview-url-and-the-health-authority-name-in-the-app)

#### How to test:

By changing the URL inside the env.bt file to point to a URL that provides a different HA, the content within the app will change. We are using the trusted_authorities repo to set this `https://raw.githubusercontent.com/Path-Check/trusted-authorities/master/staging/authorities.1.0.1.yaml`. I would recommend using a new branch on that repo, updating the data appropriately, and changing the url inside the env.bt to point to that url instead.

### DIFF
--- a/app/locales/en.json
+++ b/app/locales/en.json
@@ -26,7 +26,7 @@
     "isolate_description": "You <1>have some symptoms</1> that may be related to COVID-19. Call your healthcare provider if your symtoms get worse. Start home isolation.",
     "isolate_cta": "Next",
     "share_title": "Publish anonymized data",
-    "share_description": "Your anonymous data helps the Boston Public Health Commission analyze, track, and contain the spread of COVID-19. Your data will be aggregated that from other PathCheck users who choose to share.\n\nYour data contains your survey responses and zip code, but no additional or personally-identifying information.\n\nThe Boston Public Health Commission will retain your anonymous data for a specific period of time.",
+    "share_description": "Your anonymous data helps the {{ha_name}} analyze, track, and contain the spread of COVID-19. Your data will be aggregated that from other PathCheck users who choose to share.\n\nYour data contains your survey responses and zip code, but no additional or personally-identifying information.\n\nThe {{ha_name}} will retain your anonymous data for a specific period of time.",
     "share_screening_title": "Anonymized Screening Data",
     "share_screening_description": "Share anonymized answers to screening questions with your local healthcare authority.",
     "share_cta_skip": "No thanks, I don't want to share data",
--- a/app/views/ExposureHistory/NextSteps.tsx
+++ b/app/views/ExposureHistory/NextSteps.tsx
@@ -1,14 +1,24 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
 import { TouchableOpacity, View, StyleSheet } from 'react-native';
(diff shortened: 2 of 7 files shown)


app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -1,24 +1,34 @@
...
 import { Typography } from '../../components/Typography';
 import { NavigationBarWrapper } from '../../components/NavigationBarWrapper';
 import { Screens, useStatusBarEffect } from '../../navigation';
+import getHealthcareAuthorities from '../../store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction';
+import healthcareAuthorityOptionsSelector from '../../store/selectors/healthcareAuthorityOptionsSelector';
 
 import { Buttons, Spacing, Typography as TypographyStyles } from '../../styles';
 
 const NextSteps = (): JSX.Element => {
+  const dispatch = useDispatch();
```
johnschoeman, line 15: Using redux here seems a bit heavy to me, is there a reason we went with it particular?

It also strikes me as something that we want to keep separate from the user selecting HA flow as this data likely will come from the HA at the distribution level. so it might be a good idea to allow these data sources to diverge.  What do you think about introducing an env var `HA_NAME=ha_name` and provide this via react-native-config?  This seems lightest weight at this point and we can add a context or subscribe to the gobal state store in a future commit if it turns out we need it after learning more about this feature.

app/locales/en.json
```diff
@@ -26,7 +26,7 @@
     "isolate_description": "You <1>have some symptoms</1> that may be related to COVID-19. Call your healthcare provider if your symtoms get worse. Start home isolation.",
     "isolate_cta": "Next",
     "share_title": "Publish anonymized data",
-    "share_description": "Your anonymous data helps the Boston Public Health Commission analyze, track, and contain the spread of COVID-19. Your data will be aggregated that from other PathCheck users who choose to share.\n\nYour data contains your survey responses and zip code, but no additional or personally-identifying information.\n\nThe Boston Public Health Commission will retain your anonymous data for a specific period of time.",
+    "share_description": "Your anonymous data helps the {{ha_name}} analyze, track, and contain the spread of COVID-19. Your data will be aggregated that from other PathCheck users who choose to share.\n\nYour data contains your survey responses and zip code, but no additional or personally-identifying information.\n\nThe {{ha_name}} will retain your anonymous data for a specific period of time.",
```
johnschoeman, line 29: it might be worth pinging design on the content here. It might be nice to have a generic description and just have the HA name in a title or something similar.

johnschoeman (approved):
Looks good! a few thoughts.  Biggest on around making sure redux is the correct choice here. Ill defer to your thoughts as you have the context here. but I worry about adding more coupling to the gobal app state than necessary as I see this as a build time configuration thing.

app/locales/en.json
johnschoeman: so basically if we can avoid the `{{ }}` that would be a win imo.

PR #1123, opened 2020-06-26
[SAF-712] - perform intersection check after editing HA list

vlasac:
#### Description:
This PR solves the problem of skipped intersection checks after editing the list of Healthcare Authorities.

#### How to test:
- Open the list of available HAs
- Select or deselect any number of HAs
- The exposure check should start after 1 second of no interaction with the list
- If you have GPS data on your device, you might get an exposure notification


app/services/BackgroundTaskService.js
```diff
@@ -1,11 +1,12 @@
 import BackgroundFetch from 'react-native-background-fetch';
 
+import IntersectService from './IntersectService';
```
JacobJaffe: Can we export the singleton rather than have a singleton as a member of the class?

app/services/IntersectService.js
```diff
@@ -0,0 +1,19 @@
+import { checkIntersect } from '../helpers/Intersect';
```
JacobJaffe: Can we start new files in TS?

app/views/Main.js
```diff
@@ -41,8 +53,8 @@ export const Main = () => {
     NotificationService.configure(notification.status);
   }, [setCanTrack, notification.status]);
 
-  useEffect(() => {
-    updateStateInfo();
+  useDeepCompareEffect(() => {
```
JacobJaffe: Why do we need this deep effect? Is this just for checking if selected authorities is different? 
I would recommend writing this more explicitly for what is triggering the effect / documenting why this is needed otherwise.

app/views/Main.js
```diff
@@ -19,18 +19,30 @@ import PermissionsContext, {
...
+import { debounce } from 'lodash';
 
 export const Main = () => {
   useStatusBarEffect('light-content');
   const navigation = useNavigation();
   const { notification } = useContext(PermissionsContext);
   const hasSelectedAuthorities =
     useSelector(selectedHealthcareAuthoritiesSelector).length > 0;
+
+  const debounceUpdateStateInfo = useRef(
```
JacobJaffe: Won't this delay the UI state changes by a second? Should this be throttled?

app/views/Main.js
```diff
@@ -19,18 +19,30 @@ import PermissionsContext, {
...
+    selectedHealthcareAuthoritiesSelector,
+  );
+
   const [canTrack, setCanTrack] = useState(true);
 
   const checkForPossibleExposure = () => {
     BackgroundTaskServices.start();
-    checkIntersect();
+    const intersectionStatus = IntersectService.instance.checkIntersect(true);
+    console.log(`[intersection] ${intersectionStatus}`);
```
JacobJaffe: Can we remove console logs?

app/views/Main.js
```diff
@@ -53,7 +65,12 @@ export const Main = () => {
       AppState.removeEventListener('change', updateStateInfo);
       unsubscribe();
     };
-  }, [navigation, updateStateInfo]);
+  }, [
+    navigation,
+    updateStateInfo,
+    selectedAuthorities,
```
JacobJaffe: can these just be split into two different useEffects? 
More broadly, I don't understand why we're doing the triggering background tasks depndent on changes to the redux state within the render cycle of a screen

JacobJaffe (changes requested):

app/views/Main.js
vlasac: Since we're using debouncing, the list of HAs could potentially end up the same as before any of the items were clicked. The deep compare effect is preventing redundant intersect checks from firing if the HA list is indeed the same as before.

app/views/Main.js
vlasac: I created the redux middleware that triggers the intersection logic, which then starts the timer, so I was able to completely remove this logic from `Main.js` . Thanks for the advice on that

app/views/Main.js
vlasac: Debouncing is removed because that is now handled in a similar way in the `IntersectService` class.

app/views/Main.js
vlasac: As the whole logic is now out of the render cycle, this is also not needed indeed.

app/store/middleware/onChangedSelectedHealthAuthorities.ts
```diff
@@ -0,0 +1,23 @@
...
+const onChangedSelectedHealthAuthorities: Middleware<Dispatch> = (
+  store: MiddlewareAPI,
+) => (next: Dispatch<AnyAction>) => (action: AnyAction): unknown => {
+  const {
+    healthcareAuthorities: { selectedAuthorities: stateBefore },
+  } = store.getState();
+
+  const result = next(action);
+
+  const {
```
JacobJaffe, line 13: Can these be wrapped in a try to ensure the chain will  always complete (return result)

app/services/IntersectService.ts
```diff
@@ -0,0 +1,35 @@
...
+    intersect(healthcareAuthorities, bypassTimer).then(() => {
+      this.isServiceRunning = false;
+      if (this.nextJob) {
+        const job = this.nextJob;
+        this.nextJob = null;
+        this.checkIntersect(job, bypassTimer);
+      }
+    });
+
+    BackgroundTaskService.start();
```
JacobJaffe, line 28: Why do we need to call start every time? Doesn't this include config setting?

app/services/IntersectService.ts
```diff
@@ -0,0 +1,35 @@
...
+      this.nextJob = healthcareAuthorities;
+      return 'skipped';
+    }
+    this.isServiceRunning = true;
+
+    intersect(healthcareAuthorities, bypassTimer).then(() => {
+      this.isServiceRunning = false;
+      if (this.nextJob) {
+        const job = this.nextJob;
+        this.nextJob = null;
```
JacobJaffe, line 23: This looks to me like it will cancel lined up jobs?
I'm confused what the intention here is:
is this intended as a job queue, or to guarantee only one intersection search per HA is occurring?
I think it's the second for the case we're solving, so maybe we can tighten these invariants to only being a map of in progress one-off intersections?

My concern here i that this could be triggered for two separate valid HA lists and one would cancel the other

JacobJaffe (changes requested):

app/services/IntersectService.ts
vlasac: This will be better addressed in a followup PR.

app/services/IntersectService.ts
vlasac: I totally agree with you and a better approach for the intersection logic would be checking intersection per health authority and also keeping the exposure data per health authority, which will put us in a better spot for handling this issue.

JacobJaffe (approved):
Sounds good on follow ups!

PR #1126, opened 2020-06-26
Correct EN Permissions State on App Launch

mattThousand:
### Why
We'd like the UI to reflect the current authorization/enabled state of native EN permissions on app launch

### This Commit
This commit introduces a method to fetch the current permissions status, and calls it such that `ExposureNotificationContext` is up to date on app launch

Jun-26-2020 13-49-39
Jun-26-2020 13-50-26


app/bt/ExposureNotificationContext.tsx
```diff
@@ -33,6 +33,8 @@ const ExposureNotificationsProvider = ({
   const [deviceStatus, setDeviceStatus] = useState<DeviceStatus>(initialStatus);
 
   useEffect(() => {
+    getCurrentENPermissionsStatus();
```
JacobJaffe: Maybe this should happen outside of the abstraction? 
So GPS and BT initialize their permission states?

JacobJaffe (commented):
Just a comment about the abstraction level for initialization. Non-blocking, just curious

octohub (approved):

app/bt/PermissionsContext.tsx
```diff
@@ -0,0 +1,129 @@
...
+  };
+  exposureNotifications: {
+    status: ENPermissionStatus;
+    check: () => void;
+    request: () => void;
+  };
+}
+
+const initialState = {
+  notification: {
```
mattThousand, line 33: we'll be adding UI to request `Notifications` permissions during onboarding soon, so adding this here now

app/bt/ExposureNotificationContext.tsx
mattThousand: good call - I think `ExposureNotificationContext` could be replaced by a context that handles "permissions" more generally, the way you have it set up on the GPS side. We'll be adding `Notifications` permissions on the BT side soon, so that could live in `bt/PermissionsContext` too 🤔

app/helpers/PermissionsHelpers.ts
```diff
@@ -0,0 +1,24 @@
+export enum PermissionStatus {
```
johnschoeman: less of a fan of `helper` files, especially for things that are domain concepts rather than utility functions.  What do you think about promoting this file to the top level as a named concept? something like:
`./app/permissionStatus.ts`

johnschoeman (approved):
This looks great!  Totally inline with the idea of making a common Permissions Context that is shared across all tracingStrategies which provide a list of possible Permissions as a prop to the provider.

PR #1127, opened 2020-06-26
Lokalise latest june 26

alpita-masurkar:
#### Description:
Lokalise latest updates- Tagalog mostly that we need for Guam
(Also trying to see if my snapshots have updated correctly... might convert to draft to fix this if the problem persists)


johnschoeman (approved):
LGTM!

PR #1128, opened 2020-06-26
Updated iOS schemes

jvisenti:
#### Description:

Our current deployment strategy for iOS is manual and error prone. This PR cleans up existing build configurations and transitions the project to using xcconfig files. There are 2 release tracks (staging and production) that use different bundle identifiers and versioning. This also gives us the flexibility to install development, staging, and prod builds side-by-side on the same device, ideally with different icons in the future to disambiguate.

* Updated build configurations
* Use xcconfig files
* Removed unnecessary schemes


johnschoeman (approved):
LGTM! thanks for introducing some better practices here!

Looks like we also need update the scheme/product names in the github actions. Let me/someone know if you want to pair on the github yaml, as they are a bit under-maintained at this point and I wouldnt want you to spend more time than necessary groking it.

jvisenti:
Closed and reopened as #1134 because GitHub wasn't updating the PR with the right commit after a rebase and force push.

PR #1130, opened 2020-06-28
Remove redundant styles

johnschoeman:
Why:
We currently have some redundant style information being used by some
older components. We would like to remove these and use the design
system styles everywhere.

This commit:
Removes constants/fonts and constants/colors and updates the components
which use them.


JacobJaffe (approved):
👏

PR #1131, opened 2020-06-28
Remove unused divider component

johnschoeman:
Why:
Currently we have an unused divider component in the components folder.
We would like to remove dead code as a matter of best practice.


jeffstolz (approved):
Definitely agree that we should remove this component so it doesn't get re-implemented 👍

PR #1132, opened 2020-06-28
Convert button.js to tsx and remove theme

johnschoeman:
Why:
We would like remove unused code and have as much of the app covered by
static analysis as possible. Currently the Button component is a heavily
used component that has a dependency on the ThemeProvider and does not
have type coverage. We would like to remove the dependency on the theme
provider and convert it to typescript.

This commit:
Converts Button.js to Button.tsx, removes the theme and converts a few
call sites of `<Button />` to a simpler implementation of
`<TouchableOpacity />`.

Note that we likely want to remove Button.tsx altogether in a future
commit as it doesnt provide much value and it has a maintenance cost. We
are not removing it altogether in this commit to reduce the size of the
commit and for sake of time.


JacobJaffe:
It's unclear to me why we'd go from `<Button>` to passing `Buttons.largeBlue` or other styles?
What's the ideal abstraction here?

I'm hesitant to remove structured styling components completely -- e.g. for buttons, having defined components gives a win on consistent accessibility. 
IMO we should have `<PrimaryButton>` `<SecondaryButton>` components, which would never need styling overriding, to get us out of these permutations of buttons with various overrides

johnschoeman:
> It's unclear to me why we'd go from `<Button>` to passing `Buttons.largeBlue` or other styles?
> What's the ideal abstraction here?

Hmm, i would say there is no ideal abstraction here, in the sense of we should be concrete in our intent and express it clearly. (If we start going down the path of `GenericTappable` or the like, then we are just making things harder to understand than they need to be)

> I'm hesitant to remove structured styling components completely -- e.g. for buttons, having defined components gives a win on consistent accessibility.
> IMO we should have `<PrimaryButton>` `<SecondaryButton>` components, which would never need styling overriding, to get us out of these permutations of buttons with various overrides

Its important for maintainability purposes to keep styles distinct form markup and view logic.  The one thing we know for sure is that requirements on look and feel will change overtime. So the goal with the `Buttons.largeBlue` is to cleanly separate style from view markup. By conflating the concepts we put the project at risk as we wont be able to update markup and style independently without needing significant rewriting of the app.

So if your idea is to have `<PrimaryButton>` and `<SecondaryButton>` have different markup and different view logic then sure, go for it. If its to enforce styles, then you're adding rigidity in a place we actively dont want it and making things more complex than they need to be. It would be better to have name styles of something like: `Buttons.primary` and `Buttons.secondary` which i think is better than `Buttons.largeBlue` tbh and could be applied to the the `PrimaryButton` and `SecondaryButton` components if we wanted, but I'll defer to the designers on best practices on design system stuff. cc: @ericwbailey , @jeffstolz.

ericwbailey:
I agree with John. If we're going to decouple the button from theme, we'd want to ensure that a label is required and ties into the computed accessible name.

I think this sentiment reflects my thoughts on the matter:

>cleanly separate style from view markup.

In that styled components seem to be a lot of unnecessary abstraction for questionable benefits, especially in a mobile app context.

My one concern is how customized whitelabeling will get, but I think that's a bigger question in terms of impact.

jeffstolz (approved):
I know there is some discussion going on here - I also agree with the benefit of these changes. I'd prefer to get this merged sooner rather than later as I'm currently implementing UI that could definitely benefit from this refactor.

PR #1133, opened 2020-06-29
Remove theme from onboarding screens

johnschoeman:
Why:
Currently the onboarding screens include a no longer wanted theme
provider. Given that we actively do not want to allow for dynamic theme
changes in the app we would like to remove the theme component.

This commit:
Refactors the onboarding components such that we can remove the theme
component while keeping the same visual design. We ended up needing to
remove the permissions screens from the purview of the
`OnboardingTemplate` as the permissions screens are significantly
different from the onboarding content screens in terms of markup, view
logic, and style so as to keep the onboarding template for them would
require introducing significant unnecessary complexity and ultimately
make the screens less amenable to future changes in requirements.

Note that this is entirely a refactor and no new functionality has been
added.


app/views/onboarding/OnboardingPermissions.js
```diff
@@ -1,196 +0,0 @@
-import React, { useContext, useState } from 'react';
```
johnschoeman, line 1: This file was no longer being used.

app/views/onboarding/OnboardingNotifications.tsx
```diff
@@ -1,49 +1,126 @@
...
+    await requestPermission();
+    continueOnboarding();
+  };
+
+  const handleOnPressMaybeLater = () => {
+    continueOnboarding();
+  };
 
   return (
-    <OnboardingTemplate
```
JacobJaffe, line 35: Why move away from template here?
Isn't the only thing different for these screens:

1. the color (light vs. dark theme)
2. having a tertiary button? ("maybe later")

We keep wins on screen accessibility / layouts (something we've struggled to consistently maintain so far) by having a same template across our on boarding screens

johnschoeman:
```
We ended up needing to
remove the permissions screens from the purview of the
OnboardingTemplate as the permissions screens are significantly
different from the onboarding content screens in terms of markup, view
logic, and style so as to keep the onboarding template for them would
require introducing significant unnecessary complexity and ultimately
make the screens less amenable to future changes in requirements.
```

A bit wordy, but i tried going down that route and it was a lot of effort to pull out that abstraction once the theme was gone.

In general trying to be overly dry with markup leads to views that are resistant to changes in future requirements. While its tempting to pull out abstractions based off of structure, over time you end up with these components that have dozens of conditionals, have multiple consumers all with slightly different needs, which leads to a situation where its more work thank necessary to update views to match new requirements.

 Its better to be dry in the data model and in the styles and maintain flexibility in the view layer.

Its a similar line of thinking to composition over inheritance if you are familiar.

johnschoeman: As for accessibility and layout, I agree we should be doing a better job with that, but that wont necessarily be addressed by putting template components everywhere. We still have to do the work of following best practices there.

jeffstolz (approved):
Great refactor - I am currently working on the redesigned onboarding screens and would like to rebase on to this work before submitting my PR.

PR #1134, opened 2020-06-29
Updated iOS schemes

jvisenti:
#### Description:

Our current deployment strategy for iOS is manual and error prone. This PR cleans up existing build configurations and transitions the project to using xcconfig files. There are 2 release tracks (staging and production) that use different bundle identifiers and versioning. This also gives us the flexibility to install development, staging, and prod builds side-by-side on the same device, ideally with different icons in the future to disambiguate.

* Updated build configurations
* Use xcconfig files
* Removed unnecessary schemes


jvisenti:
@johnschoeman I've updated the workflows so that they pass, but there is a new failing Build step that seems to be related to GCP. Does this need to be resolved before merge?

johnschoeman (approved):
Still looks good to me!

johnschoeman:
> @johnschoeman I've updated the workflows so that they pass, but there is a new failing Build step that seems to be related to GCP. Does this need to be resolved before merge?

I'm not sure why this was added, but it seems odd to have a GCP thing in our CI for a mobile app, so I would say no.  I'll look into where this check came from.

JacobJaffe (approved):
Confirmed dev archive local succeeds! Thanks!

PR #1135, opened 2020-06-29
Move eula checkbox to welcome screen

johnschoeman:
Why:
Currently the user flow for the initial welcome screen requires the user
to open the terms modal and then tap an agreement checkbox. We would
like to move the checkbox to the welcome screen itself and have the
modal only contain the terms content, per the current designs.

This commit:
Refactors Onboarding1 and EulaModal screen such that the checkbox can be
moved. While doing so we changed the name of Onboarding1 to Welcome and
converted the screen to typescript. We also updating some of the styling
layout and brought the pill button style into the main styles folder.

Co-Authored-By: ericwbailey <eric.w.bailey@gmail.com>

### Before

welcome-before

### After

welcome-after


app/Entry.js
```diff
@@ -40,7 +40,7 @@ import { ENLocalDiagnosisKeyScreen } from './views/Settings/ENLocalDiagnosisKeyS
 import { FeatureFlagsScreen } from './views/FeatureFlagToggles';
 import ImportScreen from './views/Import';
 import { EnableExposureNotifications } from './views/onboarding/EnableExposureNotifications';
-import Onboarding1 from './views/onboarding/Onboarding1';
+import Welcome from './views/onboarding/Welcome';
```
jeffstolz, line 43: I like this name change and will do the same for the other onboarding screens so they're more descriptive of the actual screen function

app/styles/buttons.ts
```diff
@@ -106,3 +112,10 @@ export const largeTransparent: ViewStyle = {
   ...large,
   ...transparent,
 };
+
+export const pill: ViewStyle = {
+  borderRadius: 100,
```
jeffstolz, line 117: I'm assuming this value is very specific to this component and that's why it's not in a variable

app/views/onboarding/EulaModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+      setHtml(await loadLocalResource(eulaPath));
+    };
+    loadEula();
+  }, [selectedLocale, setHtml, eulaPath]);
+
+  return (
+    <Modal animationType='slide' transparent visible={showModal}>
+      <View style={styles.container}>
+        <SafeAreaView style={{ flex: 1 }}>
+          <View style={{ flex: 7, paddingHorizontal: 5 }}>
```
jeffstolz: Any specific reason why you chose to define the flex values inline and not in the stylesheet?

jeffstolz (approved):
One question, otherwise looks good

app/styles/buttons.ts
devinjameson: It may make sense to either create a variable called `borderRadiusMax` and set it to a big number like `1000` or, if we can, just use `100%` here. Not sure if that works in RN world though. @jeffstolz The goal here is to get fully rounded corners.

app/views/onboarding/EulaModal.tsx
devinjameson: I somewhat understand the `flex: 1` being inline, but not sure about line 75.

app/views/onboarding/Welcome.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+  useEffect(() => {
+    getUserLocaleOverride().then((locale) => {
+      if (locale) {
+        setLocale(locale);
+      }
+    });
+  });
+
+  const toggleCheckbox = () => {
```
devinjameson, line 50: Is there a reason we are using strings instead of storing the agreement state in a boolean? Then we could do something like `setEulaAgreementAccepted(!eulaAgreementAccepted)`. Line 32 makes sense coming from Elm, but since there are only two states why not just true/false?

app/views/onboarding/Welcome.tsx
```diff
@@ -0,0 +1,247 @@
...
+                testID={'welcome-eula-checkbox'}
+                style={styles.checkbox}
+                onPress={toggleCheckbox}
+                accessible
+                accessibilityRole='checkbox'
+                accessibilityLabel={t('onboarding.eula_message')}>
+                <Image source={checkboxIcon} style={styles.checkboxIcon} />
+              </TouchableOpacity>
+
+              <View>
```
devinjameson, line 144: I wouldn't necessarily change anything here because I think it's just a downside of React, but to my eye this becomes really hard to read once we have all this nesting and indentation.

devinjameson (commented):
Nice! Few comments.

app/styles/buttons.ts
jeffstolz: Yeah I'd use `100%` here then - I think it's much more explicit about its purpose and more fail-safe

app/views/onboarding/Welcome.tsx
johnschoeman: totally, I go back n forth on this. We chatted irl, but for future readers: On the one hand, having true/false is significantly less code (potentially low signal to noise ratio code at that), on the other hand by providing a named type we better express the domain concept. `The user has agreed to the terms` is more clear than `the user has set the terms agreement state to true` and it also leads us open to future iterations where we might have more than just 2 states, for example, there might be a world were we want to force an active choice with something like: `type EulaAgreement = "UnAnswered" | "Agreed" | "Disagreed"` and having the union type now allows for this easily.

app/views/onboarding/EulaModal.tsx
johnschoeman: Just a refactor of existing code, didnt get around to cleaning up every last thing.

app/styles/buttons.ts
johnschoeman: totally agree, will tackle in another pr as to not have to go through all of ci again.

PR #1138, opened 2020-06-29
Fixed iOS run scripts by specifying configuration

jvisenti:
#### Description:

`react-native run-ios` uses `Debug` configuration unless otherwise specified, and #1134 removed our use of Debug config, so this PR updates our scripts to use the correct configuration.

#### How to test:

Ensure the following commands succeed:
`yarn run-ios-gps`
`yarn run-ios-bt`


johnschoeman (approved):

yonatanmk (approved):
Yup this fixes the those commands

PR #1139, opened 2020-06-29
Update styling of exposure history flow

devinjameson:
#### Description:
- Disables the selection of future dates in the calendar
- Updates styling on the exposure history screen

#### Linked issues:
https://trello.com/c/IEWxxQLX

#### Screenshots:
Screen Shot 2020-06-29 at 5 59 00 PM

#### How to test:
- Try to select a date in the future on the calendar and validate that you cannot
- Validate that the visual styles on the Exposure History screen match the mockups

Co-Authored-By: Jeff Stolz <jstolz123@gmail.com>
Co-Authored-By: Eric Bailey <eric.w.bailey@gmail.com>

### DIFF
--- a/app/views/ExposureHistory/Calendar.tsx
+++ b/app/views/ExposureHistory/Calendar.tsx
@@ -36,14 +36,14 @@ const Calendar = ({
     return (
       <View style={styles.calendarRow}>
         {week.map((datum: ExposureDatum) => {
+          const isSelected = datum.date === selectedDatum?.date;
+
           return (
             <TouchableOpacity
               key={`calendar-day-${datum.date}`}
               onPress={() => onSelectDate(datum)}>
               <ExposureDatumIndicator
-                isSelected={
-                  selectedDatum ? datum.date === selectedDatum.date : false
-                }
+                isSelected={isSelected}
                 exposureDatum={datum}
               />
             </TouchableOpacity>
@@ -60,7 +60,7 @@ const Calendar = ({
         {labels.map((label: string, idx: number) => {
           return (
             <View style={styles.labelStyle} key={`calendar-label-${idx}`}>
-              <Text>{label}</Text>
+              <Text style={styles.labelTextStyle}>{label}</Text>
             </View>
           );
         })}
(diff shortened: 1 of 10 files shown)


johnschoeman (approved):
dope. looks good to me homie.

just looks like you need to update the snapshots with `yarn update-snapshots`

app/views/ExposureHistory/Calendar.tsx
```diff
@@ -36,14 +40,15 @@ const Calendar = ({
...
       <View style={styles.calendarRow}>
         {week.map((datum: ExposureDatum) => {
+          const isSelected = datum.date === selectedDatum?.date;
+
           return (
             <TouchableOpacity
               key={`calendar-day-${datum.date}`}
-              onPress={() => onSelectDate(datum)}>
+              onPress={() => onSelectDate(datum)}
+              disabled={isAfterToday(datum.date)}>
```
johnschoeman: so this is a minor, extremely pedantic, point of view, but thought you might want to geek out on pedantic coding points of view like the good ol days.

Using the disabled prop here is mildly more imperative, rather than declarative, approach to this feature than putting the conditional in the `onSelectDate` function.  That is, the high level declaration of the behavior you want is 'when a user taps a date, if that date is not in the future, move the circle to that date.' So by adding isAfterToday(datum.date) to the disabled prop you are indicating the `how` to do this rather than the `what` you want.

Another way to think about it is you have two places where the 'what do I do when the user taps a thing' logic happens rather than one.

Of course, this is an extremely silly of thing to split hairs on, and i wouldnt take the time to change anything in this commit.

devinjameson: I love this stuff haha. Miss those days. I was actually wondering if this was the best place to house this logic. I guess it’s kinda pedantic in a way but I think it matters at scale. You do this 10 times and it adds up. I may take a look at this in the morning, maybe we could pair on it for a few minutes?

johnschoeman: totally! hit me up and we can take a look.

jeffstolz (approved):
I paired on this so we may want one more set of eyes, but LGTM

PR #1140, opened 2020-06-30
Connect intersection logic with Exposure History screen

vlasac:
#### Description:
This PR introduces logic that populates the Exposur History calendar on intersection check completion.

#### How to test:
- Select an item from the list of available HAs
- wait for the intersection logic to finish
- after some time, if you have been exposed, the data should be visible in the exposure history screen


app/gps/exposureInfo.ts
```diff
@@ -0,0 +1,59 @@
...
+  emit: (event: Event, data: ExposureInfo) => void;
+  addListener: (
+    event: Event,
+    callback: HandleOnExposureInfoUpdated,
+  ) => {
+    remove: () => void;
+  };
+}
+
+const ExposureEvents: GPSExposureInfoEventEmitter = {
```
johnschoeman, line 23: so this might want to be cleaned up a bit.  currently remove() , removes all of the subscribers, which maybe is _okay_ given we only expect to ever have one subscriber to this emitter, but this is not clearly indicated to the reader.

What do you think about changing this up a touch to make this more clear?

something like

```typescript
interface GPSExposureInfoEventEmitter {
  subscriber: Subscriber | null;
  emit: (event: Event, data: ExposureInfo) => void;
  addListener: (
    event: Event,
    callback: HandleOnExposureInfoUpdated,
  ) => {
    remove: () => void;
  };
}

const ExposureEvents: GPSExposureInfoEventEmitter = {
  subscriber: null,
  emit: function (event: Event, data: ExposureInfo) {
      if (this.subscriber) {
        this.subscriber.callback(data);
      }
  },
  addListener: function (event: Event, callback: HandleOnExposureInfoUpdated) {
    const nextSubscriber: Subscriber = {
      eventName: event,
      callback,
    };
    this.subscriber = nextSubscriber
    return {
      remove: () => {
        this.subscribers = null;
      },
    };
  },
};
```

app/helpers/Intersect.d.ts
```diff
@@ -1,7 +1,8 @@
 import { checkIntersect as intersect } from './Intersect';
 import { HealthcareAuthority } from '../store/types';
+import { ExposureDatum } from '../exposureHistory';
 
 export const checkIntersect: (
   healthcareAuthorities: HealthcareAuthority[] | null,
   bypassTimer: boolean,
-) => Promise<void> = intersect;
+) => Promise<Record<number, ExposureDatum>> = intersect;
```
johnschoeman:
```suggestion
type Posix = number
 
 ....
 
 
) => Promise<Record<Posix, ExposureDatum>> = intersect;
```

its helpful context for the reader to indicate that the number represents a posix value

app/helpers/Intersect.js
```diff
@@ -322,6 +322,30 @@ export function fillDayBins(
   return dayBins;
 }
 
+const transformDayBinsToExposureInfo = (dayBins) => {
```
johnschoeman, line 325: I know this interest.js file is quite large so it might not makes sense to convert the whole file, but what do you think about putting this guy into another ts file such that we can provide some type safety?

```typescript
const transformDayBinsToExposureInfo = (dayBins: DayBin[]): ExposureInfo => {
```

app/helpers/Intersect.js
```diff
@@ -322,6 +322,30 @@ export function fillDayBins(
...
+  return dayBins.reduce((ei, duration, index) => {
+    const startOfDayAgo = dayjs()
+      .startOf('day')
+      .subtract(index, 'day')
+      .valueOf();
+    ei[startOfDayAgo] =
```
johnschoeman: what do you think about using a full name here? `exposureInfo[startOfDayAgo]`

johnschoeman (approved):
Overall looks good to me! a few thoughts on making the types a bit tighter. Happy to pair on it if you are interested!

Also, what you do you think about adding a spec for `transformDayBinsToExposureInfo` ?

app/helpers/Intersect.js
vlasac: I opened a ticket in backlog to refactor all of intersect.js and convert it to typescript, but it isn't gonna be pulled into this sprint. 

I can add the type declaration for this method to the `intersect.d.ts` if that will solve this problem for now.

Also, as part of the intersection refactor, I planned on writing out some more detailed tests for the helpers and I can add that spec you mentioned in that PR.

app/helpers/Intersect.d.ts
vlasac: Agree, I'll add it now.

app/gps/exposureInfo.ts
vlasac: Sure thing, thanks for the tip!

app/services/IntersectService.ts
```diff
@@ -16,18 +35,36 @@ class IntersectService {
     }
     this.isServiceRunning = true;
 
-    intersect(healthcareAuthorities, bypassTimer).then(() => {
+    intersect(healthcareAuthorities, bypassTimer).then((result) => {
```
JacobJaffe: Can we change the name of `result` to what the data represents?

app/services/BackgroundTaskService.ts
```diff
@@ -1,41 +0,0 @@
...
-        startOnBoot: true,
-        requiredNetworkType: BackgroundFetch.NETWORK_TYPE_NONE, // Default
-        requiresCharging: false, // Default
-        requiresDeviceIdle: false, // Default
-        requiresBatteryNotLow: false, // Default
-        requiresStorageNotLow: false, // Default
-      },
-      async (taskId) => {
-        console.log('[js] Received background-fetch event: ', taskId);
-        IntersectService.checkIntersect(null, false);
```
JacobJaffe: After comments around adding the `NativeModules.AsyncStoreManager.trim()` to this, I think we're keeping this separate still

app/services/IntersectService.ts
```diff
@@ -1,28 +1,14 @@
...
-};
+import BackgroundTaskService from './BackgroundTaskService';
 
 class IntersectService {
   isServiceRunning = false;
   private nextJob: HealthcareAuthority[] | null = null;
 
   constructor() {
-    this.startBackgroundTask();
+    BackgroundTaskService.start();
```
JacobJaffe: This seems weird because of how they use each other -- 
can `BackgroundTaskService` be started from somewhere central in the app? Maybe App.js?

JacobJaffe (approved):

app/services/IntersectService.ts
vlasac: sure thing, want me to change this now or can we do it as part of the backfill work?

PR #1141, opened 2020-06-30
Fix Notification Service Import Warning

yonatanmk:
#### Description:
When navigating to the main screen of the GPS app there is an unhandled promise rejection warning. This fixes the import bug causing the warning.

#### Screenshots:
Screen Shot 2020-06-30 at 11 07 03 AM

#### How to test:
Navigate to the main screen. You should no longer see the warning in the console.


JacobJaffe (approved):
Great, thanks!

PR #1145, opened 2020-06-30
Refactored location backfilling on iOS

jvisenti:
#### Description:

This PR addresses three issues:
1. Location data is retained for more than 14 days under certain conditions
2. Backfilled points due to location logging gaps don't include hashes
3. Fixed `LocationService.js` to use the correct location provider for improved reliability

We aim to solve both of these issues by fixing up location data more aggressively. Locations are now backfilled (including hashes) when:
* The app is foregrounded
* A new location is received
* The background fetch runs (every 15 min)
Backfilled locations are only generated if there is a >5min gap between location updates.

Locations older than 14 days are trimmed when:
* The app is foregrounded
* The background fetch runs (every 15 min)

This PR also changes how we handle the "current location", saving it to the db even if the desired interval hasn't elapsed. The latest point is then updated with new location information when a newer location is received.

The maximum backfill interval was also reduced to 12hrs (down from 24hrs) to reduce impact of hashing in the event of a full-interval backfill.

#### Linked issues:

https://pathcheck.atlassian.net/browse/SAF-757
https://pathcheck.atlassian.net/browse/SAF-771

### DIFF
--- a/app/helpers/Intersect.js
+++ b/app/helpers/Intersect.js
@@ -15,12 +15,10 @@ import {
   DEFAULT_CONCERN_TIME_FRAME_MINUTES,
   DEFAULT_THRESHOLD_MATCH_PERCENT,
   MAX_EXPOSURE_WINDOW_DAYS,
-  MIN_CHECK_INTERSECT_INTERVAL,
 } from '../constants/history';
 import {
   AUTHORITY_SOURCE_SETTINGS,
   CROSSED_PATHS,
-  LAST_CHECKED,
 } from '../constants/storage';
 import { DEBUG_MODE } from '../constants/storage';
 import { GetStoreData, SetStoreData } from '../helpers/General';
@@ -355,16 +353,11 @@ function roundExposure(diffMS, gpsPeriodMS) {
  *        from the authority (e.g. the news url) since we get that in the same call.
  *        Ideally those should probably be broken up better, but for now leaving it alone.
  */
-export async function checkIntersect(
-  healthcareAuthorities,
-  bypassTimer = false,
-) {
+export async function checkIntersect(healthcareAuthorities) {
   console.log(
     `[intersect] tick entering on ${isPlatformiOS() ? 'iOS' : 'Android'}; `,
-    `is bypassing timer = ${bypassTimer}`,
   );
-
-  const result = await asyncCheckIntersect(healthcareAuthorities, bypassTimer);
+  const result = await asyncCheckIntersect(healthcareAuthorities);
   console.log(`[intersect] ${result ? 'completed' : 'skipped'}`);
   return result;
 }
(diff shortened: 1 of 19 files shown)


vlasac (approved):

ios/GPS/storage/Location.swift
```diff
@@ -11,16 +11,22 @@ import RealmSwift
...
+  enum Key: String {
+    case time
+    case latitude
+    case longitude
+    case source
+    case hashes
+  }
+
+  @objc
+  enum Source: Int {
```
johnschoeman, line 25: 💯

ios/COVIDSafePaths.xcodeproj/project.pbxproj
```diff
@@ -49,13 +49,9 @@
 		6858E21561F14E61A6A85280 /* IBMPlexSans-SemiBoldItalic.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 747AA5DBBC684319A8273F2C /* IBMPlexSans-SemiBoldItalic.ttf */; };
 		68D8970624809AA50091A254 /* Scrypt in Frameworks */ = {isa = PBXBuildFile; productRef = 68D8970524809AA50091A254 /* Scrypt */; };
 		68D8970C24809B4E0091A254 /* MAURLocation+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 68D8970B24809B4E0091A254 /* MAURLocation+Extension.swift */; };
-		68D8970D24809B4E0091A254 /* MAURLocation+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 68D8970B24809B4E0091A254 /* MAURLocation+Extension.swift */; };
```
johnschoeman, line 52: where these duplicates in the header paths? What lead to these being removed?

johnschoeman (approved):
Overall looks great to me, though my swift experience is limited. I like a lot of the cleanup that happened as well! It might makes sense to ping @mattThousand for ios layer reviews in general.

ios/COVIDSafePaths.xcodeproj/project.pbxproj
jvisenti: These files were also added to the test target, which is unnecessary and causes ambiguity if the symbols exist in both targets.

app/helpers/Intersect.js
```diff
@@ -396,7 +390,7 @@ export async function checkIntersect(
     hasMigratedOldData = true;
   }
 
-  const result = await asyncCheckIntersect(healthcareAuthorities, bypassTimer);
+  const result = await asyncCheckIntersect(healthcareAuthorities);
   console.log(`[intersect] ${result ? 'completed' : 'skipped'}`);
```
JacobJaffe: This shouldn't be able to skip now I think?  (It'll skip at background task level)

app/services/BackgroundTaskService.ts
```diff
@@ -21,11 +26,29 @@ class BackgroundTaskService {
         requiresStorageNotLow: false, // Default
       },
       async (taskId) => {
+        PushNotification.localNotification({
+          title: 'Running background task',
```
JacobJaffe: Can we remove these push notifications before merging live?

app/services/BackgroundTaskService.ts
```diff
@@ -21,11 +26,29 @@ class BackgroundTaskService {
...
+
+        if (Platform.OS === 'ios') {
+          NativeModules.SecureStorageManager.trimLocations();
+        }
+
+        const lastCheckedMs = Number(await GetStoreData(LAST_CHECKED));
+        if (dayjs().diff(lastCheckedMs, 'hour') >= 12) {
+          IntersectService.checkIntersect(null);
+          await SetStoreData(LAST_CHECKED, String(dayjs().valueOf()));
+        }
```
JacobJaffe, line 40: Can we move the 'skip' console log to here now? I think testing might be using that for ensuring runs, so that'll be good to keep consistent

PR #1146, opened 2020-06-30
Add base font family to all text

johnschoeman:
Why:
Currently we are only applying the product's font family, IBMPlexSans,
to text elements that have been given the bold style and others have the
default system font family. We would like for all of the fonts to be the
product font family.

Co-authored-by: devin jameson <devin@thoughtbot.com>


jeffstolz (approved):
Excellent! We were talking about a need for this yesterday during pairing

PR #1147, opened 2020-06-30
Add exposure history legend

johnschoeman:
Why:
We would like to provide the user a way to view a legend of what the
indicators on the exposure history screen mean.

This commit:
Introduces a legend modal to the exposure history screen.

Co-Authored-By: devin jameson <devin@thoughtbot.com>

Screen Shot 2020-06-30 at 4 24 05 PM

### DIFF
--- a/app/styles/iconography.ts
+++ b/app/styles/iconography.ts
@@ -1,7 +1,17 @@
-import { ViewStyle } from 'react-native';
+import { ViewStyle, TextStyle } from 'react-native';
 
 import * as Colors from './colors';
 import * as Spacing from './spacing';
+import * as Typography from './typography';
+
+export const smallIcon: ViewStyle = {
+  height: 30,
+  width: 30,
+  borderRadius: 100,
+  alignItems: 'center',
+  justifyContent: 'center',
+  marginVertical: Spacing.small,
+};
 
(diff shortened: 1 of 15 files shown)


app/styles/iconography.ts
```diff
@@ -1,7 +1,17 @@
...
+import { ViewStyle, TextStyle } from 'react-native';
 
 import * as Colors from './colors';
 import * as Spacing from './spacing';
+import * as Typography from './typography';
+
+export const smallIcon: ViewStyle = {
+  height: 30,
+  width: 30,
+  borderRadius: 100,
```
jeffstolz, line 10: perhaps set this to 100%?

jeffstolz (approved):
Approved, I just recommend checking this on the iPhone 8 before merging to make sure those absolute position values don't need to be adjusted

johnschoeman:
> Approved, I just recommend checking this on the iPhone 8 before merging to make sure those absolute position values don't need to be adjusted

by iPhone 8 do you mean iPhone11? if so then, yup we did so, just didnt post a pic.

app/styles/iconography.ts
johnschoeman: sadly border radius doesnt accept percentages.

PR #1150, opened 2020-07-01
Refactor self assessment screens

zhaccc:
#### Description:

- Refactor self-assessment components
- Separation of concerns on some reusable components
- Implement design changes

#### How to test:

Go to the self-assessment screen from the navbar.


app/components/Typography.tsx
```diff
@@ -5,7 +5,18 @@ import { useLanguageDirection } from '../locales/languages';
 
 import { Typography as TypographyStyles } from '../styles';
 
-type TypographicUse = 'headline1' | 'headline2' | 'body1' | 'body2' | 'body3';
+type TypographicUse =
```
johnschoeman, line 8: Are these all being used? it might be nice to move away from the `use` prop and just use the plain vanilla `style` prop provided by rn, so we dont have to maintain two separate patterns for styles.

jeffstolz (approved):
I think we should approve this and get it merged, although we'll want to do some refactoring to address the recent changes that have been applied to our designs and styling architecture. Rather than delay this PR further, I recommend we address them after merging.

app/views/assessment/EmergencyAssessment.js
```diff
@@ -0,0 +1,178 @@
...
+ * @typedef { import(".").SurveyOption } SurveyOption
+ */
+import {
+  QUESTION_KEY_AGREE,
+  SCREEN_TYPE_RADIO,
+  SCREEN_TYPE_EMERGENCY,
+} from './constants';
+import { useSurvey } from '../../helpers/CustomHooks';
+import { Info } from './Info';
+// import { Typography } from '../../components/Typography';
```
johnschoeman: remove comment?

app/views/assessment/components/Button.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
johnschoeman, line 1: is it possible to not have the assessment screen use a bespoke button component? could we be using the same button as the rest of the app?

app/views/assessment/endScreens/Caregiver.js
```diff
@@ -1,32 +1,37 @@
...
+import { Button } from '../components/Button';
 
 import { Colors } from '../../../styles';
 
 /** @type {React.FunctionComponent<{}>} */
 export const Caregiver = ({ navigation }) => {
-  let { t } = useTranslation();
-  let { completeRoute } = useContext(MetaContext);
+  const { t } = useTranslation();
+  const { completeRoute } = useContext(MetaContext);
```
johnschoeman: This MetaContext is pretty undescriptive it would be great to replace with something more focused in a future commit.

app/views/assessment/index.js
```diff
@@ -127,84 +139,66 @@ const Assessment = ({ navigation }) => {
...
-                ...screenOptions,
+                ...screenOptions(),
+                headerLeft: () => null,
                 headerRight: () => null,
-                headerStyle: {
-                  backgroundColor: Colors.primaryBackgroundFaintShade,
-                },
               }}
             />
+            <Stack.Screen
```
johnschoeman, line 159: Would it makes sense to have the navigator for the self assessment screen live with the other navigators? Or is there a good reason to keep it separate?

johnschoeman (approved):
Over all looks like a great step in the right direction. A few questions which i think would make more sense to capture in follow up prs. On that note, this is a pretty big PR in terms of changes, given how quickly this project moves and how many people are actively working on things it would be great to start breaking these down into smaller commits where possible.

app/views/assessment/components/Button.js
zhaccc: There was a discussion about this and we decided not to use that button yet since it was refactored and added to too often. That being said I would like to open a different PR to implement the button we use throughout the app.

app/components/Typography.tsx
zhaccc: These are being used but I would really like to step away from those and use `style` prop as suggested.

app/views/assessment/endScreens/Caregiver.js
zhaccc: Agreed, naming on this one is no good.

app/views/assessment/components/Button.js
johnschoeman: makes sense to me!

app/views/assessment/index.js
zhaccc: Hmm, I am not sure yet because the file contains a lot of logic specific for the assessment. I do think we should be able to do this once we step away from the wey question flow is handled right now, and remove some of that logic.

PR #1151, opened 2020-07-01
Show System Enable EN Dialog and Bubble Up Results to JS Layer

octohub:
#### Description:

Taps into native Android layer to display system level Exposure Notifications "Turn On" dialog. Next up I will need to reflect to the JS layer when it is enabled to all the UI to update accordingly.

#### Screenshots:

Screenshot_20200701-103352

#### How to test:

Install fresh version of the bt app. Go through onboarding and at the end click "Enable Exposure Notifications."


app/bt/nativeModule.ts
```diff
@@ -76,6 +76,7 @@ export const submitDiagnosisKeys = async (
   cb: (errorMessage: string, successMessage: string) => void,
 ): Promise<void> => {
   keySubmissionModule.postDiagnosisKeys(cb);
+  permissionsModule.requestExposureNotificationAuthorization(cb);
```
johnschoeman: Why is this request necessary in this function?  It seems a bit out of place to me here. should we be requesting permissions earlier in the flow?

android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/utils/CallbackMessages.java
```diff
@@ -0,0 +1,8 @@
+package covidsafepaths.bt.exposurenotifications.utils;
+
+public class CallbackMessages {
```
johnschoeman, line 3: whats the plan with these guys? do we need to be handling these cases on the JS side?

johnschoeman (approved):
Code looks good to me! though my java/android experience is limited.  A few questions on the logic flow and callback message intent

app/bt/nativeModule.ts
octohub: Great catch. That was a mistake. Removing.

android/app/src/bt/java/covidsafepaths/bt/exposurenotifications/utils/CallbackMessages.java
octohub: The idea is to have one place on the Android native layer that shows all the messages being bubbled up to the js layer. `String+Extensions.swift` on the iOS side I think would be the equivalent, but correct me if I'm wrong on that.

Right now I don't think we need to handle them on the JS side. For the `requestExposureNotificationAuthorization` I'm replicating iOS which sends back a `genericSuccess` message or an error string.

johnschoeman: got it, thanks for the context, makes a lot of sense to me.

PR #1152, opened 2020-07-01
Request notification permissions for bt app

johnschoeman:
Why:
We would like for the bt build of the app to request notification
permissions during onboarding as they are necessary for the Exposure
Notification flow.

This commit:
Adds the screen to the onboarding flow for the BT build. Note that we
needed to duplicate the component from the GPS build. Ideally a future
commit will refactor the flow such that it can be agnostic to the
tracing strategy and we can avoid this code duplication.

Co-Authored-By: mattthousand <matt@nicethings.io>

### DIFF
--- a/app/Entry.js
+++ b/app/Entry.js
@@ -31,6 +31,7 @@ import {
   ExportLocally,
 } from './views/Export';
 import { PublishConsent } from './bt/PositiveDiagnosis/PublishConsent';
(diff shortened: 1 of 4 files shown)


mattThousand (approved):

ArgoJRB (approved):
LGTM!

PR #1154, opened 2020-07-01
added health authorities constant to config file

ArgoJRB:
added a constant for health authority name to authorities.ts, and used it to replace static content on next steps and sharing screens

#### Linked issues:

https://trello.com/c/L4pNhsN4/71-as-a-dev-team-we-have-a-way-to-customize-the-health-authority-name-in-the-app

#### Screenshots:

Screen Shot 2020-07-01 at 2 48 03 PM

#### How to test:

Change the name of the health authority in authorities.ts - make sure the new name is in the "published anonymized data" copy on the sharing screen the end of the health questionnaire, and one the next steps screen.

### DIFF
--- a/app/constants/authorities.ts
+++ b/app/constants/authorities.ts
@@ -3,3 +3,5 @@ export const PUBLIC_DATA_URL =
 
 export const AUTHORITIES_LIST_URL =
   'https://raw.githubusercontent.com/Path-Check/safeplaces-frontend/develop/healthcare-authorities.yaml';
+
+export const AUTHORITY_NAME = 'Boston Public Health Commission';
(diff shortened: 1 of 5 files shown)


app/constants/authorities.ts
```diff
@@ -3,3 +3,5 @@ export const PUBLIC_DATA_URL =
 
 export const AUTHORITIES_LIST_URL =
   'https://raw.githubusercontent.com/Path-Check/safeplaces-frontend/develop/healthcare-authorities.yaml';
+
+export const AUTHORITY_NAME = 'Boston Public Health Commission';
```
johnschoeman, line 7: Can we hook this up to a env variable? It would be nice to be able to update this without needing to make a merge into the codebase.

johnschoeman (approved):
One thought, might make sense to merge this and grab it in a follow up for sake of time though!

app/constants/authorities.ts
ArgoJRB: I think that wouldn't be too hard - but yeah, wouldn't hurt to merge this so you can start using it for test builds while I add that

PR #1155, opened 2020-07-01
iOS Exposure Detection Improvements

mattThousand:
### Why
We'd like to detect exposures successfully using keys from our GAEN server.

### This Commit
This commit introduces a `DOWNLOAD_PATH` environment variable that is necessary for parsing the index file from the GAEN server, and modifies error handling such that a more descriptive error message is surfaced to the user from the debug menu.

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -8,7 +8,7 @@ import BackgroundTasks
 final class ExposureManager: NSObject {
   
   @objc static let shared = ExposureManager()
-
+  
   private static let backgroundTaskIdentifier = Bundle.main.bundleIdentifier! + ".exposure-notification"
   
   let manager = ENManager()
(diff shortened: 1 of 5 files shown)


ios/BT/ExposureManager.swift
```diff
@@ -142,58 +142,66 @@ final class ExposureManager: NSObject {
...
       // Notify the user if bluetooth is off
       self?.notifyUserBlueToothOffIfNeeded()
-
+      
       // Perform the exposure detection
-      let progress = ExposureManager.shared.detectExposures { success in
-        task.setTaskCompleted(success: success)
+      let progress = ExposureManager.shared.detectExposures { error in
+        if error != nil {
+          task.setTaskCompleted(success: false)
```
johnschoeman, line 193: Should we be forwarding this error as a message?

johnschoeman (approved):
Lgtm! A question.

PR #1159, opened 2020-07-01
Refactor OnboardingTemplate component and use it on a few more screens

devinjameson:
#### Description:
- Refactors the `OnboardingTemplate` component to be more configurable and renames it to `ExplanationScreen`
- Uses the new `ExplanationScreen` component on a few more screens in the onboarding flow

#### Linked issues:
https://trello.com/c/JsfKPfSg/142-as-a-developer-i-would-like-a-shared-template-that-i-can-implement-for-the-similarly-styled-onboarding-export-and-self-assessmen

#### Screenshots:
demo

#### How to test:
Go through the onboarding flow and verify the screens looks correct.

Co-authored-by: Jeff Stolz <jstolz123@gmail.com>
Co-authored-by: Eric Bailey <eric.w.bailey@gmail.com>

### DIFF
--- a/app/bt/NotificationPermissionsBT.tsx
+++ b/app/bt/NotificationPermissionsBT.tsx
@@ -38,24 +38,24 @@ const NotificationsPermissions = (): JSX.Element => {
   return (
     <DescriptionTemplate
       iconXml={Icons.Bell}
-      title={t('onboarding.notification_header')}
-      titleStyle={styles.titleStyle}
+      header={t('onboarding.notification_header')}
+      headerStyle={styles.header}
       body={t('onboarding.notification_subheader')}
-      bodyStyle={styles.bodyStyle}
+      bodyStyle={styles.body}
       primaryButtonLabel={t('label.launch_enable_notif')}
       primaryButtonOnPress={handleOnPressEnable}
       secondaryButtonLabel={t('onboarding.maybe_later')}
       secondaryButtonOnPress={handleOnPressMaybeLater}
-      background={Images.BlueGradientBackground}
+      backgroundImage={Images.BlueGradientBackground}
     />
   );
 };
 
 const styles = StyleSheet.create({
-  titleStyle: {
+  header: {
     color: Colors.white,
   },
-  bodyStyle: {
+  body: {
     color: Colors.white,
   },
 });
--- a/app/bt/NotificationPermissionsBT.tsx
+++ b/app/bt/NotificationPermissionsBT.tsx
@@ -6,7 +6,9 @@ import { StyleSheet } from 'react-native';
 import PermissionsContext from './PermissionsContext';
 import { Screens } from '../navigation';
 import { useStatusBarEffect } from '../navigation';
-import DescriptionTemplate from '../views/common/DescriptionTemplate';
+import ExplanationScreen, {
+  IconStyle,
+} from '../views/common/ExplanationScreen';
 
 import { Icons, Images } from '../assets';
 import { Colors } from '../styles';
@@ -47,6 +49,7 @@ const NotificationsPermissions = (): JSX.Element => {
   const descriptionTemplateStyles = {
     headerStyle: styles.header,
     bodyStyle: styles.body,
+    iconStyle: IconStyle.Blue,
   };
 
   const descriptionTemplateActions = {
@@ -55,7 +58,7 @@ const NotificationsPermissions = (): JSX.Element => {
   };
 
   return (
-    <DescriptionTemplate
+    <ExplanationScreen
       descriptionTemplateContent={descriptionTemplateContent}
       descriptionTemplateStyles={descriptionTemplateStyles}
       descriptionTemplateActions={descriptionTemplateActions}
--- a/app/bt/NotificationPermissionsBT.tsx
+++ b/app/bt/NotificationPermissionsBT.tsx
@@ -37,7 +37,7 @@ const NotificationsPermissions = (): JSX.Element => {
 
   const descriptionTemplateContent = {
     backgroundImage: Images.BlueGradientBackground,
-    iconXml: Icons.Bell,
+    icon: Icons.Bell,
     header: t('onboarding.notification_header'),
     body: t('onboarding.notification_subheader'),
     primaryButtonLabel: t('label.launch_enable_notif'),
--- a/app/bt/NotificationPermissionsBT.tsx
+++ b/app/bt/NotificationPermissionsBT.tsx
@@ -1,28 +1,15 @@
 import React, { useContext } from 'react';
-import {
-  ImageBackground,
-  View,
-  ScrollView,
-  StyleSheet,
-  TouchableOpacity,
-} from 'react-native';
 import { useTranslation } from 'react-i18next';
 import { useNavigation } from '@react-navigation/native';
-import { SvgXml } from 'react-native-svg';
+import { StyleSheet } from 'react-native';
 
 import PermissionsContext from './PermissionsContext';
 import { Screens } from '../navigation';
-import { Typography } from '../components/Typography';
 import { useStatusBarEffect } from '../navigation';
+import DescriptionTemplate from '../views/common/DescriptionTemplate';
 
 import { Icons, Images } from '../assets';
-import {
-  Buttons,
-  Spacing,
-  Colors,
-  Iconography,
-  Typography as TypographyStyles,
-} from '../styles';
+import { Colors } from '../styles';
 
 const NotificationsPermissions = (): JSX.Element => {
   const navigation = useNavigation();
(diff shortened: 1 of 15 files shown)


app/tracingStrategy.ts
```diff
@@ -13,11 +15,11 @@ export interface TracingStrategy {
 }
 
 export interface StrategyAssets {
-  personalPrivacyBackground: string;
+  personalPrivacyBackground: ImageSourcePropType;
```
devinjameson, line 18: This is the correct type for an image source. [See React Native docs.](https://reactnative.dev/docs/0.32/image#source)

app/views/common/DescriptionTemplate.tsx
```diff
@@ -0,0 +1,174 @@
...
+  primaryButtonContainerStyle?: ViewStyle;
+  primaryButtonTextStyle?: TextStyle;
+  primaryButtonOnPress: () => void;
+  secondaryButtonLabel?: string;
+  secondaryButtonContainerStyle?: ViewStyle;
+  secondaryButtonTextStyle?: TextStyle;
+  secondaryButtonOnPress?: () => void;
+  backgroundImage: ImageSourcePropType;
+  backgroundStyle?: ViewStyle;
+  invertIcon?: boolean;
```
devinjameson: I decided not to touch this, but the name seems wrong to me because setting this to true doesn't exactly "invert" the icon. Perhaps something like `useGoldIcon` would be more clear.

app/views/common/DescriptionTemplate.tsx
```diff
@@ -0,0 +1,174 @@
...
+  secondaryButtonOnPress,
+  backgroundImage,
+  backgroundStyle,
+  invertIcon,
+}: DescriptionTemplateProps): JSX.Element => {
+  useStatusBarEffect('dark-content');
+
+  const iconStyle = invertIcon ? styles.goldIcon : styles.blueIcon;
+
+  const primaryButtonTextStyles = {
```
devinjameson: We have to do this instead of inlining `style={[styles.primaryButtonText, primaryButtonTextStyle}]` because the `Typography` component's `style` prop expects a `TextStyle`.

app/views/common/DescriptionTemplate.tsx
```diff
@@ -0,0 +1,174 @@
...
+  };
+
+  const contentStyles = {
+    ...styles.contentText,
+    ...bodyStyle,
+  };
+
+  return (
+    <View style={styles.outerContainer}>
+      {Layout.screenWidth <= Layout.smallScreenWidth ? null : (
```
devinjameson: This strikes me as something that should be pulled out into a constant, but I'm not sure exactly what it's doing.

app/views/common/DescriptionTemplate.tsx
```diff
@@ -0,0 +1,174 @@
...
+          style={[styles.background, backgroundStyle]}
+        />
+      )}
+      <View style={styles.content}>
+        <ScrollView
+          alwaysBounceVertical={false}
+          style={styles.innerContainer}
+          contentContainerStyle={{ paddingBottom: Spacing.large }}>
+          <View style={iconStyle}>
+            <SvgXml xml={iconXml} />
```
devinjameson: We can run into weird behavior here that is not intuitive to debug if we pass in a string that doesn't match an icon that exists in our assets. Perhaps we should add some type safety here.

app/views/onboarding/NotificationDetails.tsx
```diff
@@ -3,22 +3,27 @@ import { useTranslation } from 'react-i18next';
 import { isGPS } from '../../COVIDSafePathsConfig';
 
 import { useStrategyContent } from '../../TracingStrategyContext';
-import OnboardingTemplate from './OnboardingTemplate';
+import DescriptionTemplate from '../common/DescriptionTemplate';
 
-const NotificationDetails = (props) => {
+interface NotificationDetailsProps {
+  navigation: any;
```
devinjameson, line 9: I think its best to be explicit about the fact that we don't have any type safety around this.

johnschoeman: is that true though? could we be using the `useNavigation` hook?

devinjameson: Ah I’ll check that out thanks!

devinjameson: Do you have time to pair on this @johnschoeman?

johnschoeman: have a few more meetings today, but we can sync up early next week!

app/views/common/ExplanationScreen.tsx
```diff
@@ -0,0 +1,197 @@
...
+            <SvgXml xml={explanationScreenContent.icon} />
+          </View>
+          <Typography style={headerStyles}>
+            {explanationScreenContent.header}
+          </Typography>
+          <Typography style={contentStyles}>
+            {explanationScreenContent.body}
+          </Typography>
+        </ScrollView>
+        <TouchableOpacity
```
JacobJaffe: I don't think this should be a `TouchableOpacity`, but a `Button` component -- these buttons need to communicate more states on some screens, like loading, which they currently do, e.g. in our workflow on ExportTemplate (Which it looks like to me that this will encapsulate by generalizing the templates?)

```js
          <Button
            label={nextButtonLabel}
            onPress={onNext}
            loading={buttonLoading}
          />
```

JacobJaffe (changes requested):

app/views/common/ExplanationScreen.tsx
devinjameson: Hey @JacobJaffe, can you clarify why this should be a `Button`? I almost always use `TouchableOpacity` instead of `Button` because `Button` doesn't allow for custom styling. [See React Native docs.](https://reactnative.dev/docs/button)

From the linked article:
> If this button doesn't look right for your app, you can build your own button using TouchableOpacity or TouchableNativeFeedback.

devinjameson: Re: `ExportTemplate`, how do you feel about getting this in place first and perhaps making this configurable for additional screens in the future if it makes sense?

app/bt/NotificationPermissionsBT.tsx
```diff
@@ -48,79 +37,42 @@ const NotificationsPermissions = (): JSX.Element => {
...
-        </TouchableOpacity>
-      </View>
-    </ImageBackground>
+  const explanationScreenActions = {
+    primaryButtonOnPress: handleOnPressEnable,
+    secondaryButtonOnPress: handleOnPressMaybeLater,
+  };
+
+  return (
+    <ExplanationScreen
```
johnschoeman, line 61: Nice this reads a lot better to me!

app/views/common/ExplanationScreen.tsx
```diff
@@ -0,0 +1,197 @@
...
+}
+
+const ExplanationScreen = ({
+  explanationScreenContent,
+  explanationScreenStyles,
+  explanationScreenActions,
+}: ExplanationScreenProps): JSX.Element => {
+  useStatusBarEffect('dark-content');
+
+  function getIconStyle(iconStyle: IconStyle) {
```
johnschoeman: does this want to be the es6 syntax? e.g. 
```
const determineIconStyle = (iconStyle: IconStyle): ViewStyle => {
```

also as a personal preference, I try to avoid `get` in function names unless I am intending to use it as a getter method on an object.

app/views/common/ExplanationScreen.tsx
johnschoeman: I would lean towards not being overly DRY in the view layer in general, but for the buttons it makes sense to me to introduce something like `PrimaryButton` and `SecondaryButton` as components. So long as we make sure to not conflate styles, markup, view logic and, business logic. That is, the button component should only handle the view logic and markup. we should be able to have an interface thats ergonomic for both designers and developers.  

But I agree with you at this point dev. it makes sense to ship this and spend the extra time making it configurable in the future if it is necessary.

JacobJaffe: Ah, to clarify I mean our `components/button`

JacobJaffe: > Re: ExportTemplate, how do you feel about getting this in place first and perhaps making this configurable for additional screens in the future if it makes sense?

Sure, I just want to ensure we're not introducing regressions in the future with these migrations from our button component

johnschoeman (approved):
Looking good overall! One thing that jumps out at me is that its getting to be a bigish PR, which is fine, but wanted to point out that in general itll be worth the effort to keep some of these refactoring prs smaller as itll be easy to let things get quite large.

app/views/common/ExplanationScreen.tsx
devinjameson: Ah I see @JacobJaffe, the latest commit refactors to use the custom `Button` component.

app/views/common/ExplanationScreen.tsx
devinjameson: Good call thank you!

JacobJaffe (approved):
🎉 LGTM! apologizes for button confusion

app/views/common/ExplanationScreen.tsx
```diff
@@ -136,14 +136,15 @@ const ExplanationScreen = ({
...
         />
-        {explanationScreenContent.secondaryButtonLabel && (
-          <Button
-            label={explanationScreenContent.secondaryButtonLabel}
-            onPress={explanationScreenActions.secondaryButtonOnPress}
-            style={secondaryButtonStyles}
-            textStyle={secondaryButtonTextStyles}
-          />
-        )}
+        {explanationScreenActions.secondaryButtonOnPress &&
```
devinjameson, line 139: @johnschoeman I'm going to merge this in because it's been open for a while, but I'm curious if there is a better way to do this check. I tried to pull it out into a constant but the check stopped working.

devinjameson:
> 🎉 LGTM! apologizes for button confusiono

No prob!

PR #1161, opened 2020-07-02
(iOS) Schedule background task on app launch.

mattThousand:
### Why
We'd like the app to detect exposures at regular intervals while the app is in the background

### This Commit
This commit schedules the initial background exposure detection task on app launch


johnschoeman (approved):
Lgtm!

PR #1163, opened 2020-07-02
[SAF-736] - Separate exposure history calculation between GPS and BT

vlasac:
#### Description:
This PR addresses the SAF-736 ticket requirement that when no data is recorded on a day, it should be presented to the user with a different message. This PR also includes additional work for splitting the exposure history calculation between GPS and BT.

NOTE: This PR is still a bit of a WIP, as I need to find a way to pass the GPS-specific initial set of data to the exposure screen. The initial state is currently set to match the BT logic.

#### Screenshots:
image

#### How to test:
- Initiate the intersection logic (select any HA from the list of available)
- Wait for the intersection to finish
- Go to Exposure History screen and verify


app/bt/exposureNotifications.ts
```diff
@@ -13,9 +19,27 @@ export interface RawExposure {
...
+        date,
+      };
+    }
+  });
+};
+
 export const toExposureInfo = (rawExposures: RawExposure[]): ExposureInfo => {
-  const possibleExposures = rawExposures.map(toPossible);
-  return groupedByDate(possibleExposures, combinePossibles);
+  return rawExposures.map(toPossible).reduce(groupByDate, {});
```
johnschoeman, line 41: nice.

app/exposureHistory.ts
```diff
@@ -15,30 +15,33 @@ export interface NoKnown {
...
-      return exposureInfo[date];
-    } else {
-      return {
-        kind: 'NoKnown',
-        date,
-      };
-    }
-  });
+export interface ExposureCalendarOptions {
+  initDate: Posix;
```
johnschoeman: what do you think about `startDate`, or `todaysDate`, its not clear to me what `init` stands for here? is it when we start counting? or is it the lastDay, is it todays date? it would be nice have this clear in the name.

app/factories/rawExposure.tsx
```diff
@@ -0,0 +1,15 @@
+import { Factory } from 'fishery';
+
+import { daysAgo, beginningOfDay } from '../helpers/dateTimeUtils';
+import { RawExposure } from '../bt/exposureNotifications';
+
+export default Factory.define<RawExposure>(() => {
```
johnschoeman, line 6: nice! thanks for adding a test factory.

app/gps/intersect/toExposureInfo.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+
+    describe('when given the DayBins with positive duration today and 2 days ago', () => {
+      it('returns an exposure history with the exposures on the correct days', () => {
+        const tuesday = dayjs('2020-06-23').valueOf();
+        const friday = dayjs('2020-06-26').valueOf();
+
+        const today = friday;
+
+        const dayBins: DayBins = [300000, -1, 0, 600000, 0];
```
johnschoeman: super helpful to see some data for this. as a reader not familiar with the gps flow, i find `DayBins` not super clear.

That said, what do you think typing this like

```typescript
type IntersectionDuration = number
type DayBin = IntersectionDuration[]

const dayBins: DayBin = [30000, -1, 0]
```

or something like that. it probably makes sense in a different commit, but in general, what exactly is a dayBin? and can we add more type safety here than a list of numbers?

app/gps/intersect/toExposureHistory.ts
```diff
@@ -0,0 +1,60 @@
+import {
```
johnschoeman: I might call this file, `exposureHistory` rather than `toExposureHistory` its better to be grouping code by domain concept rather than function name

johnschoeman (approved):
Nice, this is looking really good to me! a few minor comments.

I also dig the introduction of the `toExposureHistory` to both tracing strategies. Though i think i still might makes sense to have the emitters emit `ExposureInfo` and have the conversion to a user facing `ExposureHistory` done as a separate step. This will allow for us to do other things in the pipeline if we need to in the future.

app/bt/exposureNotifications.ts
vlasac: thanks! :)

app/TracingStrategyContext.tsx
```diff
@@ -44,7 +44,9 @@ export const TracingStrategyProvider = ({
       }}>
       <StrategyPermissionsProvider>
         <ExposureHistoryProvider
-          exposureInfoSubscription={strategy.exposureInfoSubscription}>
+          exposureEventsStrategyContext={
```
johnschoeman: What do you think about:
```suggestion
          exposureEventsStrategy={
```
?
`Context` has a bit of name collision imo. This is less of a context and more of a prop passed to the Context?

app/views/ExposureHistory/History.spec.tsx
```diff
@@ -33,15 +33,14 @@ describe('History', () => {
         const twoDaysAgo = DateTimeUtils.beginningOfDay(
           DateTimeUtils.daysAgo(2),
         );
-        const datum = factories.exposureDatum.build({
-          kind: 'Possible',
+        const rawExposure = factories.rawExposure.build({
+          id: 'Possible',
```
johnschoeman: should this be `kind` here?

johnschoeman (approved):
Overall looks solid to me! a few minor things i noticed. Awesome work!

PR #1164, opened 2020-07-02
.env setup script for bt

mattThousand:
### Why
We'd like developers in the `Path-Check` org to be able to easily populate env files with the values they need for individual HA's.

### This Commit
This commit creates a `yarn` script that enables developers to populate their local `bt` `.env` files by running the command `yarn set-config pc`


.github/workflows/build.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           ref: master
 
       - name: Setup env
-        run: sh ./setup_env.sh
+        run: sh ./setup_env.sh github
```
mattThousand, line 29: @johnschoeman the `build` workflow now passes `github` to the setup script, which is a no-op

.env.bt.example
```diff
@@ -0,0 +1,5 @@
+TRACING_STRATEGY=bt
```
johnschoeman: What do you think about example.env.bt?  Will make the git ignore simpler

README.md
```diff
@@ -117,6 +117,18 @@ dev_setup.sh
 dev_setup.bat
 ```
 
+#### Environment
+
+Populate the following 3 `.env` files with the relevant urls for your GAEN server:
+
+```
+.env.bt
+.env.bt.staging
```
johnschoeman: We might want to drop the staging ?

app/views/ExposureHistory/__snapshots__/Calendar.spec.tsx.snap
```diff
@@ -942,7 +941,8 @@ exports[`Calendar renders 1`] = `
             style={
               Object {
                 "alignItems": "center",
-                "borderColor": "rgba(0, 0, 0, 0)",
```
johnschoeman: Probably not related to your pr, but something is wonky with this snapshot

package.json
```diff
@@ -18,6 +18,7 @@
     "install:pod": "cd ios && bundle install --quiet && bundle exec pod install --silent",
     "postinstall": "./scripts/install-submodules.sh node_modules/\\@mauron85/react-native-background-geolocation && patch-package && node ./scripts/replace.js && npx react-native-jetifier",
     "postversion": "react-native-version",
+    "set-config": "./setup_env.sh",
```
johnschoeman: What do you think about set-env or set-ha?

setup_env.sh
```diff
@@ -1,7 +1,33 @@
...
-git update-index --assume-unchanged .env.bt.release
\ No newline at end of file
+git submodule update --remote --merge
+
+if [[ $# -eq 0 ]] ; then
+    tput setaf 1; echo 'You must supply a config argument'
+    exit 1
+fi
+
+if [ $DOWNLOAD_PATH != github ]; then
```
johnschoeman: Could we flatten out this nesting with an if then else fi?

setup_env.sh
```diff
@@ -1,7 +1,33 @@
 #!/bin/bash                                                                                                         
 
-cp -R pathcheck-mobile-resources/environment/ ./
+DOWNLOAD_PATH=$1
```
johnschoeman: Is download path the best name here? Would authority_name be better?

johnschoeman (approved):
Thanks for taking this on! A few questions.

app/views/ExposureHistory/__snapshots__/Calendar.spec.tsx.snap
mattThousand: indeed; there's something non-deterministic going on 🤔

setup_env.sh
mattThousand: hm, maybe `HA`? It's used in the interaction between the GAEN server and the app (the GAEN server prefixes the url path of the key archives with this value), but I agree, since it's tied to the name of the HA, it would be clearer if that were reflected in the naming

app/views/ExposureHistory/__snapshots__/Calendar.spec.tsx.snap
mattThousand: Perhaps there's something on this screen that renders a certain way depending on the current date/time?

README.md
```diff
@@ -117,6 +117,17 @@ dev_setup.sh
 dev_setup.bat
 ```
 
+#### Environment
+
+Populate the following 3 `.env` files with the relevant urls for your GAEN server:
```
JacobJaffe:
```suggestion
Populate the following 2 `.env` files with the relevant urls for your GAEN server:
```

README.md
```diff
@@ -117,6 +117,17 @@ dev_setup.sh
...
+
+```
+.env.bt
+.env.bt.release
+```
+
+**Note:** Members of the `Path-Check` org can complete this step by running `yarn set-ha` and passing in the 2-letter ha abbreviation as the first argument (i.e. `yarn set-ha pc`)
```
JacobJaffe, line 129: Questions for my understanding:
1. is "pc" HA "path check" here? 
2. What will be the strategy for knowing what HAs are available?

mattThousand: 1. Yes, "pc" HA  is "pathcheck" - this configuration will ideally point to its own gaen server instance in the future  
2. I'm thinking that information should be private (but stored in place where all members of the Path Check org know where to look for it). At the moment, members of the org can look here https://github.com/Path-Check/pathcheck-mobile-resources/tree/master/environment

PR #1165, opened 2020-07-02
Implement eight digit code for UI

zhaccc:
#### Description:

Implement support for 8 digit code instead of 6 digit code.

The design should take a look at this because since we use 8 text inputs it now looks kinda squashed (check the screens below).

#### Screenshots:

_iPhone SE (1st Generation):_

_iPhone 11 Pro Max:_


app/views/Export/ExportCodeInput.js
```diff
@@ -24,7 +24,7 @@ import { useStrategyContent } from '../../TracingStrategyContext';
 import { Icons } from '../../assets';
 import { Colors } from '../../styles';
 
-const CODE_LENGTH = 6;
+const CODE_LENGTH = 8;
```
JacobJaffe: This should only be for BT I believe

JacobJaffe (changes requested):

app/views/Export/ExportCodeInput.js
zhaccc: Yep, as we discussed.

johnschoeman (approved):
LGTM, would appreciate your review on this one: https://github.com/Path-Check/covid-safe-paths/pull/1174 ! as it moves away from sharing this screen at this point.

PR #1166, opened 2020-07-02
Remove exposure likely from history legend

johnschoeman:
Why:
We currently don't handle the exposure is likely indicator for the
Exposure history screen so we would like to remove it from the legend.
We plan to add something similar back when we have that data to present
to the user.

Screen Shot 2020-07-02 at 12 03 13 PM


jeffstolz (approved):
LGTM

devinjameson (approved):
Looks good.

PR #1170, opened 2020-07-03
Remove calendar snapshot spec

johnschoeman:
Why:
Currently we have a snapshot spec for a component which renders
differently based off of the current date. This leads to an intermittent
test failure.

This commit:
Removes the snapshot spec and replaces it with a basic spec to test that
the component renders without raising. Future commits should include
more useful specs.


octohub (approved):
Thank you sir

PR #1171, opened 2020-07-03
Add spec coverage for possible exposure indicator

johnschoeman:
Why:
Currently the feature of showing a 'Next Steps' button on the
ExposureHistory screen has not test coverage. As this is an important
part of the user flow, we would like to add some basic test coverage.

### DIFF
--- a/app/views/ExposureHistory/History.spec.tsx
+++ b/app/views/ExposureHistory/History.spec.tsx
@@ -1,0 +1,73 @@
+import React from 'react';
+import {
+  fireEvent,
+  wait,
+  cleanup,
+  render,
+} from '@testing-library/react-native';
+
+import { calendarDays, toExposureHistory } from '../../exposureHistory';
+import { DateTimeUtils } from '../../helpers';
+import { factories } from '../../factories';
+
+import History from './History';
+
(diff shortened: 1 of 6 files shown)


app/views/ExposureHistory/History.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+    });
+  });
+});
+
+const buildBlankExposureHistory = () => {
+  const datum = factories.exposureDatum.build();
+  const exposureInfo = {
+    [datum.date]: datum,
+  };
+  const calendar = calendarDays(Date.now(), 21);
```
octohub: Perhaps put `21` in a constant because it's used in multiple places.

octohub (approved):
Just one minor comment, great work. And assuming all the checks pass it's good to merge.

app/views/ExposureHistory/History.spec.tsx
johnschoeman: totally agree. also not a huge fan of how the 21 is floating around in general. I feel its usual to keep separate, but it feels like there wants to be a function that pre-applies this.

PR #1172, opened 2020-07-04
Add Spec Coverage for ExportCodeInput screen.

mattThousand:
### Why:
Currently the feature of enabling/disabling the `Next` button on the
`ExportCodeInput` screen depending on the length of user input does not have test coverage. As this is an important part of the user flow, we would like to add some basic test coverage.

### This Commit
This commit adds tests for the above, and also converts related files to TypeScript.

85642488-a7b49700-b65f-11ea-92f2-39bc50c4bf1b


yarn.lock
```diff
@@ -1274,7 +1274,6 @@
 "@testing-library/jest-native@^3.1.0":
   version "3.1.0"
   resolved "https://registry.yarnpkg.com/@testing-library/jest-native/-/jest-native-3.1.0.tgz#414c834dfb524ca82b86509d5479e365d4d5bbec"
-  integrity sha512-MZdsTsD6xrP5V0FTz2gVPG00ZMoKn4WniRitUH40zZTQ0DfMa5CVDpkAd5ed6qX1coDwS8vSCZS/JGMIDYjRQQ==
```
octohub: Should these integrity values be removed?

octohub (approved):
Looks solid, I would like @johnschoeman to take a look as well given his domain expertise.

app/views/__tests__/ExportCodeInput.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+import {
+  NavigationScreenProp,
+  NavigationRoute,
+  NavigationState,
+  NavigationParams,
+} from 'react-navigation';
+import { render, cleanup } from '@testing-library/react-native';
+import '@testing-library/jest-native/extend-expect';
+import { fireEvent } from '@testing-library/react-native';
+import { TracingStrategy } from '../../../app/tracingStrategy';
```
johnschoeman: I might group these imports such that third party code is separated from our code by a line break

app/views/__tests__/ExportCodeInput.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+interface NavProps {
+  navigation: NavigationScreenProp<NavigationState, NavigationParams>;
+  route: NavigationRoute;
+}
+
+afterEach(cleanup);
+
+describe('<ExportCodeInputScreen />', () => {
+  beforeEach(() => {});
```
johnschoeman: is this beforeEach necessary?

app/views/__tests__/ExportCodeInput.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+  it('when there is no text input', () => {
```
johnschoeman, line 34:
```suggestion
  describe('when there is no text input', () => {
     it('renders with the next button disabled', () => {
     
```

app/views/__tests__/ExportCodeInput.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    for (let i = 0; i <= 5; i++) {
+      const input = getByTestId(`input${i}`);
+      fireEvent.changeText(input, '1');
+    }
+    const button = getByTestId('next-button');
+    expect(button).not.toBeDisabled();
+  });
+});
+
+const renderTracingStrategyProvider = (strategy: TracingStrategy) => {
```
johnschoeman, line 70: This totally works, but is a bit of a smell to me. I feel like the tests would be better Unit tests if the CodeInput was less couple to the data model and we could pass in all of the data required as a prop.  so basically decouple the notion of a `Screen` which know how to handle navigation and application IO stuff and a `CodeInput` which know how to validate code inputs from a user.

johnschoeman (approved):
LGTM, a few comments

app/views/__tests__/ExportCodeInput.spec.tsx
mattThousand: I think I follow but just want to make sure; are you envisioning 2 separate components, an `ExportScreen` and a `CodeInput`?

johnschoeman: yup something like that! I wouldnt say that is necessary for this commit, but I think that would a way to make things a bit more modular.

PR #1173, opened 2020-07-06
Limit Exposure Detection to at most once every 3 hours

mattThousand:
### Why
The `Exposure Detection` framework will error out if keys are submitted >20 times per day.

### This Commit
This commit ensures that exposure detection will be run at most 8 times per day. On iOS we do not have the ability to schedule the exact interval at which our background task is run, but we can make sure that when it is run, we check that at least 3 hours have elapsed since the last execution before submitting keys to the framework.

IMG_1913 2

### DIFF
--- a/ios/BT/ExposureManager.swift
+++ b/ios/BT/ExposureManager.swift
@@ -14,6 +14,14 @@ final class ExposureManager: NSObject {
   let manager = ENManager()
   
   var detectingExposures = false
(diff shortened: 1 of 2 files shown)


octohub (approved):
Looks good to merge 💯

johnschoeman (approved):
LGTM, 💯

PR #1174, opened 2020-07-06
Introduce affected user flow for bt

johnschoeman:
Why:
Currently there is a handful of bugs in the Affected User flow (called
Export in the app currently) for the BT version. This is largely due to
the fact that the Export Flow was built with the assumption that the
user would be selecting the HealthAuthority as is the case with the GPS
version. We would like to update the Screens to be agnostic to this,
however for sake of time, we are introducing a similar flow for the BT
version and mounting the correct flow based off of the tracing Strategy.

This commit:
Introduces the AffectedUserFlow for the BT app.

Co-Authored-By: alejandrodustet <aledustet@gmail.com>

### GIF

affected-user-flow


app/bt/AffectedUserFlow/AffectedUserContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const AffectedUserProvider = ({
+  children,
+}: {
+  children: JSX.Element;
+}): JSX.Element => {
+  const [code, setCode] = useState('');
+
+  const healthAuthority = {
+    name: 'Test HA',
```
johnschoeman: We might want to clean up this?

jeffstolz (approved):
You probably want a pair of dev eyes on it, but I love the cleanup and the `AffectedUserFlow` naming convention 👍 . Lots of hard coded spacing values, which I'm sure are intentional and within a specific context, so I won't flag them.

mattThousand (approved):

PR #1179, opened 2020-07-06
Fix setup_env script.

mattThousand:


tparesi (approved):
LGTM - thanks for doing this

PR #1180, opened 2020-07-06
Update styles for affected user flow bt

johnschoeman:
Why:
Currently the affected user flow on the bt build has some stylistic
improvements to be made.

This commit:
Updates the style for the flow and uses react navigation to handle the
navigation header components.

### Before

affected-user-flow

### After

affected-user-flow-style

### DIFF
--- a/app/bt/AffectedUserFlow/CodeInput.tsx
+++ b/app/bt/AffectedUserFlow/CodeInput.tsx
@@ -12,17 +12,16 @@ import {
 import { SafeAreaView } from 'react-native-safe-area-context';
 
 import { Button } from '../../components/Button';
-import { IconButton } from '../../components/IconButton';
 import { Typography } from '../../components/Typography';
 import { useAffectedUserContext } from './AffectedUserContext';
 
 import { Screens } from '../../navigation';
-import { Icons } from '../../assets';
 import {
   Spacing,
   Buttons,
+  Layout,
+  Forms,
   Colors,
-  Outlines,
   Typography as TypographyStyles,
 } from '../../styles';
 
@@ -39,9 +38,7 @@ const CodeInputScreen = (): JSX.Element => {
   const length = CODE_LENGTH;
 
   const handleOnChangeText = (code: string) => {
-    if (code.length <= length) {
-      setCode(code);
-    }
+    setCode(code);
   };
 
   const handleOnPressNext = async () => {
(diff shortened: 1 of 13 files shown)


.eslintrc.js
```diff
@@ -82,6 +82,7 @@ module.exports = {
     'react/react-in-jsx-scope': 2, // Prevent missing React when using JSX
     'react/self-closing-comp': 2, // Prevent extra closing tags for components without children,
     'react/prefer-stateless-function': 2, // Use functional components vs classes
+    'react/display-name': 0,
```
aledustet: This rule is sort of helpful for viewing the dom tree while debugging something and for error outputs, to remove the errors, we would have to name the functions and not resort to anonymous declarations, I found that it is annoying sometimes but it encourages to properly name components. Definitely not important but to remove the error on the header declarations we could use a named function or pass the `CloseButton` or `BackButton` functions directly as the value(don't quote me on that). totally up to you but we could leave the rule if we want to.

app/bt/AffectedUserFlow/CodeInput.tsx
```diff
@@ -100,6 +84,7 @@ const CodeInputScreen = (): JSX.Element => {
                 <TextInput
                   testID={'code-input'}
                   value={code}
+                  maxLength={length}
```
aledustet, line 87: noice

app/bt/AffectedUserFlow/index.tsx
```diff
@@ -19,40 +23,94 @@ const fade: StackCardStyleInterpolator = ({ current }) => ({
...
+        <Stack.Screen
+          name={Screens.AffectedUserCodeInput}
+          component={CodeInput}
+          options={{
+            headerTransparent: true,
+            headerLeft: () => {
+              return <BackButton />;
+            },
+            headerTitle: '',
+            headerRight: () => {
```
aledustet: Could we pass `CloseButton` directly here? would this take care of the `display-name` rule error?
We could also name the function:

``` typescript
headerRight: function closeButton() { return <CloseButton />  }
```

app/views/onboarding/NotificationDetails.tsx
```diff
@@ -1,18 +1,15 @@
...
+import { Screens } from '../../navigation';
 
-interface NotificationDetailsProps {
-  navigation: any;
-}
-
-const NotificationDetails = ({
-  navigation,
-}: NotificationDetailsProps): JSX.Element => {
+const NotificationDetails = (): JSX.Element => {
```
aledustet, line 10: 👏  I think I saw a couple of warnings on `app/views/onboarding/PersonalPrivacy.tsx` and `app/views/onboarding/ShareDiagnosis.tsx` that are related to this. Is there a place where we could ask people to use the hook or the props from the library instead of typing it with any?

aledustet (approved):
Looking good, definitely non-blocking comments here that can be transformed into tickets later.

.eslintrc.js
johnschoeman: I definitely agree with your points here!  Ill make the updates to allow to add this back in.

app/views/onboarding/NotificationDetails.tsx
johnschoeman: Perhaps, I think a lot of this is coming from having ts being introduced gradually into the code base.  Imo, the best way to socialize not using any is via pairing and code review.  We do have a linting rule for no-explicit-any which we could up to error if we want?

PR #1182, opened 2020-07-07
Update styles for affected user flow

johnschoeman:
Why:
The current style implementation for the affected user flow in the BT
version of the app does not treat small format screens well.

This commit:
Updates the screens in the flow to be more responsive.

### Before

affected-user-flow-style-before
affected-user-flow-style-before-11

### After

affected-user-flow-style-after-11

affected-user-flow-style-after


app/bt/AffectedUserFlow/CodeInput.tsx
```diff
@@ -64,40 +66,36 @@ const CodeInputScreen = (): JSX.Element => {
...
+              <Typography style={styles.subheader}>
+                {t('export.code_input_body_bluetooth')}
+              </Typography>
+            </View>
+
+            <View>
+              <TextInput
+                testID={'code-input'}
+                value={code}
+                placeholder={'00000000'}
```
aledustet, line 87: Placeholders on the Android inputs sometimes look like the actual value there, I believe there is a `placeholderTextColor` or something similar that might be useful for the android version, not sure if this got fixed before 😬  maybe worth checking.

app/bt/AffectedUserFlow/CodeInput.tsx
```diff
@@ -115,34 +113,29 @@ const CodeInputScreen = (): JSX.Element => {
...
-    flex: 1,
   },
-  headerContainer: {
-    flex: 1,
-    marginBottom: Spacing.large,
+  container: {
+    height: '100%',
+    justifyContent: 'space-between',
+    padding: Spacing.medium,
+    paddingTop: Layout.oneTenthHeight,
```
aledustet, line 127: man, I love this variable name :chef-kissing-fingers:

app/bt/AffectedUserFlow/Complete.tsx
```diff
@@ -1,25 +1,82 @@
...
+
+  const handleOnPressDone = () => {
+    navigation.navigate(Screens.Settings);
+  };
+
+  const title = t('export.complete_title');
+  const body = StrategyCopy.exportCompleteBody;
 
   return (
-    <ExportTemplate
```
aledustet, line 16: 👏👏👏👏

app/bt/AffectedUserFlow/PublishConsent.tsx
```diff
@@ -1,38 +1,111 @@
...
+      setIsLoading(false);
     };
     BTNativeModule.submitDiagnosisKeys(cb);
   };
 
+  const title = StrategyCopy.exportPublishTitle;
+  const body = t('export.publish_consent_body_bluetooth');
+
   return (
-    <ExportTemplate
```
aledustet, line 27: 👏👏👏👏👏

app/bt/AffectedUserFlow/PublishConsent.tsx
```diff
@@ -1,38 +1,111 @@
...
   );
 };
 
+const styles = StyleSheet.create({
+  contentContainer: {
+    paddingTop: Layout.oneTenthHeight,
+    paddingHorizontal: Spacing.large,
+    paddingBottom: Spacing.huge,
+  },
+  backgroundImage: {
```
aledustet, line 87: Not a fan of extracting code for the sake of it but this style feels like one that will be copied over verbatim a few times, would it make sense to extract it like we did with the `secondaryContent` or `header2` ?

aledustet (approved):
🚢 :shipit:

app/bt/AffectedUserFlow/PublishConsent.tsx
johnschoeman: It might do, but we where chatting about this while pairing and we think the root issue is the way the navigation is set up for this stack requires the 100% cover background image. It would be preferable to solve the root issue and remove the need for `<ImageBackground/>` altogether.

PR #1183, opened 2020-07-07
[SAF-785] Exposure notification localisation

vlasac:
#### Description:
This PR adds missing localisation strings for the exposure history/notifications screen.

### DIFF
--- a/app/views/ExposureHistory/NextSteps.tsx
+++ b/app/views/ExposureHistory/NextSteps.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { TouchableOpacity, View, StyleSheet } from 'react-native';
 import { useNavigation } from '@react-navigation/native';
+import { useTranslation } from 'react-i18next';
 
 import { Typography } from '../../components/Typography';
 import { NavigationBarWrapper } from '../../components/NavigationBarWrapper';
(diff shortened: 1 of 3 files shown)


app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -12,21 +13,23 @@ import { AUTHORITY_NAME as healthAuthority } from '../../constants/authorities';
...
-
-  const contentTextOne =
-    'It is possible that you may have crossed paths with somebody who has been diagnosed with COVID-19.';
-
-  const contentTextTwo =
-    "This does not mean that you are infected, but you should take precautions anyway. People who don't exhibit symptoms can sometimes still be contagious.";
-
-  const buttonText = 'Take Self Assessment';
+  const headerText = t('exposure_history.next_steps.ha_self_assessment', {
+    healthAuthority,
```
johnschoeman: This was already here, but i might consider using `healthAuthorityName` here instead.

And as a side note worth mentioning the concept of a health authority is somewhat different between GPS and BT at this point, i see the team needing to work through this distinction coming up soon. This HA here seems to be more of the BT flavor of the concept, just so you know.

johnschoeman (approved):
LGTM! :ship:

app/views/ExposureHistory/NextSteps.tsx
vlasac: thanks, I've updated the naming, but I see that the name of the HA is static (`Boston Public Health Commission`).
Shouldn't we pass it as a prop or add it to the context?

JacobJaffe (approved):
LGTM! yay for full localization

app/views/ExposureHistory/NextSteps.tsx
JacobJaffe: One note on the HA part -- We'll be gating that off for GPS for now, I believe that's just for testing?

PR #1184, opened 2020-07-07
Update react native config to 1.2.1

aledustet:
Why:
----
The version for the library was fixed to commit 1eb6ac01991210ddad2989857359a0f6ee35d734
and an issue exists with our environment variables specifically for the
Bluetooth build, in an effort for troubleshooting it a first step is to
use the latest version of the library. The diff for the versions can be
examined [here]

[here]: https://github.com/luggit/react-native-config/commit/1eb6ac01991210ddad2989857359a0f6ee35d734?diff=split

This Commit:
----
- Upgrade `react-native-config` to latest version (1.2.1)

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -193,7 +193,7 @@ PODS:
   - React-jsinspector (0.61.5)
   - react-native-background-timer (2.2.0):
     - React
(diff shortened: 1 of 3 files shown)


johnschoeman (approved):
💯

PR #1185, opened 2020-07-07
Download only keys that haven't already been checked locally from the GAEN server

mattThousand:
### Why
We'd like the user to download only keys they haven't already checked from the GAEN server

### This Commit
This commit tracks the index of the last key checked from the GAEN server, and ensures that only keys that have not already been checked will be downloaded and submitted to the EN framework.

### DIFF
--- a/ios/BT/API/Model/ExposureConfiguration.swift
+++ b/ios/BT/API/Model/ExposureConfiguration.swift
@@ -15,7 +15,7 @@ struct ExposureConfiguration: Codable {
 extension ExposureConfiguration {
 
   static var placeholder: ExposureConfiguration = {
-    ExposureConfiguration(minimumRiskScore: 1,
+    ExposureConfiguration(minimumRiskScore: 0,
                           attenuationDurationThresholds: [50, 70],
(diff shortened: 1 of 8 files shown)


ios/BT/API/Model/ExposureConfiguration.swift
```diff
@@ -15,7 +15,7 @@ struct ExposureConfiguration: Codable {
 extension ExposureConfiguration {
 
   static var placeholder: ExposureConfiguration = {
-    ExposureConfiguration(minimumRiskScore: 1,
+    ExposureConfiguration(minimumRiskScore: 0,
```
johnschoeman, line 18: does this mean that we will only be receiving notifications on the js side that have at least a riskScore of 1?

johnschoeman (approved):
LGTM, one question.

ios/BT/API/Model/ExposureConfiguration.swift
mattThousand: New exposures will trigger a push notification and be stored in Realm on the native side; so when they are fetched from the JS side they will reflect this exposure configuration

PR #1186, opened 2020-07-07
Language Selection Screen

JacobJaffe:
Syncs Develop with the ahead change in production hotfix for Language Selection. 

### Motivation:
The native picker fails (causes flickering and rapidly selecting fields) on specific Android devices. 

### Changes:
- remove `NativePicker`
- adds `LanguageSelection` screen
- adds es-pr as a supported prod language 
- moves language _initialization_ logic 100% to the `app/locales/languages.js` file. This means that screens don't have to be aware of the underlying system language choice, which was causing problems only sometimes syncing, depending on component life cycles


app/views/LanguageSelection.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <FlatList
+        keyExtractor={(_, i) => `${i}`}
+        data={localeList}
+        renderItem={({ item: { value, label } }) => (
+          <TouchableHighlight
+            underlayColor={Colors.underlayPrimaryBackground}
+            style={{
+              paddingVertical: 20,
+              paddingHorizontal: 24,
+            }}
```
vlasac: Can we use `Spacing` here?
```suggestion
            style={{
              paddingVertical: Spacing.medium,
              paddingHorizontal: Spacing.large,
            }}
```

vlasac (approved):
Everything looks good to me, if using the constants from `Spacing` isn't ideal here, LGTM!

app/views/LanguageSelection.tsx
JacobJaffe: Good call!

PR #1187, opened 2020-07-07
Start the background service only on gps builds

aledustet:
Why:
----
The `IntersectionService` was returning null when the build is BT
because the location services are not available. This was resulting on
multiple un-handled rejections when building the app for BT.

This Commit:
----
- Add check for GPS strategy before starting the background service

[trello](https://trello.com/c/V5zjN6zm)


JacobJaffe (approved):
Great, thanks!

PR #1188, opened 2020-07-07
Dayjs localization updates

tnekret:
#### Description:

Matched dayjs locales to app locales.

This is work in progress, waiting on translations.


app/constants/locale.js
```diff
@@ -0,0 +1,5 @@
+export const CONVERT_ISO_TO_IETF = {
```
johnschoeman: if this is only being used in the languages.js file, what do you think about inlining it to that file?

johnschoeman (approved):

tparesi (approved):
LGTM

app/constants/locale.js
tnekret: fixed!

summetj:
Should this be stale?

JacobJaffe (approved):
🔥 this is elegant

PR #1190, opened 2020-07-07
allow export from manually entered HA

JacobJaffe:
Previously we only allowed selecting a manually entered HA on the `PartnersEdit` screen. 

Addresses QA need for dynamic testing


johnschoeman (approved):
Not in the loop on this functionality, but code looks good to me.

PR #1191, opened 2020-07-07
SAF - 135 download over wifi only - part 1

tindn:
#### Description:

Add the ability to only download HA data over wifi only if the user chooses to do so.

Changes include:
- Add the redux state to store app settings. Add the setting for downloading data over wifi only.
- In the Intersect code, before downloading the page data, check the added setting and connection type. If the user only wants to download over wifi and the current connection isn't wifi, skip the download of the page data.
- Add caching of page data so previously downloaded page can be used for intersection check.
- Add reactotron (only during development) for better network monitoring. Install the reactotron desktop app to use here https://github.com/infinitered/reactotron. Note, if you're running the app on an android device, besides reverse proxying port 8081, you'll also need to reverse proxy port 9090 by running the command ` adb -s android-device-id reverse tcp:9090 tcp:9090` where android device id is the id of the phone, you can find this by the command `adb devices` (while having the android device plugged in).

#### Linked issues:

https://pathcheck.atlassian.net/browse/SAF-135

#### Screenshots:

No screenshots yet as there isn't any UI available at this point. Currently waiting on design for this.

#### How to test:

1. By default, the app always allows downloading of page data. You'll see the page data request in reactotron as you change the selected HA. This should only happen once as the page data would be cached.
1. You can update the cache key to force cache miss, triggering the download again. 
1. Use redux dev tool to dispatch the action below to simulate wifi only mode.
```js
{
type: 'TOGGLE_DOWNLOAD_HA_DATA_OVER_WIFI_ONLY'
}
```
4. Turn the wifi off on the device. The download will no longer happen.


app/store/reducers/settingsReducer.ts
```diff
@@ -0,0 +1,13 @@
+import { createReducer } from '@reduxjs/toolkit';
+
+import toggleDownloadHaDataOverWifiOnlyAction from '../actions/settings/downloadHaDataOverWifiAction';
+
+const initialState = { downloadHaDataOverWifiOnly: false };
```
JacobJaffe: `downloadHaDataOverWifiOnly` -> Curious if this should be more general, e.g. `donloadLargeDataOverWifiOnly` ?

app/helpers/Intersect.js
```diff
@@ -571,3 +572,41 @@ export function disableDebugMode() {
...
 }
+
+async function getPageData(authority, page) {
+  // this function should always return an object.
+  // if there's no data, return an empty object.
+  const cacheKey = authority.internal_id + '|page:' + page.id;
+  let pageData = await GetStoreData(cacheKey);
+  if (!pageData) {
+    if (await shouldDownloadPageData()) {
+      try {
```
JacobJaffe: I think this function intendeds to always return `{}` on error?
Would it make sense for this try to be layer up then, and the `{}` _only_ gets returned as error handling?

I'm thinking for visibility into skipped pages we'd eventually want something like:

```js
async function getPageData(authority, page) {
  try {
     ...
     return (pageData);
   }  catch (e) {
    logSkippedAuthorityPage({timestamp, reason: e.message, authority, page});
    return ({});
  }
}
```

app/helpers/Intersect.js
```diff
@@ -571,3 +572,41 @@ export function disableDebugMode() {
...
+    } else {
+      // no cache and can't download, skip this page.
+      pageData = {};
+    }
+  }
+  return pageData;
+}
+
+async function shouldDownloadPageData() {
+  const reduxState = store.getState();
```
aledustet, line 565: I believe the store won't be available after we moved the import on this file to an accessor, updating it might require a null check but other than that not terribly different.

app/helpers/Intersect.js
```diff
@@ -571,3 +572,41 @@ export function disableDebugMode() {
   // Kick off intersection calculations to restore data
   checkIntersect();
 }
+
+async function getPageData(authority, page) {
+  // this function should always return an object.
+  // if there's no data, return an empty object.
+  const cacheKey = authority.internal_id + '|page:' + page.id;
```
aledustet, line 537: Should the key include some timestamp to ensure we have a recent version?

app/helpers/Intersect.js
```diff
@@ -571,3 +572,41 @@ export function disableDebugMode() {
...
+      try {
+        pageData = await retrieveUrlAsJson(page.filename);
+        SetStoreData(cacheKey, pageData);
+      } catch (e) {
+        // sometimes the url isn't right, the download will fail
+        pageData = {};
+      }
+    } else {
+      // no cache and can't download, skip this page.
+      pageData = {};
```
aledustet: Could we invert the nested logic to have a simpler tree here? Something like:

```typescript
const emptyPageData = {}
let pageData = await GetStoreData(cacheKey);

if (!pageData && await shouldDownloadPageData()) {
  try {
        pageData = await retrieveUrlAsJson(page.filename);
        SetStoreData(cacheKey, pageData);
    } catch (e) {
       // Maybe we need to log this or notify somehow of the broken URL?
    }
   return pageData || emptyPageData;
}
```

aledustet (approved):
Looks ready to me, probably the `Intersect.js` should be updated to use the new `StoreAccessor` but other than that looks ready.

JacobJaffe (approved):
LGTM! A few suggestions, possibly for the followups happening to this PR

app/helpers/Intersect.js
JacobJaffe, line 537: Can this also check that `page.id` is defined? I'm concerned about misconfigured cursor files with undefined ids mapping to the same cache key

app/helpers/Intersect.js
tindn: thank you. I wasn't aware of the change for store accessor. Will update.

app/helpers/Intersect.js
tindn: I'm not sure if I understood your comment.

If we append current timestamp to the key, it'll cause a cache miss every single time.

If we want to check on recent version, are you suggesting we store the saved timestamp in the payload and we check for expiration?

app/helpers/Intersect.js
tindn: to clarify, the suggested code looks the same but just combines `if (!pageData) {` and `if (await shouldDownloadPageData()) {` into one statement right?

Regarding the comment about error thrown, I wasn't aware of any logging, are you saying just console logging?

app/store/reducers/settingsReducer.ts
tindn: Sure.

app/helpers/Intersect.js
tindn: To clarify, are you saying for skipped pages, we manually throw an error to force the code path into the `catch` block?

I didn't intend for this function to always return `{}` on error, I meant for this to always return an object in all cases, so that the consumer of this function doesn't have to check for null.

app/helpers/Intersect.js
tindn: So if it's undefined, should we continue to download or skip the page? If so, it looks like the id forms part of the filename, so if the id is undefined, we might not be able to download either. But I can add a check here.

aledustet: Yes, I was thinking more of invalidation logic from the producer of the pages, a version number in the payload would make that simpler but I am not aware of how much control we have on what comes back. So maybe we can fabricate a timestamp with a `reasonable` window for invalidating our cache, I don't know how long that would be given the nature of the content in these pages.

app/helpers/Intersect.js
aledustet: You're right! Is just a combination to remove one branch of the decision flow, not at all an improvement but if we feel that one less branch could be beneficial maybe worth doing. About the logging, I'm not aware of ay aggregator service we are using but if not retrieving page data is something that needs correction on a third party, some visibility of these errors could help with this goal. Not sure if just logging it in the device logs would suffice or if not having a page available would grant any action, but if it does it might be worth storing some log info whether here or on a logging service that I am not sure we have at the moment.

app/helpers/Intersect.js
tindn: Yeah, I think cache invalidation has to come from the server, as any invalidation logic we put in would be quite arbitrary. We had a discussion with the backend team, and they're planning on adding a hash for invalidation. When that happens, we can add the hash to the cache key which will be better than any timestamp right now. So we can go with what I have currently and add the hash in whenever it's available, how does that sound?

vlasac (approved):
Everything looks good to me!

app/helpers/Intersect.js
JacobJaffe: >  it looks like the id forms part of the filename

I don't believe that'll be part of the contract for the API across all servers (current example server implementation is moving to a full file url, so servers might specify urls that are different than the id?)

I think treating it as an error and skipping sounds good here

app/helpers/Intersect.js
JacobJaffe: Yeah, similarly to what @aledustet is recommending, it reads to me like all of the `skipped` pages occur from errors, so it would be nice for logging. That said, logging is hypothetical around this right now, so maybe just worth revisiting when introduced.

PR #1192, opened 2020-07-07
Fix cyclic dependency warning

aledustet:
Why:
----
Since the store needs to be made available when checking for intersects for the health authority access, there is a cyclic dependency warning that shows up every time the application builds. We need a way to avoid this cyclic dependency from happening.

How:
----
The approach was to add a getter and a setter singleton to access the store via a third file from `Intersect.js`. A possible future approach could be to move the intersect logic onto the redux store but it is future work to be done since it will involve a more in-depth rework of the logic.

This Commit:
----
- Add a `StoreAccessor` to make the store available without importing the actual store.
- Use the StoreAccessor from the `Intersect` helper

How to test:
----
- After building the GPS flavor, go to the partners screen and after selecting  one of the partners we should not see output in the console with the error:
```
Attempting to access a not set store checking for intersect
```

[trello](https://trello.com/c/Cb9wAu2f)

Co-Authored-By: Jacob Jaffe <jacobmaxjaffe@gmail.com>

### DIFF
--- a/app/helpers/Intersect.js
+++ b/app/helpers/Intersect.js
@@ -32,7 +32,7 @@ import {
 import { MIN_LOCATION_UPDATE_MS } from '../services/LocationService';
 import languages from '../locales/languages';
 
-import { store } from '../store';
+import StoreAccessor from '../store/StoreAccessor';
 
 import getCursor from '../api/healthcareAuthorities/getCursorApi';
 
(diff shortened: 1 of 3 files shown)


JacobJaffe (approved):
Hoary for generic redux store types !

johnschoeman (approved):

PR #1194, opened 2020-07-07
Remove use of theme from App.

tparesi:
#### Description:

Remove the usage of Theme from the app, since it was not being used inside the app (only in tests).

#### Linked issues:

[Trello Card](https://trello.com/c/TFCTdpqB/24-as-a-dev-team-we-do-not-have-an-unused-theme-component-in-the-code-base)

#### How to test:

Please click around to make sure there were no visual regressions.


tparesi:
As we were reviewing PR's, saw that Jacob already did this: https://github.com/Path-Check/covid-safe-paths/pull/1189

Closing

PR #1196, opened 2020-07-07
Fix typo on NotificationDetails and type warning

aledustet:
Why:
----
We have a type on the `NotificationDetails` component being exported as
'NotificatioNDetails`.
There is an explicit any on a couple of files that were using the
navigation.

This Commit:
----
- Use the `useNavigation` hook in favor of untyped props interface
- Replace uses of `NotificatioNDetails` for `NotificationDetails`


tparesi (approved):
LGTM

johnschoeman (approved):
Love it

PR #1197, opened 2020-07-07
Update affected user flow navigation

devinjameson:
Why: prior to this commit, the navigation for the affected user flow was
handled through the top nav bar. Given the user journey for this flow,
it makes more sense to have primary and secondary action buttons at the
bottom of the screen.

This commit:
- Removes the top bar navigation from the affected user flow
- Adds a secondary button to some screens on the affected user flow

## Before
demo

## After
demo


tparesi (approved):
That looks much better!

johnschoeman (approved):
LGTM!

PR #1198, opened 2020-07-07
Run Key Detection Job at Most Once per 24 hr Period

mattThousand:
### Why
We'd like to run the key detection task at most every 24 hours to accommodate potentially large numbers of keys and the necessity of a very long-running job

### This Commit
This commit increases the duration between runs of the exposure detection task to 24 hours updates the download path to point to the correct endpoint. It also lowers the exposure risk threshold for testing pruposes.

### DIFF
--- a/ios/BT/API/Model/ExposureConfiguration.swift
+++ b/ios/BT/API/Model/ExposureConfiguration.swift
@@ -15,7 +15,7 @@ struct ExposureConfiguration: Codable {
 extension ExposureConfiguration {
 
   static var placeholder: ExposureConfiguration = {
-    ExposureConfiguration(minimumRiskScore: 1,
+    ExposureConfiguration(minimumRiskScore: 0,
                           attenuationDurationThresholds: [50, 70],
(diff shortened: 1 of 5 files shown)


johnschoeman (approved):

tparesi:
Is this a temporary thing? I am asking, bc Amanda and I are working on adding when the Exposure History was last "detected", but if it is now a daily job, maybe this timestamp is not necessary on the Exposure History page:

https://trello.com/c/mF6fMBd5/42-as-a-user-when-i-am-on-the-exposure-history-screen-i-see-a-not-of-the-last-time-that-the-history-was-updated-which-is-correct

mattThousand:
> Is this a temporary thing? I am asking, bc Amanda and I are working on adding when the Exposure History was last "detected", but if it is now a daily job, maybe this timestamp is not necessary on the Exposure History page:
> 
> https://trello.com/c/mF6fMBd5/42-as-a-user-when-i-am-on-the-exposure-history-screen-i-see-a-not-of-the-last-time-that-the-history-was-updated-which-is-correct

It's a daily job but it doesn't run at a specified time, so the "last detection" UI would still be great

PR #1199, opened 2020-07-08
fix feature flag toggle

tnekret:
#### Description:
Fixed feature flag toggle not triggering

#### How to test:

- Go to About page

- Click 10 times on PathCheck header

- Feature flags should be enabled


johnschoeman (approved):
💯

PR #1200, opened 2020-07-08
Added staging build icon

jvisenti:


ios/Podfile.lock
```diff
@@ -193,7 +193,7 @@ PODS:
   - React-jsinspector (0.61.5)
   - react-native-background-timer (2.2.0):
     - React
-  - react-native-config (1.2.1):
+  - react-native-config (0.11.7):
```
jvisenti: I don't know why this is changing, but running `yarn install:pod` locally updates the lock file.

johnschoeman: probably need to run `yarn install` first. A previous commit updated `react-native-config` to 1.2.1 and that probably what we want here.

johnschoeman (approved):
LGTM

JacobJaffe (approved):
🎉 terrific, thanks

ios/Podfile.lock
jvisenti: Ah that makes sense, all set now

PR #1201, opened 2020-07-08
Submit code to verification server

johnschoeman:
Why:
When a user has received a verification code from their health provider
they need to be able to submit the code to the verification server in
order to receive a certificate to be able to then submit their exposure
tokens.

This commit:
Wires up the network logic for submitting the code to a verification
server and provided some error handling for rejected requests. Note that
we have not finished the work for the next step of submitting the
received token to generated a signed certificate, but we have that api
function started. A future commit will finish this interaction.

### GIF

verification-code

### DIFF
--- a/app/bt/AffectedUserFlow/CodeInput.tsx
+++ b/app/bt/AffectedUserFlow/CodeInput.tsx
@@ -2,6 +2,7 @@ import React, { useState } from 'react';
 import { useNavigation } from '@react-navigation/native';
 import { useTranslation } from 'react-i18next';
 import {
+  ActivityIndicator,
   Alert,
   KeyboardAvoidingView,
   StyleSheet,
@@ -13,9 +14,9 @@ import {
 } from 'react-native';
 import { SafeAreaView } from 'react-native-safe-area-context';
 
-import { Button } from '../../components/Button';
 import { Typography } from '../../components/Typography';
 import { useAffectedUserContext } from './AffectedUserContext';
+import * as API from './verificationAPI';
 
 import { Screens } from '../../navigation';
 import {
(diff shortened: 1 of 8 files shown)


devinjameson (approved):
👍

app/bt/AffectedUserFlow/CodeInput.tsx
```diff
@@ -104,18 +131,24 @@ const CodeInputScreen = (): JSX.Element => {
               </View>
 
               <Typography style={styles.errorSubtitle} use='body2'>
-                {codeInvalid ? t('export.code_input_error') : ' '}
+                {errorMessage}
               </Typography>
             </View>
+            {isLoading ? <LoadingIndicator /> : null}
```
aledustet, line 137: Curious if there's a difference or preference on readability for using `isLoading && <LoadingIndicator />` I think this is great just curious 😄

app/bt/AffectedUserFlow/verificationAPI.ts
```diff
@@ -0,0 +1,122 @@
+import Config from 'react-native-config';
+
+const baseUrl = Config.GAEN_VERIFY_URL;
```
aledustet, line 3: Do we have a convention on naming constants? I see the config ones using all caps snake case, should this constants be named as such?

app/bt/AffectedUserFlow/verificationAPI.ts
```diff
@@ -0,0 +1,122 @@
...
+  error: string;
+  testDate: string;
+  testType: string;
+  token: string;
+}
+
+export const postVerificationCode = async (
+  code: string,
+): Promise<NetworkResponse<CodeVerificationSuccess, CodeVerificationError>> => {
+  const url = verifyUrl;
```
aledustet: How do we feel about inlining the url with the fetch call? Or maybe move the assignment closer?

app/bt/AffectedUserFlow/verificationAPI.ts
```diff
@@ -0,0 +1,122 @@
...
+export type CodeVerificationError =
+  | 'InvalidCode'
+  | 'VerificationCodeUsed'
+  | 'Unknown';
+
+interface VerifiedCodeResponse {
+  error: string;
+  testDate: string;
+  testType: string;
+  token: string;
```
aledustet: How do we feel about using the `Token` alias here for extra swag factor?

aledustet (approved):
This LGTM maybe we could add a test in the future but looks solid.

app/bt/AffectedUserFlow/verificationAPI.ts
devinjameson: SWAG 🦩

app/bt/AffectedUserFlow/CodeInput.tsx
johnschoeman: Certainly a style preference.  I like using the ternary over the and operator, as I always know i can use it for conditional rendering, so i like to just use the same thing everywhere. And while it's less pretty, having the `null` explicit feels a bit more honest and clear, as where the `isLoading && <LoadingIndicator />` expression still resolves to `null`, it's not obvious at a glance that this is the case.  So definitely a single vs. double quote level of importance, but the ternary is my preference.  Would be happy to adopt a project wide style either way.

app/bt/AffectedUserFlow/verificationAPI.ts
johnschoeman: hmm, no, but I'd prefer defaulting to snakeCase over SCREAMING_SNAKE_CASE for consts that arent global-ish.

app/bt/AffectedUserFlow/verificationAPI.ts
johnschoeman: works for me! i do like keeping the urls at the top for documentation purposes, but we can inline em.

app/bt/AffectedUserFlow/CodeInput.tsx
```diff
@@ -131,6 +164,21 @@ const CodeInputScreen = (): JSX.Element => {
   );
 };
 
+const LoadingIndicator = () => {
```
mattThousand, line 167: just a question, not sure what the answer is, but should we have 1 or 2 loading indicators that we store in a single place and use throughout the app for stylistic consistency? 🤔

johnschoeman: @devinjameson and I chatted about this. We arent 100% sure at this point, as this is the only place we want this right now, but if we do have a few it makes sense to abstract at least the styles out to be reused. Either way it would makes sense in a follow up PR and we cut a card for this work here: https://trello.com/c/bBehk7Pr/168-as-a-user-i-see-a-consistent-and-well-styled-loading-indicator-throughout-app-when-data-is-being-loading

app/bt/AffectedUserFlow/verificationAPI.ts
```diff
@@ -0,0 +1,118 @@
+import Config from 'react-native-config';
+
+const baseUrl = Config.GAEN_VERIFY_URL;
+const verifyUrl = `${baseUrl}/api/verify`;
+const certificateUrl = `${baseUrl}/api/certificate`;
+
+const defaultHeaders = {
+  'content-type': 'application/json',
+  accept: 'application/json',
+  'X-API-Key': Config.GAEN_VERIFY_API_TOKEN,
```
mattThousand, line 10: these header keys _could_ be constants, might be handy if we're constructing headers elsewhere

app/bt/AffectedUserFlow/verificationAPI.ts
johnschoeman: or at least intended to be global

example.env.bt
```diff
@@ -2,4 +2,8 @@ TRACING_STRATEGY=bt
 POST_DIAGNOSIS_KEYS_URL=gaen_server_url_here
 DOWNLOAD_KEY_FILE_URL=gaen_server_url_here
 INDEX_FILE_URL=gaen_server_url_here
-AUTHORITIES_YAML_ROUTE=yaml_route_here
\ No newline at end of file
+HMAC_KEY=abcd1234
+GAEN_VERIFY_URL=https://example.com
+GAEN_VERIFY_API_TOKEN=abcd1234
```
mattThousand, line 7: mind making a PR to `pathcheck-mobile-resources` to update the `.env`s with the production values? 😁

mattThousand (approved):
🎉

app/bt/AffectedUserFlow/verificationAPI.ts
johnschoeman: totally, I would expect that this would be the only bit that needs it though as we will only have the two endpoints to hit for the verification flow.

app/bt/AffectedUserFlow/CodeInput.tsx
aledustet: Thanks for offering your point of view, I see why consistency across the board is better, letting go of loosely held opinion of my own. I like this idea a lot.

PR #1202, opened 2020-07-08
Show Alert to User When Enabling EN for App if EN is not Authorized by the OS

mattThousand:
### Why
Currently the UI does not update in the scenario where Exposure Notifications have been deauthorized by the operating system, and the user presses the `Allow Exposure Notifications` button on the home screen. We'd like to present instructions to the user such that they know how to proceed.

### This Commit
This commit presents an alert dialog to the user instructing them to authorize Exposure Notifications from the Settings app. 
Note:
* Waiting on final copy to merge
* Android does not have a notion of enabled/disabled, only authorized/unauthorized; @octohub just putting this on your radar in case it is a scenario that you'll need to handle 

IMG_0C6B0400D350-1


app/bt/Home/Home.tsx
```diff
@@ -37,8 +44,25 @@ const Home = ({
     : t('home.bluetooth.tracing_off_subheader');
   const buttonText = t('home.bluetooth.tracing_off_button');
 
+  const showUnAuthorizedAlert = () => {
```
devinjameson: Thoughts on `showUnauthorizedAlert` here? I find it more readable.

devinjameson (approved):
Nice!

johnschoeman (approved):
💯 , ship it for sure, but I also find that this would really benefit from some spec coverage for documentation and test coverage purposes, as its a bit of a confusing part of the app.  What do you think about cutting a card for adding some GUTs for this functionality?

Added a spec, let me know what you think 👨‍🎨

PR #1204, opened 2020-07-09
Implement next steps screen redesign

zhaccc:
#### Description:

This PR implement redesign on the `Next Steps` screen.
Also, by pressing a button on-screen user is now redirected to his preferred HA website that describes what individuals should do if they have been exposed to Covid-19 (this is currently hardcoded as we still don't store user-preferred HA).


app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -63,8 +70,9 @@ const styles = StyleSheet.create({
   headerContainer: {
     flex: 1,
   },
-  headerText: {
-    ...TypographyStyles.header3,
+  footerText: {
+    ...TypographyStyles.footer,
+    marginBottom: 20,
```
devinjameson: Thoughts on using a `Spacing` variable here?

app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -76,6 +84,8 @@ const styles = StyleSheet.create({
   },
   button: {
     ...Buttons.largeBlue,
+    padding: 30,
```
devinjameson: Thought on a `Spacing` variable here, too?

app/locales/en.json
```diff
@@ -146,10 +146,10 @@
...
     },
     "next_steps": {
-      "ha_self_assessment": "{{healthAuthority}} recommends you take a self-assessment",
-      "possible_crossed_paths": "It is possible that you may have crossed paths with somebody who has been diagnosed with COVID-19.",
+      "ha_self_assessment": "Follow the advice from {{healthAuthorityName}} about what to do next",
```
devinjameson: Thoughts on ending this sentence with a period?

devinjameson (commented):
Looks good! Just a few comments.

app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -32,22 +38,23 @@ const NextSteps = (): JSX.Element => {
   const buttonText = t('exposure_history.next_steps.button_text');
 
   const handleOnPressTakeAssessment = () => {
-    navigation.navigate(Screens.SelfAssessment);
+    Linking.openURL(AUTHORITY_ADVICE_URL);
```
johnschoeman: what do you think bout something like: 

```suggestion
    AUTHORITY_ADVICE_URL ? Linking.openURL(AUTHORITY_ADVICE_URL) : navigation.navigate(Screens.SelfAssessment)
```

its not clear to me that every HA will have an adviceUrl and it might be nice to leave ourselves open to this case

app/assets/svgs/export.ts
```diff
@@ -1,4 +1,4 @@
 export default `<?xml version="1.0" encoding="UTF-8"?>
-<svg width="16" height="21" viewBox="0 0 16 21" fill="none" xmlns="http://www.w3.org/2000/svg">
```
johnschoeman, line 2: what lead to this change?

johnschoeman (approved):
Looks good! a few minor thoughts

app/assets/svgs/export.ts
zhaccc: Color fill property was hardcoded with a specific color so we could not programmatically change it within the component. Also since this icon is not used anywhere except in the next steps screen so it seemed right to do this change now.

app/views/ExposureHistory/NextSteps.tsx
zhaccc: Yep, that seems right.

app/views/ExposureHistory/NextSteps.tsx
```diff
@@ -1,15 +1,21 @@
...
 
 import { Buttons, Spacing, Typography as TypographyStyles } from '../../styles';
+import { Icons } from '../../assets';
 
-import { AUTHORITY_NAME as healthAuthorityName } from '../../constants/authorities';
+import {
+  AUTHORITY_NAME as healthAuthorityName,
+  AUTHORITY_ADVICE_URL,
+} from '../../constants/authorities';
+import { Colors } from 'react-native/Libraries/NewAppScreen';
```
johnschoeman: Why are we using these colors? should we be using the common styles.

zhaccc: Whoops, VS code auto-import... 🤦

PR #1206, opened 2020-07-09
Update affected user flow to not use modal mode

devinjameson:
Why: prior to this commit, the affected user flow stack was being
displayed within a modal using React Navigation. However, it is not
possible (as of now) to disable the "swipe down to close" gesture for
modals presented via React Navigation. Given that we want to disable
this gesture, we had to take a different approach.

This commit:
- Removes the modal mode from the options for the affected user flow
  stack
- Instead uses a custom cardStyleInterpolator to achieve the
  animateFromBottom transition when opening this stack

## Before
demob4

## After
demo

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -3,6 +3,7 @@ import React, {
   useState,
   useEffect,
   FunctionComponent,
+  useCallback,
 } from 'react';
 
 import {
@@ -87,10 +88,6 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
     return subscription.remove;
   }, [exposureInfoSubscription, toExposureHistory]);
 
-  useEffect(() => {
-    getCurrentExposures();
-  }, [toExposureHistory]);
-
   const observeExposures = () => {
     setUserHasNewExposure(false);
   };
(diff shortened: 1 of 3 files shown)


johnschoeman (approved):
LGTM, i would just double check that this doesnt negatively affect the export flow on the gps build

devinjameson:
> LGTM, i would just double check that this doesnt negatively affect the export flow on the gps build

@johnschoeman How might I do this? Is it something I can test on my own or do I need to ask someone? I haven't touched the GPS side.

johnschoeman:
> > LGTM, i would just double check that this doesnt negatively affect the export flow on the gps build
> 
> @johnschoeman How might I do this? Is it something I can test on my own or do I need to ask someone? I haven't touched the GPS side.

we can pair on it. but in short youll need to run ```yarn run-ios-gps``` for the gps scheme.  I can also show you how to do it in xcode sometime.

johnschoeman:
further activity

app/ExposureHistoryContext.tsx
```diff
@@ -87,10 +88,6 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
     return subscription.remove;
   }, [exposureInfoSubscription, toExposureHistory]);
 
-  useEffect(() => {
-    getCurrentExposures();
-  }, [toExposureHistory]);
-
```
mattThousand, line 93: @devinjameson can you put this guy back? The calendar doesn't populate with exposures on app launch if this is removed 🙇

mattThousand: cc @aledustet

PR #1208, opened 2020-07-09
Refactor method of displaying screens over tab bar

devinjameson:
Why: prior to this commit, we were displaying modals that show over the
tab bar by including them in the top-level stack navigator, even though
they are structurally within individual tabs. This pattern is
non-optimal.

This commit:
- Moves the affected user flow and language selection to within the
settings tab
- Hides the tab bar when the affected user flow is mounted
- Moves the language selection out of a modal view for consistency with
  the other settings screens

## Before
before

## After
after


aledustet (approved):
:shipit: 🚢

johnschoeman (approved):
LGTM, same comment as the other prs. lets just make sure this still works as expected with `yarn run-ios-gps`

PR #1209, opened 2020-07-09
Refactor nav bar on settings screens

devinjameson:
Why: prior to this commit, we were using a custom nav bar component for
navigation on the settings screens.

This commit: updates the nav bar to use React Navigation instead.

## Before
before

## After
after


ios/Podfile.lock
```diff
@@ -530,4 +530,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: d641f53b267b5fdc2398388c8211cf0739401d29
 
-COCOAPODS: 1.9.2
```
aledustet: Should we document the why of this downgrade?

app/Entry.js
```diff
@@ -101,15 +101,55 @@ const SelfAssessmentStack = () => (
...
 const MoreTabStack = () => (
-  <Stack.Navigator screenOptions={SCREEN_OPTIONS}>
-    <Stack.Screen name={Screens.Settings} component={SettingsScreen} />
-    <Stack.Screen name={Screens.About} component={AboutScreen} />
-    <Stack.Screen name={Screens.Licenses} component={LicensesScreen} />
-    <Stack.Screen name={Screens.FeatureFlags} component={FeatureFlagsScreen} />
-    <Stack.Screen name={Screens.Import} component={ImportScreen} />
-    <Stack.Screen name={Screens.ImportFromUrl} component={ImportFromUrl} />
-    <Stack.Screen name={Screens.ENDebugMenu} component={ENDebugMenu} />
+  <Stack.Navigator>
```
aledustet: Solid! Is it possible to add as the default options the `moreTabScreenOptions` and override the two that don't have the same settings with an empty set? Is probably not important but the diff might get smaller. 😁

aledustet (approved):
👏 💪 
We might want to check the cocoa pods version but other than that is ready to 🚢

ios/Podfile.lock
devinjameson: Can you actually help me understand why this happened?

app/Entry.js
devinjameson: Totally! Good idea.

ios/Podfile.lock
aledustet: Oh it's probably an outdated version cocoapods 🤔 maybe running `gem install cocoapods` will take care of it?

app/Entry.js
```diff
@@ -12,7 +12,7 @@ import {
 } from '@react-navigation/stack';
 import { useSelector } from 'react-redux';
 
-import SettingsScreen from './views/Settings';
```
johnschoeman, line 14: 🥇

app/views/More/ENLocalDiagnosisKeyScreen.tsx
```diff
@@ -69,14 +68,8 @@ export const ENLocalDiagnosisKeyScreen = ({
...
 
-  const backToDebugMenu = () => {
-    navigation.goBack();
-  };
-
   return (
-    <NavigationBarWrapper
-      title={'Local Diagnosis Keys'}
-      onBackPress={backToDebugMenu}>
+    <>
```
johnschoeman: can this fragment be removed?

johnschoeman (approved):
💯 this is a huge improvement.  I would just double check that everything works as intended on the gps build as well. for context we are planning a gps release next week and we need to be diligent about not introducing regressions.

app/views/More/ENLocalDiagnosisKeyScreen.tsx
devinjameson: Good catch!

johnschoeman:
more activity

aledustet:
@johnschoeman the rebase was not as smooth as I expected, would you mind having a look?

PR #1211, opened 2020-07-10
Update test preset to allow for text matchers and add test to affected user flow

aledustet:
Why:
----
When we are running our tests, most of the components that get rendered into a screen are `native` powered, making it difficult for test renderers to display them in a proper way. There is a set of presets that makes this possible. In an effort to make our integration and unit tests more similar to what the user is seeing and giving us more confidence in what we can test, we need to add the react-native testing library preset that will allow us to run matchers on what is being displayed to the user on the screen. This way our tests can be more similar to the actual behavior of the application. This aligns with the philosophy of the [react-native-testing-library](https://www.native-testing-library.com/)
This change was inspired because we started adding tests to the affected user flow for the BT flavor of the application. The logic that handled the validation is outside a test harness, and we need to get it to a comfortable spot while we keep making iterations to it.

**CAVEAT**
The `happy` path and the API test are left out since those are actively being changed, but this PR puts us in a position to add the tests as we develop those features ... more tests are on the way.

For Reviewers:
----
The diff is HUGE but mostly because of snapshots upgrades. Things worth noticing are an addition to the jest config to ignore some of the packages we use when transforming them before running the test suite and one more mock inclussion.

This Commit:
----
- Add preset `@testing-library/react-native` to jest config
- Update snapshots for the new renderer result
- Stop babel from transforming modules when the test runner is working
- Test behavior of the verification code validation logic against the verification server

### DIFF
--- a/app/bt/AffectedUserFlow/CodeInput.spec.tsx
+++ b/app/bt/AffectedUserFlow/CodeInput.spec.tsx
@@ -1,9 +1,15 @@
 import React from 'react';
-import { cleanup, render } from '@testing-library/react-native';
+import {
+  cleanup,
+  render,
+  fireEvent,
+  wait,
+} from '@testing-library/react-native';
 import '@testing-library/jest-native/extend-expect';
 
 import CodeInputScreen from './CodeInput';
 import { AffectedUserProvider } from './AffectedUserContext';
+import * as API from './verificationAPI';
 
 afterEach(cleanup);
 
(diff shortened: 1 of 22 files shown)


app/bt/AffectedUserFlow/CodeInput.spec.tsx
```diff
@@ -18,4 +24,77 @@ describe('CodeInputScreen', () => {
     expect(getByTestId('affected-user-code-input-screen')).not.toBeNull();
     expect(getByTestId('code-input')).toHaveTextContent('');
   });
+
+  describe('validates the verification code', () => {
+    it('informs of an invalid code error', async () => {
```
johnschoeman, line 29: Do we also want to have a spec for the a valid response being received?

app/bt/Home/__snapshots__/index.spec.tsx.snap
```diff
@@ -156,10 +156,8 @@ exports[`HomeScreen renders 1`] = `
             Enable Exposure Notifications to receive information about possible exposures
           </Text>
         </View>
-        <View
-          accessible={true}
-          focusable={true}
-          isTVSelectable={true}
+        <TouchableOpacity
```
johnschoeman, line 159: Why does this pr update these snapshots?

johnschoeman (approved):
This looks so good! thanks for adding this.

app/bt/AffectedUserFlow/CodeInput.spec.tsx
```diff
@@ -18,4 +24,77 @@ describe('CodeInputScreen', () => {
...
+        .spyOn(API, 'postVerificationCode')
+        .mockResolvedValueOnce(wrongTokenResponse);
+
+      const { getByTestId, getByLabelText, getByText } = render(
+        <AffectedUserProvider>
+          <CodeInputScreen />
+        </AffectedUserProvider>,
+      );
+      fireEvent.changeText(getByTestId('code-input'), '12345678');
+      fireEvent.press(getByLabelText('Submit'));
```
tparesi, line 45: Amanda and I are writing tests for the onboarding flow. Since most of the text is from translations, is there a good way to use the translations, or is hardcoding a fine approach?

tparesi (approved):
This looks great! I have one question - more for me than the PR

app/bt/AffectedUserFlow/CodeInput.spec.tsx
johnschoeman: Not having translations in specs makes them more readable and the intent of the specs isnt to test translations per say, so I prefer having the actual text.  And simply updating the specs if the copy updates.

aledustet: Good question, this crossed my mind but transforming things and exploding builds side-tracked me 😂 In the past I usually mock the `t` function and search for the keys, but that puts us in a pickle where we might not have the key added in at least the default translation so I would vote for either having it this way or for mocking the function and validating that translations keys are present but I don't have an idea on how to do that. Happy to brainstorm some though 😄

tparesi: I am down with this way - wanted some validation. This is great!

PR #1213, opened 2020-07-10
Adjust key submission logic for iOS

mattThousand:
### Why
We'd like to submit only keys from within the last 336 (14 days) hours to the server

### This Commit
This commit filters the keys according to the specified lifetime.


johnschoeman (approved):
LGTM

PR #1215, opened 2020-07-10
update version in build.gradle, change version in About screen

vlasac:
#### Description:
This PR updates the `About` screen to show correct build numbers for android and iOS

### DIFF
--- a/app/views/About.js
+++ b/app/views/About.js
@@ -10,7 +10,7 @@ import {
   Alert,
   TouchableWithoutFeedback,
 } from 'react-native';
-import { getVersion } from 'react-native-device-info';
+import { getVersion, getBuildNumber } from 'react-native-device-info';
 
 import { useStrategyContent } from '../TracingStrategyContext';
 import { NavigationBarWrapper, Typography } from '../components';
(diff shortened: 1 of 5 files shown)


tnekret (approved):

app/views/About.js
```diff
@@ -12,14 +12,14 @@ import {
...
 import { useStrategyContent } from '../TracingStrategyContext';
-import packageJson from '../../package.json';
 import { NavigationBarWrapper, Typography } from '../components';
 
 import { useDispatch } from 'react-redux';
 import toggleAllowFeatureFlagsAction from '../store/actions/featureFlags/toggleAllowFeatureFlagsEnabledAction';
 import { Colors, Spacing, Typography as TypographyStyles } from '../styles';
 
 const CLICKS_TO_ENABLE_FEATURE_FLAGS = 10;
+const APP_VERSION = 'ALPHA.1.0.0.b7';
```
tindn: you can use `react-native-device-info` package to get the app version, saving you the trouble of having to manually update this every time.

jvisenti (approved):
LGTM, although if @tindn's comment about automatically pulling the app version is straightforward we should do that.

johnschoeman (approved):
Took a look, I agree with @tindn , we should avoid hard coding this info here.

app/views/About.js
vlasac: Thanks, the package is not yet added to the codebase, I'll add it now.

vlasac: I also added the build number to the About screen because on iOS the version would always be `1.0.0` until we decide to bump it, making it harder for testers to know which exact version they are using.

PR #1216, opened 2020-07-10
Transform languages helper to typescript

aledustet:
Why:
----
Started adding some functionality that turned to be not required and in
the process started moving this file from js to ts.

This PR:
----
- Transform languages helper to Typescript


app/locales/languages.ts
```diff
@@ -130,26 +136,39 @@ export const initDevLanguages = () => {
...
+}> => {
+  const resources = i18next.options.resources;
+  if (!resources) {
+    return [];
+  }
+  return Object.entries(resources)
     .map(([langCode, lang]) => ({
       value: langCode,
-      label: lang.label,
+      label: lang.label as string,
```
johnschoeman, line 150: any way we can avoid this type casting here?  I would tackle this in a separate pr maybe, but seems like `Object.entries` like a code smell?  Why do we even need to do this conversion? What's the difference between `value` and `label` here?

johnschoeman (approved):
Looks great! thanks for tackling the ts conversion here.

app/locales/languages.ts
aledustet: The return value of the resources method returns a union type called `ResourceKey` that looks like this:

``` typescript

type ResourceKey =
  | string
  | {
      [key: string]: any;
    };
```

We need to decide which type of the union to use and the API exposed here is consumed with the assumption all of those values are strings, so someone needs to do the casting whether here or on the consumers. Happy to shift it but someone will cast it 😓

PR #1217, opened 2020-07-10
Non-empty App Icon for BT Dev

mattThousand:
### Why
There is an empty app icon for non-release builds on the BT app

### This Commit
This commit duplicates the release app icon asset for BT Dev


aledustet (approved):
:shipit:

PR #1218, opened 2020-07-10
Download Exposure Keys and Provide them to GAEN

octohub:
#### Description:

As soon as EN are enabled, this will schedule a job to run once per day to download keys and provide them to EN. This is still a work in progress as the app is hitting a dummy text file to determine which zip files to download. This allows for the job to complete much more quickly to help more rapid testing. Also hooked up "Detect Exposures Now" functionality in the debug menu.

There are no visible user changes. That will be in a follow up PR. But I wanted to at least get the logic in first.

#### How to test:
1. Fresh install of app.
2. Enable EN in onboarding.
3. Observe in Android Logcat that the keys are checked against EN and the exposure status is reported.


mattThousand (approved):
LGTM!

PR #1219, opened 2020-07-10
Add tests for the Share Diagnosis screen

aledustet:
Why:
----
We have some regression on the onboarding flow that needs to be corrected. Some patterns to test the different flavors
of the application(bt or gps) are not present in the test suite.

This Commit:
----
- Add tests for the ShareDiagnosis screen
- Separate test files dependant of the flavor config(gps or bt) into files with bt or gps extension
- Don't show notification permission for android bluetooth
- Delete global navigation mock

Co-Authored-By: Thomas G Paresi <tparesi@gmail.com>
Co-Authored-By: amandabeiner <aebeiner@gmail.com>

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -1,4 +1,9 @@
-import React, { createContext, useState, useEffect } from 'react';
+import React, {
+  createContext,
+  useState,
+  useEffect,
+  FunctionComponent,
+} from 'react';
 
 import {
   blankExposureHistory,
@@ -40,7 +45,6 @@ export interface ExposureEventsStrategy {
 }
 
 interface ExposureHistoryProps {
-  children: JSX.Element;
   exposureEventsStrategy: ExposureEventsStrategy;
 }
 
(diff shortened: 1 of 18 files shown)


app/ExposureHistoryContext.tsx
```diff
@@ -53,10 +57,10 @@ const blankHistoryConfig: ExposureCalendarOptions = {
 
 const blankHistory = blankExposureHistory(blankHistoryConfig);
 
-const ExposureHistoryProvider = ({
+const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
```
johnschoeman, line 60: 💯

jest/setupFile.js
```diff
@@ -75,33 +75,3 @@ jest.mock('@react-navigation/stack', () => {
     },
   };
 });
-
```
johnschoeman, line 78: 💯 , this was bothering me so much, thanks for pulling it out of the global context.

johnschoeman (approved):
Overall looks great to me!  No comments on the specs they are 🥇 .

I wonder if having `ShareDiagnosis.bt.spec` and `ShareDiagnosis.gps.spec` indicate that this screen wants to be pulled into the tracing strategy? and therefore we would have a `ShareDiagnosis` screen for each flavor of the app.  This would allow us to hoist the internal conditionals up and remove the `isGPS` checks altogether?  Either way, I would ship this pr and open a new one if you agree with that approach.

PR #1220, opened 2020-07-11
Accommodate key file processing limit on ios.

mattThousand:
### Why
We'd like the iOS client to process on 15 key files per day, according to the limits discussed in: https://github.com/google/exposure-notifications-server/issues/425

If we do not accommodate the limitations of the framework, exposure detection will fail.

### This Commit
- Tracks the number of key files processed over the course of 24 hours, ensuring that the ceiling of 15 is not exceeded
- Stores the last processed file url in order to continue processing files exactly where it left off when 24 hours have elapsed and processing capacity can be reset


octohub:
The `Test / unit-test-ios (pull_request)` failed. Just rerun it and it should be good to go. This has happened to my previous PRs as well. Any ideas how to fix it @mattThousand ? I want to avoid intermittently failing tests.
```
Failing tests:

Test session results, code coverage, and logs:
	/Users/runner/Library/Developer/Xcode/DerivedData/COVIDSafePaths-dnkqeqcpjaiydhhdlnadaetchitt/Logs/Test/Test-GPS_Development-2020.07.11_15-15-18-+0000.xcresult
	GPSTests:
		GPSSecureStorageTest.testOldLocationsTrimmed()


** TEST FAILED **
```

octohub (approved):

ios/BT/ExposureManager.swift
```diff
@@ -99,35 +90,37 @@ final class ExposureManager: NSObject {
...
       if progress.isCancelled {
         detectingExposures = false
         completionHandler?(GenericError.unknown)
       } else {
         switch result {
         case let .success(newExposures):
-          BTSecureStorage.shared.dateLastPerformedExposureDetection = Date()
           BTSecureStorage.shared.exposureDetectionErrorLocalizedDescription = .default
           BTSecureStorage.shared.exposures.append(objectsIn: newExposures)
+          BTSecureStorage.shared.urlOfMostRecentlyDetectedKeyFile = lastProcessedUrlPath
```
johnschoeman, line 122: Looking like we have a of related data being treated separately here? Is there a data clump we want to pull out here such that we save a single object to storage rather than 4 separate entries?

ios/BT/Extensions/Foundation/Date+Extensions.swift
```diff
@@ -2,4 +2,11 @@ extension Date {
   var posixRepresentation: Int {
     Int(timeIntervalSince1970) * 1000
   }
+
+  func difference(from date: Date, only component: Calendar.Component, calendar: Calendar = .current) -> Int {
```
johnschoeman, line 6: Should we be doing this calculation with a posix value?

ios/BT/Storage/BTSecureStorage.swift
```diff
@@ -7,7 +7,7 @@ final class BTSecureStorage: SafePathsSecureStorage {
   static let shared = BTSecureStorage()
 
   override var keychainIdentifier: String {
-    "org.pathcheck.bt.realm"
+    "\(Bundle.main.bundleIdentifier!).realm"
```
johnschoeman, line 10: 💯

johnschoeman (approved):
Overall looks good to me, but pretty heavy on the swift, so might makes sense to get eyes from another swift developer.

ios/BT/Extensions/Foundation/Date+Extensions.swift
mattThousand: The API is a bit misleading, but it's neat - see `Persisted` - it call through to `setUserState`, so per your instinct these are all properties on a single model in the database (`UserState`)

ios/BT/ExposureManager.swift
mattThousand: The API is a bit misleading, but it's neat - see Persisted.swift - it calls through to setUserState, so per your instinct these are all properties on a single model in the database (UserState)

ios/BT/Extensions/Foundation/Date+Extensions.swift
mattThousand: not in this case; we are comparing 2 `Date` objects (a `Date` stored in Realm and also a newly instantiated swift `Date`). Unless we are working with a date coming from the JS layer or a date that will eventually need to be sent up to the JS layer, we do not need to convert to a posix representation

PR #1226, opened 2020-07-13
Run Tests on GPS Release Branches

JacobJaffe:
Adds `gps-release**` to `Test` and `Validate` checks

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ on:
       - develop
       - master
       - bt-release
-      - gps-release
+      - gps-release**
 
 jobs:
   unit-test:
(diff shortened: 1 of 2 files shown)


tindn:
do we need to do the same for bt-release? or do they have a different workflow?

JacobJaffe:
> do we need to do the same for bt-release? or do they have a different workflow?

@johnschoeman for visibility -- thoughts?

@tindn this would need to be for both develop & the 1.2.0 branch

johnschoeman:
> > do we need to do the same for bt-release? or do they have a different workflow?
> 
> @johnschoeman for visibility -- thoughts?
> 
> @tindn this would need to be for both develop & the 1.2.0 branch

Lets not add this to bte-release at this point as we wont be actively using it for a while.  We will likely want something similar in the near future, but it makes sense as a separate pass, imo, thanks for asking!

johnschoeman (approved):

PR #1231, opened 2020-07-13
Fix screen background color on onboarding etc

amandabeiner:
### Testing Notes ###
- Run a fresh install of the app, and navigate through the onboarding
screens.
- Expect all screens to have legible copy
- Navigate through the affected user flow intro screens
- Expect all screens to have legible copy

 ### Fixed Issues ###
Fixes issue where text was not visible on some devices because it was a
white background with nearly white text.

#### Linked issues:
[Trello](https://trello.com/c/ms89BJfQ/180-as-a-user-i-want-to-be-able-to-see-all-text-on-the-screen-during-onboarding-process-so-i-know-what-i-am-agreeing-to)

Note that we may want to do another pass to make sure that the text
does not interact with background images on small devices with enlarged
accessibility text.

#### Screenshots:

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-13 at 17 46 17

Co-Authored-By: jeffstolz <jstolz123@gmail.com>
Co-Authored-By: tparesi <tparesi@gmail.com>

### DIFF
--- a/app/styles/layout.ts
+++ b/app/styles/layout.ts
@@ -5,9 +5,6 @@ import * as Spacing from './spacing';
 export const screenWidth = Dimensions.get('window').width;
 export const screenHeight = Dimensions.get('window').height;
 
-// iphone 8
-export const smallScreenWidth = 375;
-
 export const baseScreenPadding = Spacing.xSmall;
 
 export const xxSmallWidth = 0.05 * screenWidth;
--- a/app/views/common/ExplanationScreen.tsx
+++ b/app/views/common/ExplanationScreen.tsx
@@ -17,7 +17,6 @@ import { Typography } from '../../components/Typography';
 import {
   Buttons,
   Colors,
-  Layout,
   Spacing,
   Iconography,
   Typography as TypographyStyles,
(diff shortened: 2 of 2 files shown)


johnschoeman (commented):
LGTM!
its a bit of an odd pattern so nice to see it removed.

Are we sure this fixes the bug in the trello ticket? My impression was that it was the fact that the blueBackgroundImage was hanging on load? so the quick fix would be to replace it with a view and a `style={{backgroundColor: Colors.blueBackground}}` kinda thing?  And we could figure out the root cause for the issue of loading the image in a future commit?

also looks like this screen wants a `useStatusBarEffect('light-content')` as well?

tparesi:
@johnschoeman Hasn't the status bar always worked as it is now?

We were able to reproduce the bug on an SE and a Pro - and this change fixed both of those issues.

jeffstolz:
> LGTM!
> its a bit of an odd pattern so nice to see it removed.
> 
> Are we sure this fixes the bug in the trello ticket? My impression was that it was the fact that the blueBackgroundImage was hanging on load? so the quick fix would be to replace it with a view and a `style={{backgroundColor: Colors.blueBackground}}` kinda thing? And we could figure out the root cause for the issue of loading the image in a future commit?
> 
> also looks like this screen wants a `useStatusBarEffect('light-content')` as well?

Yup, we checked and it fixes the issue that was occurring for all phones with a smaller width than iPhone11 (including iPhoneX) - wasn't a render issue, the ternary was just removing the bg image on these devices that had a slightly smaller device width. We feel like this is pretty stable now that it's removed, but if we decide to remove the image entirely and go with a backgroundColor we'll want to do that in other places as well that use the gradient background, and get design approval for that.

You cool with us addressing the status bar in a later commit since it isn't related to this ticket specifically?

johnschoeman:
> Yup, we checked and it fixes the issue that was occurring for all phones with a smaller width than iPhone11 (including iPhoneX) - wasn't a render issue, the ternary was just removing the bg image on these devices that had a slightly smaller device width. We feel like this is pretty stable now that it's removed, but if we decide to remove the image entirely and go with a backgroundColor we'll want to do that in other places as well that use the gradient background, and get design approval for that.

got it, thanks for the explanation, makes sense to me!

> You cool with us addressing the status bar in a later commit since it isn't related to this ticket specifically?

yeah of course, i'll defer to your judgement on that.
(oops i thought i had already approved.)

johnschoeman (approved):

PR #1223, opened 2020-07-13
Fetch verification certificate

johnschoeman:
Why:
After verifying the users provided code to the verification server and receiving a token, we need to generated a hashed version of the users keys (and potentially metadata) and post this to verification server to receive a certificate which we can use to post exposure keys to the key server.

This commit:
Introduces the logic for generating and HMAC of the current exposure keys and posting this data along with the previously received token to api/certificate.

We introduced react-native-simple-crypto for calculating the HMAC hashing on the JS side as this will reduce the amount of code that the native layer will need to implement. In a future iteration, it might make sense to do all of the cryptographic logic in the native layer as this will allow us to remove the dependency react-native-simple-crypto

A native module promise, `ExposureKeyModule.fetchExposureKeys`, was added to fetch the exposure keys and serialize them to send the encrypted payload to the verification server.

Next Steps:
- Save the hmacKey to `realm`
- Save the certificate generated on the verification server to `realm`
- Post the exposure keys to the `GAEN` with the certificate and the HMAC key
- Handle errors on the exposure keys, post exposure data flow
- Add in user metadata to the request that goes to the Verification Server

Co-Authored-By: Alejandro Dustet<aledustet@gmail.com>

PR #1232, opened 2020-07-13
Publish keys with cert to exposure key server

johnschoeman:
Publish keys with cert to exposure key server

Why:
----
We need to send the verification credentials generated by the verification server to the exposure key server so that the exposure keys provided by the device can be cross verified by the servers. This process requires communication from the verification server flow to the exposure key server flow.

This commit:
----
Adds the generated hmac key and certificate, added in a prior commit #1223 , to the submitDiagonsisKeys method on the native module. We updated the PublishConsentScreen to handle delegating this to the Native Layer. We also removed the stubbed logic for saving the HMAC key to realm as this is not required for the at this point.

Co-Authored-By: Alejandro Dustet <alejandro@thoughtbot.com>

submit-keys


app/bt/AffectedUserFlow/PublishConsent/PublishConsentScreen.tsx
```diff
@@ -0,0 +1,31 @@
...
+const PublishConsentScreen: FunctionComponent = () => {
+  useStatusBarEffect('light-content');
+  const navigation = useNavigation();
+  const { certificate, hmacKey } = useAffectedUserContext();
+
+  if (hmacKey && certificate) {
+    return <PublishConsentForm hmacKey={hmacKey} certificate={certificate} />;
+  } else {
+    return (
+      <View>
```
johnschoeman, line 18: This we will want to take a second pass at in a future commit.

mattThousand (approved):

PR #1238, opened 2020-07-14
Bypass Export API Feature Flag

JacobJaffe:
Adds a feature flag to bypass the API checks on the export flow. This allows for easy QA testing of the export flow:
https://pathcheck.atlassian.net/browse/SAF-809

Also refactors the feature flag screen / settings screen to visually group feature flag settings cleaner, and separates Google Import vs. JSON API import.

image
image
image


app/gps/Export/ExportCodeInput.js
```diff
@@ -198,6 +205,16 @@ export const ExportSelectHA = ({ route, navigation }) => {
               label={t('common.next')}
               onPress={validateCode}
             />
+            {/* Bypass API to allow easy testing of screens */}
+            <FeatureFlag flag={FeatureFlagOption.BYPASS_EXPORT_API}>
```
tindn: Just a thought, since this will bypass the upload step anyway, should we just always validate the code instead of having a second button?

JacobJaffe: 💯 That makes a lot more sense given how the other screens bypass on the same buttons

vlasac (approved):

tindn (approved):

johnschoeman (approved):
Code looks good to me! one thought that pops to mind, is that the settings screen is diverging quite a bit. i might make sense to have this specific to the tracing strategy, but definitely a thing for a future commit.

PR #1240, opened 2020-07-14
Move the constants/authorities file to .env

tparesi:
#### Description:

WHY: 
The contents of the authorities.ts file were following a similar pattern as the .env files. Also, some of the contents were being used for the GPS strategy and some for the BT strategy. 

We moved the contents of the constants/authorities.ts into the appropriate .env file (GPS or BT).

#### Linked issues:

[Trello](https://trello.com/c/L4pNhsN4/71-as-a-dev-team-we-have-a-way-to-customize-the-health-authority-name-in-the-app)
[Pathcheck Mobile Resources PR](https://github.com/Path-Check/pathcheck-mobile-resources/pull/1)

### DIFF
--- a/.env.gps
+++ b/.env.gps
@@ -9,6 +9,7 @@
 
 TRACING_STRATEGY=gps
 AUTHORITIES_YAML_ROUTE=https://raw.githubusercontent.com/Path-Check/trusted-authorities/master/staging/authorities.1.0.1.yaml
+AUTHORITIES_LIST_URL=https://raw.githubusercontent.com/Path-Check/safeplaces-frontend/develop/healthcare-authorities.yaml
 
 flag_google_import=true
 flag_custom_url=true
--- a/App.js
+++ b/App.js
@@ -3,7 +3,7 @@ import { MenuProvider } from 'react-native-popup-menu';
 import SplashScreen from 'react-native-splash-screen';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
-import Config from 'react-native-config';
+import env from 'react-native-config';
 import 'array-flat-polyfill';
 
 import { Entry } from './app/Entry';
@@ -16,7 +16,7 @@ import BackgroundTaskService from './app/services/BackgroundTaskService';
 import { isGPS } from './app/COVIDSafePathsConfig';
 
 const determineTracingStrategy = () => {
-  switch (Config.TRACING_STRATEGY) {
+  switch (env.TRACING_STRATEGY) {
     case 'gps': {
       return gpsStrategy;
     }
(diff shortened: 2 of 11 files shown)


app/views/assessment/endScreens/Share.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import env from 'react-native-config';
```
johnschoeman, line 3: I think we are using `Config` elsewhere in the app for this, I personally prefer the `env`, but wanted to point this out if you have thoughts on being consistent. As in should we change elsewhere from Config to env?

johnschoeman (approved):
LGTM,
if you could also make sure to update the mobile-resources git project? and of course give a spot check that the gps flows still work.

app/views/assessment/endScreens/Share.js
tparesi: We liked env better too. I like that change - will take care of that.

.env.gps
```diff
@@ -9,6 +9,7 @@
 
 TRACING_STRATEGY=gps
 AUTHORITIES_YAML_ROUTE=https://raw.githubusercontent.com/Path-Check/trusted-authorities/master/staging/authorities.1.0.1.yaml
+AUTHORITIES_LIST_URL=https://raw.githubusercontent.com/Path-Check/safeplaces-frontend/develop/healthcare-authorities.yaml
```
JacobJaffe, line 12: This should be removed from the app altogether, this is an outdated list (prior to MVP1)

app/services/HCAService.js
```diff
@@ -12,6 +12,8 @@ import { GetStoreData, SetStoreData } from '../helpers/General';
 import languages from '../locales/languages';
 import LocationService from './LocationService';
 
+const { AUTHORITIES_LIST_URL } = env;
+
 /**
  * Singleton class to interact with health care authority data
```
JacobJaffe, line 18: I don't believe this class is actually used anymore. (The methods are used in the permissions context, but I believe everything ends as a no-op)

.env.gps
tparesi: Sweet - will take care of that.

app/services/HCAService.js
tparesi: Okay - great to know. I will see if removing it makes sense in this PR, if not, we can remove it in a follow up PR.

johnschoeman: great, we should remove the dead code then.

tparesi:
@JacobJaffe Please take a second look, as we removed the HCAService from the GPS PermissionsContext

app/services/HCAService.js
tparesi: I will do this on a new branch/PR

App.js
```diff
@@ -3,7 +3,7 @@ import { MenuProvider } from 'react-native-popup-menu';
 import SplashScreen from 'react-native-splash-screen';
 import { Provider } from 'react-redux';
 import { PersistGate } from 'redux-persist/integration/react';
-import Config from 'react-native-config';
+import env from 'react-native-config';
```
aledustet, line 6: niceee

aledustet (approved):
:shipit:

PR #1243, opened 2020-07-14
HT Prod Language

JacobJaffe:
Align with actual release. We currently have HT live, not PR.

### DIFF
--- a/app/locales/languages.js
+++ b/app/locales/languages.js
@@ -73,7 +73,7 @@ export async function setUserLocaleOverride(locale) {
 /* eslint-disable no-underscore-dangle */
 const PROD_RESOURCES = {
   en: { label: en._display_name, translation: en },
-  es_PR: { label: es_PR._display_name, translation: es_PR },
+  ht: { label: ht._display_name, translation: ht },
 };
 
 /** Languages only available in feature flag. */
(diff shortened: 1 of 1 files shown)


tindn (approved):

app/locales/languages.js
```diff
@@ -73,7 +73,7 @@ export async function setUserLocaleOverride(locale) {
 /* eslint-disable no-underscore-dangle */
 const PROD_RESOURCES = {
```
johnschoeman, line 74: This constant and the prod vs dev setup strikes me as a bit odd. And it looks like we will get into a situation where we will need to update the code for every client?

I would ship as is for sake of time, but it will makes sense to take another pass at this. especially considering for bt side we will have specific languages we know we will want to default to.

Ideally we would be choosing the devices locale + our defaults of en and es_PR or something like that and not need to consider dev vs prod, or this HA vs that HA, or gps vs bt vs wifi vs university, etc.

johnschoeman (approved):
:shipit:

app/locales/languages.js
JacobJaffe: Agree on the motivation to change default locales. One thing to note here is we're gating on _ready_ languages right now, to avoid putting in new languages until they reach a majority coverage

PR #1244, opened 2020-07-14
Remove HCAService - no longer used

tparesi:
#### Description:

Remove no longer used HCAService that was returning a no-op.


app/views/onboarding/LocationsPermissions.tsx
```diff
@@ -30,13 +30,12 @@ import {
 const LocationsPermissions = (): JSX.Element => {
   const { t } = useTranslation();
   const dispatch = useDispatch();
-  const { authSubscription, location } = useContext(PermissionsContext);
```
tparesi, line 33: Not sure if this is needed?

johnschoeman: by this, do you mean `location`? if so i believe yes.

johnschoeman (approved):
Removing dead code! nice!  LGTM, but i would definitely get eyes from someone familiar with the GPS flow.

app/views/onboarding/LocationsPermissions.tsx
tparesi: Sorry, I meant to only remove the authSubscription here. Fixing!

tparesi: No, I meant the authSubscription inside the LocationsPermissions. Below, I kept the location stuff still.

JacobJaffe (approved):
Thanks for cleaning this up!

PR #1246, opened 2020-07-14
Download Proper Key Index File and Limit to 20 Key Batches Per Day

octohub:
#### Description:

Downloads proper index index file instead of a test one. Also limits key batches to 20 per day which is the Android GAEN limit.

#### Linked issues:

https://trello.com/c/8j1k6Ss2

#### How to test:

1. Fresh install Android BT app.
2. Enable EN.
3. Observe logcat to watch progress of key batches downloaded and submitted to the GAEN.


mattThousand:
Looks great, just one question - what do you think about storing the url in Realm rather than SharedPreferences?

mattThousand (approved):
LGTM!

PR #1247, opened 2020-07-14
Avoid running Intersect logic on BT flavor

aledustet:
Reviewers:
----
This problem was brought to our attention because of a warning that was showing up every time we were building the app with the `bt` configuration. This was the error we were seeing:

```
WARN  Possible Unhandled Promise Rejection (id: 0):
TypeError: null is not an object (evaluating '_reactNative.NativeModules.SecureStorageManager.getLocations')
```

This happens because the native module responsible for that logic`SecureStorageManager` is not available on the `bt` build.
We corrected the first occurrence of this error, happening on app load after checking that the app was on the `gps` configuration before starting the background service. After we closed the app, the error kept showing up, because the middleware `onChangedSelectedHealthAuthorities` was also starting a check on the intersect service.

This solution doesn't feel great so let's take this PR as a discussion the approach. I'm in no means a redux expert so apologies in advance if this is atrocious 😅  

Why:
----
The store applies a middleware to check for changes on the health authority and then it runs the `IntersectService` pertinent to check for exposures with the gps detection logic. This middleware gets applied regardless of what the logic is in the app and we are seeing errors because the native module used for this purpose is not available on the `bt` build.

This PR:
----
- Move the creation of the store to a method to accept middleware as an option
- Add middleware to the tracing strategy


amandabeiner (approved):
This LGTM! Pair-reviewed on this with @tparesi, and we think it makes sense that we will need different middlewares for the different app flavors and that that would require changes to the redux middlewares.

Our only question was whether it was possible to keep the store instantiation inside the store/index.ts file, but it appears that that would require duplicating the `determineTracingStrategy` logic, which is likely not worth the tradeoff.

aledustet:
> This LGTM! Pair-reviewed on this with @tparesi, and we think it makes sense that we will need different middlewares for the different app flavors and that that would require changes to the redux middlewares.
> 
> Our only question was whether it was possible to keep the store instantiation inside the store/index.ts file, but it appears that that would require duplicating the `determineTracingStrategy` logic, which is likely not worth the tradeoff.

Yeah, I started by having the `isGPS` in there but felt even worse than this. Not sure also if having a middleware that runs after every change is necessary but I'm not aware of the requirements in that workflow. Maybe we could remove the middleware and call the intersect service on demand but I'm not sure how many ways are there to change the selected HA.

app/store/index.ts
```diff
@@ -31,6 +32,18 @@ const persistConfig = {
 
 const persistedReducer = persistReducer(persistConfig, rootReducer);
 
-export const store = createStore(persistedReducer, {}, enhancers);
-StoreAccessor.setStore(store);
-export const persistor = persistStore(store);
+export const createStoreAndPersistor = (
```
JacobJaffe: I don't think this should export the `createStoreAndPersistor` since this is meant to be a one time operation, given the `StoreAccessor`; I think we should stick to exporting the singleton

JacobJaffe (commented):
I'm wary of adding in parameters to the store creation. IMO if these need to heavily diverge BT / GPS, a separate GPS / BT store would make more sense, or configs for them. Following this logic of moving the middleware out of the `app/store/index.ts`, more configured logic (enhancers, persist config, etc.) would become more separate from the actual store logic. 

I'd opt for just gating the `store/index` file on the specific middleware to have or not have here, rather than passing in the general configs, as we don't know currently how much that would need to be extended. That way, it will be easier to move from a single line of code in the store initialization than from spread out logic in multiple files that configures the store if / when more configuration fidelity is determined needed

app/store/index.ts
aledustet: I think this makes way more sense, thanks! I took a hack at moving the initialization to this class and pass config to it, right now is just a flag but it could become an object with a config if it gets hairier than the one middleware.

app/store/StoreAccessor.ts
```diff
@@ -15,6 +43,29 @@ class StoreAccessor {
   getStore = () => {
     return this.store;
   };
+
+  initialize = (withIntersectMiddleware: boolean): [Store, Persistor] => {
```
aledustet, line 47: Took a hack at moving the initialization here, right now is just a flag but I think we could extend it to a config object, particular to each strategy if the requirements change. Right now it just diverges on the one middleware so having a flag here looked okayish. I think it is an improvement from my first shot at it. Sorry to put you all through this but your input is amazing. So I am selfishly soliciting more of it. 😄

App.js
```diff
@@ -47,6 +49,9 @@ const App = () => {
     isGPS && BackgroundTaskService.start();
   }, []);
 
+  const store = isGPS ? gpsStore : btStore;
```
aledustet, line 52: Took a hack at moving the initialization here, right now is just using a flag but I think we could extend it to a config object, particular to each strategy if the requirements change. Right now it just diverges on the one middleware so having a flag here looked okayish. I think it is an improvement from my first shot at it. Sorry to put you all through this but your input is amazing. So I am selfishly soliciting more of it. 😄

johnschoeman (commented):
This strikes me as adding a bit of complexity that we may be able to avoid altogether. Im not sure there is anything in the BT side that wants to live in redux so im hesitant to add a redux store specifically for BT.  Could we conditionally render the store altogether?  It might make sense to hold off on this until we have some direction of the GPS vs BT questions.

aledustet:
> This strikes me as adding a bit of complexity that we may be able to avoid altogether. Im not sure there is anything in the BT side that wants to live in redux so im hesitant to add a redux store specifically for BT. Could we conditionally render the store altogether? It might make sense to hold off on this until we have some direction of the GPS vs BT questions.

Feeling more and more like this, the only use of redux common to both flavors is holding state for the onboarding flow, so we might be able to move past it by saving one value to our async storage. This seemed like a high priority item prior to getting a decision on how to handle better both flavors so I'm going to close this one to keep it from getting attention on the PR list, please feel free to add more color or context here as we might revisit this approach depending on future developments.

PR #1248, opened 2020-07-14
Add timestamp for last detection

tparesi:
Why:
----
On the exposure history screen, the header is displaying the amount of days on the calendar but not the time since the exposure history detection process was run.

This PR:
----
- Add a module for exposing the last detection date from the native side
- Add component to display the time passed since the last exposure
detection happened

Reviewers:
----
From the debug menu, tap on the `Detect Exposures Now` button and then when you navigate to the history screen, the time since the last exposure detection was run should be a few seconds ago.

[Trello](https://trello.com/c/mF6fMBd5/42-as-a-user-when-i-am-on-the-exposure-history-screen-i-see-a-not-of-the-last-time-that-the-history-was-updated-which-is-correct)


app/bt/nativeModule.ts
```diff
@@ -26,6 +27,16 @@ export const subscribeToExposureEvents = (
   );
 };
 
+export const fetchLastExposureDetectionDate = async (): Promise<Dayjs | null> => {
```
johnschoeman:
```suggestion
export const fetchLastExposureDetectionDate = async (): Promise<Posix | null> => {
```

What do you think about keeping posix as the data model and waiting to convert to human readable text in the view layer?

app/views/ExposureHistory/DateInfoHeader.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { render, wait } from '@testing-library/react-native';
+import dayjs from 'dayjs';
+import '@testing-library/jest-native/extend-expect';
+
+import DateInfoHeader from './DateInfoHeader';
+import * as nativeModule from '../../bt/nativeModule';
+
+jest.mock('../../bt/nativeModule');
+
+describe('DateInfoHeader', () => {
```
johnschoeman, line 11: Nice! I love the spec!

app/views/ExposureHistory/DateInfoHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { Typography } from '../../components/Typography';
+import { Typography as TypographyStyles } from '../../styles';
+import { fetchLastExposureDetectionDate } from '../../bt/nativeModule';
+import ExposureHistoryContext from '../../ExposureHistoryContext';
+
+const DateInfoHeader: FunctionComponent = () => {
+  const [
+    lastExposureDetectionDate,
+    setLastExposureDetectionDate,
+  ] = useState<Dayjs | null>(null);
```
johnschoeman:
```suggestion
  ] = useState<Posix | null>(null);
```

app/views/ExposureHistory/DateInfoHeader.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { t } = useTranslation();
+  const { exposureHistory } = useContext(ExposureHistoryContext);
+
+  useEffect(() => {
+    fetchLastExposureDetectionDate().then((exposureDetectionDate) => {
```
johnschoeman: I tend to try to keep IO as close to the edge of the system as possible as well as keeping the view layer and data layer independent, i.e. functional core / imperative shell. What do you think about having this call happen in the ExposureHistoryContext?  And having the view component consume this data from the context?

app/views/ExposureHistory/History.tsx
```diff
@@ -74,11 +75,7 @@ const History = ({ exposureHistory }: HistoryProps): JSX.Element => {
             </TouchableOpacity>
           </View>
           <View style={styles.headerRow}>
-            {!isGPS ? (
-              <Typography style={styles.subHeaderText}>
-                {lastDaysText}
-              </Typography>
-            ) : null}
+            {isBT ? <DateInfoHeader /> : null}
```
johnschoeman: is there a way we can do this without this conditional?  Does the GPS side have a similar feature requirement?

ios/BT/bridge/ENExposureModule.m
```diff
@@ -0,0 +1,19 @@
...
+#import "BT-Swift.h"
+
+@interface ENExposureModule: NSObject <RCTBridgeModule>
+@end
+
+@implementation ENExposureModule
+
+RCT_EXPORT_MODULE();
+
+RCT_REMAP_METHOD(fetchLastDetectionDate,
```
johnschoeman: Do we need to introduce a new module for this? does this want to live in an existing module?

johnschoeman (approved):
Looks good! I had a few dataModel type thoughts.

1. Lets use Posix for all internal date representations, `type Posix = number` and wait to convert to a Dayjs when we present to the viewer?

2. Lets avoid having IO and app state management in leaf nodes or presentational components, as keeping the coupling between the view layer and the data model low will be extremely helpful down the line as these screens are highly likely to have a lot of change coming down the line.

app/views/ExposureHistory/History.tsx
johnschoeman: if we do the call in the Exposure history context, and have the component not render if there isnt the data, then we should be able to remove the conditional altogether.  If the GPS side eventually wants to use it, they simply need to provide a posix of the last updated at time.

app/views/ExposureHistory/DateInfoHeader.tsx
aledustet: I like this idea, however, if we separate this to the context we are potentially running a process that gets info from the native side surrounding a number of elements that potentially don't interact with it. So looking it from an encapsulation perspective, if we put this in the context it feels to me that we are saying, `this piece is necessary for the functioning of this block` when in reality we could live without it. Not feeling strongly either way just curious about how the encapsulation stands off with the data vs view layer organization.

app/views/ExposureHistory/History.tsx
aledustet: I think that calling something regardless could cause a warning like the one we were seeing on the middleware for the intersect service since this module won't be bundled on the GPS build. We could use the conditional on the `existence` of the module perhaps?

ios/BT/bridge/ENExposureModule.m
aledustet: Maybe exposure history is a fit? What do you think @mattThousand Will move there in the meantime

app/views/ExposureHistory/History.tsx
aledustet: I think we can make it conditionally render now that we have a value depending on an actual module response only available on the BT flavor. Good catch!

app/views/ExposureHistory/DateInfoHeader.tsx
aledustet: Ended up moving it to the context, let me know what you think 👍

amandabeiner (approved):
Looks great to me, and ❤️  the specs!!

PR #1252, opened 2020-07-15
Fix/tracing strategy bt

### LINKED ISSUE #467: Multiple flavors for android build config results in empty `Config` object on the react native side
We are using RN => 0.61.5, and the latest version of `react-native-config` 

This is an open-source project, here is [the repo]. There is a split for two flavors of the applications, one BT and one GPS, on the iOS version, the pre-build scripts copying the right env file on the right Scheme works and each version of the app loads the correct set of `env` variables. When the Android version builds for the GPS flavor the environment variables are made available on the `react-native` side and everything works. Is worth noticing that the default and original configuration was made for the GPS side, using the original `appId`. 
When the `BT` flavor builds, if we spoof inside the `dotenv.gradle` the `buildConfig` is picked up and the right env file is found, we can see it in the build output and if we print the `env` that is being used inside that file it has all the right values. However, as soon as the app builds our `Config` imported from `react-native-config` is empty.

Things we have tried:
----
- We added the appId for the BT build to the `proguard-rules`
- We tried leaving the `buildConfig` alone and only using one .env file
- We hardcoded the appID across the manifests inside the android app
- We switch the MainActivity to just point to the BT version appID

[the repo]: https://github.com/Path-Check/covid-safe-paths

### LINKED ISSUE #1221: Path Check BT crashes when executed
## Issue type

Environment

## Version 
- Release build version: 
1.22.4
## Affected Devices

| Brand        |  Model        | OS  |
| ------------ |:-------------:| ----:|

Samsung Galaxy S8 emulator
Samsung Galaxy S9 device

## Steps to reproduce
yarn run-android-bt

## Expected Behavior
Normal onboarding

- When I click `Stop Logging` I am taken to the `Import` screen. Expect to
  go back to the Welcome greeting.

image

crayne:
#### Description:

Fixes crash on launch of bt build variant because .env.bt is not read correctly 
#### Linked issues:

Fixes #1221 
Fixes luggit/react-native-config#467

#### How to test:

yarn run-android-bt -- no crash in App.js


android/app/build.gradle
```diff
@@ -161,6 +161,10 @@ android {
 
         versionCode 11
         versionName "1.0.8"
+
+        //Fix problem in bt where .env.bt is not recognized
+        resValue "string", "build_config_package", "org.pathcheck.covidsafepaths"
```
aledustet, line 166: Oh wow, this is great!!! just tested it and it definitely fixes the problem. If you have the time would you mind educating me on what the issue was(totally up to you 😄 )? Also, we had an open issue on the react-native-config package https://github.com/luggit/react-native-config/issues/467 and this fixes that question we opened if you are ok with linking this PR to the issue we can close it right away. I'll wait a bit for some other folks to check it out but I'm more than happy with merging this. 🎉🎉🎉🎉🎉

aledustet (approved):
Works like a charm 👏👏👏👏 🥳🥳🥳🥳

This PR also addresses the issue we have open: https://github.com/luggit/react-native-config/issues/467

android/app/build.gradle
crayne: I set a breakpoint in the react-native-config file ReactNativeConfigModule.java.  I saw that it was trying to get the Config constants from org.pathcheck.covidsafepathsBt.BuildConfig, which didn't exist.  So I tried changing the path to org.pathcheck.covidsafepaths.BuildConfig, and that worked.  Then I remembered this from the react-native-config README:
````
In `android/app/build.gradle`, if you use `applicationIdSuffix` or `applicationId` that is different from the package name indicated in `AndroidManifest.xml` in `<manifest package="...">` tag, for example, to support different build variants:
Add this in `android/app/build.gradle`

```
defaultConfig {
    ...
    resValue "string", "build_config_package", "YOUR_PACKAGE_NAME_IN_ANDROIDMANIFEST.XML"
}
````
I didn't see that the condition that was stated applied, but I decided to try it anyway, and I was lucky

johnschoeman (approved):
Amazing, thank you!

crayne:
No problem

octohub:
@crayne Great work! Currently looking into why the `build_staging_android_gps` and `build_staging_android_bt` build checks are failing. If you have any idea @crayne please chime in!

octohub:
@crayne Can you please rebase/merge latest from develop? I think that might fix the build issue.

octohub:
@crayne Actually, I believe it's because your are submitting a PR from a fork which lacks access to `Path-Check/pathcheck-mobile-resources`. Update: Looking into a solution.

crayne:
No change after I merge the latest from develop.

aledustet:
> No change after I merge the latest from develop.

Thanks for trying that, nothing else to do on your end, seems that the failure is related to a configuration on the CI. We're trying to get it fixed so we can add this commit in there.

octohub:
We don't have a fix at this time, but @aledustet ran this build and it passed CI:
https://github.com/Path-Check/covid-safe-paths/pull/1260

So when the time comes we will be able to merge it. @JacobJaffe Can you give final sign off on this? I want to make sure it doesn't introduce any regressions for you on the GPS side.

crayne:
Did the test work because it was built from the main repo, not from a fork?  If this is true does it mean anything else I do on a fork will have the same problem?

johnschoeman:
> Did the test work because it was built from the main repo, not from a fork? If this is true does it mean anything else I do on a fork will have the same problem?

For some context here, the submodule for the mobile-resources that was added about a week ago broke ci for prs from forks. The quick fix here is to either turn off the required flag on those tests, or open a new pr with the commit from the main repo.

@mattThousand , what are your thoughts on this one? should we remove the submodules and take another pass at this in a future date so we unblock external contributions?

summetj:
> > Did the test work because it was built from the main repo, not from a fork? If this is true does it mean anything else I do on a fork will have the same problem?
> 
> For some context here, the submodule for the mobile-resources that was added about a week ago broke ci for prs from forks. The quick fix here is to either turn off the required flag on those tests, or open a new pr with the commit from the main repo.
> 
> @mattThousand , what are your thoughts on this one? should we remove the submodules and take another pass at this in a future date so we unblock external contributions?

Currently, the CONTRIBUTING.md document asks developers to fork the repo and develop off of that, so it would be nice if people who follow those instructions can contribute functional PRs.

crayne:
I agree with Jay.  Otherwise, any changes made to a fork will need to be re-created on the main repo.  The other alternative would be to do away with the fork requirement.

octohub:
Agreed @summetj and @crayne ! We are working this out internally and will keep you posted. Thank you!

johnschoeman:
As an update: I believe we have the contributions from a fork issue resolved, but now that we have the repo split, this pr is no longer applicable so closing.  cc: @summetj , @crayne

PR #1253, opened 2020-07-15
Import location hashes on iOS

jvisenti:
#### Description:

Fixes hashes being ignored on import on iOS

#### Linked issues:

https://pathcheck.atlassian.net/browse/SAF-813


tnekret (approved):
Tested it and it works!

PR #1254, opened 2020-07-15
re-added hashes to exported file

tnekret:
This PR re-adds hashes to exported file as discussed with Diarmid since it's only used in testing environment.


tindn (approved):

JacobJaffe (approved):
LGTM!

PR #1256, opened 2020-07-15
Await Location Permissions Before Completing Onboarding

yonatanmk:
#### Description:
Currently the app progresses to the main screen from Onboarding when the user presses the Enable Location button. The app should progress after the user makes a selection in the Enable Location Popup that appears after pressing the Enable Location button.

#### How to test:
Go through onboarding and enable location. The app should wait for you to enable/disabled location permissions before navigating to the main screen.


JacobJaffe (approved):
💯

PR #1259, opened 2020-07-15
Disable insecure http requests except in debug builds

jvisenti:
https://pathcheck.atlassian.net/browse/SAF-783

**NOTE:** HTTP requests are enabled for debug builds to support downloading the bundle from the local metro server.


JacobJaffe (approved):
Terrific. Would we want this included in 1.2.0 as well?

jvisenti:
@JacobJaffe I'm not sure. It's described as:
> Slightly less urgent than mainline functional “it doesn’t work” bugs, but not far behind.

Let's merge this and cherry pick into 1.2.0 if needed?

PR #1261, opened 2020-07-15
MB iOS EN Updates

mattThousand:
### Why
We'd like testers to be able to read information that is helpful during debugging rather than opaque errors. We'd also like more of the iOS exposure detection logic to have unit test coverage.

### This Commit
This commit changes error handling such that relevant information (the number of files processed during exposure detection, the number of keys sent during key submission) are bubbled up to the JS layer and ultimately made visible to the user. This commit also adds test coverage for the logic that filters out keys older than 350 hours during key submission.

Screenshot 2020-07-15 at 3 47 01 PM
IMG_526E3C3A5899-1
Screenshot 2020-07-15 at 3 47 01 PM

### DIFF
--- a/ios/COVIDSafePathsTests/ExposureManagerTests.swift
+++ b/ios/COVIDSafePathsTests/ExposureManagerTests.swift
@@ -1,7 +1,8 @@
 import Foundation
-
+import  ExposureNotification
 import RealmSwift
 import XCTest
+
 @testable import BT
 
 class ExposureManagerTests: XCTestCase {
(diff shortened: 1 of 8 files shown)


ios/COVIDSafePathsTests/ExposureManagerTests.swift
```diff
@@ -244,6 +245,29 @@ mn/1593460800-1593475200-00022.zip
     XCTAssertEqual(BTSecureStorage.shared.userState.remainingDailyFileProcessingCapacity, Constants.dailyFileProcessingCapacity)
   }
 
+  func testFilterOldKeysForSubmission() {
+    // Keys older than 350 hours should not be submitted
```
aledustet, line 249: How do we feel about including some of this comment on the test name? Would it be helpful for a failure error to read the context of the comment also?

aledustet (approved):
🚢 :shipit:

PR #1262, opened 2020-07-15
Include hmac and certificate in post key payload

johnschoeman:
Why:
We would like to include the hmac key and jwt certificate in the post
diagnosis keys call to the verification server. Previously we were not
including these due to an error in the verification server, which has
recently been resolved.


aledustet (approved):
🚢 :shipit:

mattThousand (approved):
:shipit: 💯

PR #1263, opened 2020-07-16
iOS Descriptive Error for Background Failures 

mattThousand:
### Why
We'd like to know specific information about errors when they occur in the background

### This Commit
This commit surfaces more detailed error information in the push notification that appears in debug when an error occurs in the background during exposure detection on iOS


jvisenti (approved):

johnschoeman (approved):
LGTM

PR #1270, opened 2020-07-16
Display self assessment according to jurisdiction requirements

zhaccc:
#### Description:

This PR hides the self-assessment feature according to jurisdiction requirements. 
Also, we hide the next `Next Steps` in case we don't have HA advice URL or if we have self-assessment hidden.

#### Screenshots:

**Hidden:**

**Displayed:**


app/COVIDSafePathsConfig.ts
```diff
@@ -1,3 +1,6 @@
 import Config from 'react-native-config';
+import { AUTHORITY_ADVICE_URL } from './constants/authorities';
```
tparesi: Heads up: I recently moved AUTHORITY_ADVICE_URL to the env.bt file (accessed through the config - now named env, variable) and merged that PR in

app/views/ExposureHistory/ExposureDatumDetail.tsx
```diff
@@ -126,6 +123,27 @@ const NoDataExposureDetail = ({
   );
 };
 
+interface NextStepsButtonProps {
```
tparesi, line 134: Should we move this component to a new file?

tparesi (changes requested):
Have to rebase with develop for the changes to AUTHORITY_ADVICE_URL and then I think this looks good.

app/COVIDSafePathsConfig.ts
zhaccc: Thanks for the heads up! Noticed it when I pulled in the develop branch. I fixed the merge conflicts, should be all good now!

tparesi: Not seeing those changes pushed yet, fyi

app/views/ExposureHistory/ExposureDatumDetail.tsx
zhaccc: It's a pretty small change and since there are two more related components in this file I figured it is ok to leave it here for now.

tparesi:
@zhaccc We may want to move this over to the new repo - right?

PR #1272, opened 2020-07-16
Exposure History Bugfixes iOS

mattThousand:
### Why
We'd like the Exposure History screen to accurately display exposure dates and durations

### This Commit
Fixes two bugs:
1. Exposure history was not fetching "current exposures" on launch. We add a function invocation inside a `useEffect` hook that disappeared in a recent merge.
2. Individual exposures did not account for the fact that Apple's Exposure Notification framework reports durations rounded to 5 minute intervals with a 30 minute maximum.


app/ExposureHistoryContext.tsx
```diff
@@ -88,6 +88,10 @@ const ExposureHistoryProvider: FunctionComponent<ExposureHistoryProps> = ({
     return subscription.remove;
   }, [exposureInfoSubscription, toExposureHistory]);
 
+  useEffect(() => {
+    getCurrentExposures();
+  }, [toExposureHistory]);
```
aledustet: I believe that this one will give you a warning that it needs the signature of the callback also as a dependency, sounds weird but I remember seeing such warnings 😅  probably when you do that the linter will tell you that the function needs to be defined before you use it so moving the method declaration before this hook is likely something you'll need to do 😬

aledustet (approved):
I think this will have a linting warning but we can definitely take care of it later 🚢 :shipit:

johnschoeman (approved):

PR #1273, opened 2020-07-16
Script to update display name based on environment

aledustet:
Why:
----
Each HA will have different settings for our builds, one of those is the
display name. We need a way to update those values for each one of the
builds.

This Commit:
----
- Extend setup env script to update the display name of the android and
the ios app for the bt builds

Reviewers:
----
This script is counting on having `xmllint` and `PlistBuddy` in the
executing environment. Given that builds need to happen on `MacOS` those
two libraries are part of the system. The solution is very rigid but it
is a start on how we might tackle the configurations per HA.

### DIFF
--- a/setup_env.sh
+++ b/setup_env.sh
@@ -1,4 +1,11 @@
-#!/bin/bash                                                                                                         
+#!/bin/bash
+
+# USAGE
+# First argument is the name of the health authority to build
+# Flags:
+# --renaming: Puts the display name provided on the .env.bt files on the android
+# app for ios and android bt builds
+# THIS REQUIRES THE xmllint and the PlistBuddy utilities to be present
 
 HA=$1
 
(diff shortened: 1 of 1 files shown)


setup_env.sh
```diff
@@ -24,4 +31,35 @@ if [ ! -f pathcheck-mobile-resources/environment/$HA/.env.bt.release ]; then
...
\ No newline at end of file
+cp -r pathcheck-mobile-resources/environment/$HA/. ./
+
+# In order to read properly spaces on environment variables
+IFS='
+'
+
+if [ "$2" = "--renaming" ]; then
+# Loading the .env.bt
+export $(cat .env.bt | xargs)
```
mattThousand: in general looks great, but this will need to handle `.env.release` too, no?

aledustet: Good point, I think we should be good by default to the `release` one unless the test flight and beta releases need a specific name.

setup_env.sh
```diff
@@ -24,4 +31,35 @@ if [ ! -f pathcheck-mobile-resources/environment/$HA/.env.bt.release ]; then
...
\ No newline at end of file
+cp -r pathcheck-mobile-resources/environment/$HA/. ./
+
+# In order to read properly spaces on environment variables
+IFS='
+'
+
+if [ "$2" = "--renaming" ]; then
+# Loading the .env.bt
+export $(cat .env.release | xargs)
```
mattThousand: ah, should this be `.env.bt.release`?

mattThousand:
I've updated `pathcheck-mobile-resources` - should `yarn set-ha mn --renaming` set the value in `BT/info.plist`?

aledustet:
@mattThousand I updated the line to load the `.env` file it should be happy with spaces and lines without `""` sorry for the hiccup let me know when you try it on your end

johnschoeman:
Looks like this is no longer relevant, should this be closed @aledustet ?

aledustet:
I don't think this is relevant anymore, closing

PR #1274, opened 2020-07-16
en.json 1.2.0 update

JacobJaffe:
Merges 1.2.0 Locale changes into develop. @alpita-masurkar for visibility.

We're merging this in so that we can immediately get a localization update from these strings.

Quick hack to be non-breaking with the `isGps`, probably want this to use tracing strategy content long term?  @johnschoeman


yonatanmk (approved):

vlasac (approved):
LGTM!

johnschoeman (approved):

PR #1277, opened 2020-07-16
Add accessibility labels to images and svgs

jeffstolz:
#### Description:

This commit adds the accessibility tag and accessibility labels to images and icons throughout the BT and GPS app.

Note: These changes did not include self-assessment as that design may change and we will not be launching for MVP. We will also need to add translations for languages other than English.

#### Linked issues:

https://trello.com/c/kvxfTxo2/192-as-a-user-in-minnesota-using-a-mobile-screen-reader-i-need-to-ensure-that-all-non-text-content-has-available-and-accurate-text-a

#### Screenshots:

No changes to the UI

#### How to test:

- Run app locally on a screen reader
- Ensure all images and icons are provided with a clear, concise, and accurate description


app/bt/content.ts
```diff
@@ -30,11 +30,14 @@ export const useBTCopyContent: StrategyCopyContentHook = (t) => {
     moreInfoHowContent: t('exposure_history.bt.how_does_this_work_para'),
     moreInfoWhyContent: t('exposure_history.bt.why_did_i_get_an_en_para'),
     personalPrivacyHeader: t('label.launch_screen2_header_bluetooth'),
+    personalPrivacyIconLabel: t('label.bluetooth_icon'),
     personalPrivacySubheader: t('label.launch_screen2_subheader_bluetooth'),
     notificationDetailsHeader: t('label.launch_screen3_header_bluetooth'),
+    notificationDetailsIconLabel: t('label.heart_icon'),
```
tparesi, line 36: If notificationDetailsIconLabel and shareDiagnosisIconLabel are the same for BT and GPS, you don't need to add them to the tracing strategy specific files. These can go right into the en.json and use useTranslation hook.

tparesi (approved):
Looks good. Just had one small comment

app/bt/content.ts
jeffstolz: Ah good point, I'll make the update before merging

PR #1281, opened 2020-07-17
Added Greek translations

bspyros:
Added translations for the Greek language

#### Description:

Added translations for the Greek language


johnschoeman (commented):
Looks good!

After discussion IRL, it makes sense to add this to the lokalize git flow such that updates to the translations files are accurately accounted for in an automated way.  We are planning on getting this setup soon and we'll add this file as a starting point. So we likely wont be adding this pr directly.

PR #1282, opened 2020-07-17
Gps 1.2.0 updates

jvisenti:
* Disable insecure http request in iOS builds
* Import hashes from location files on both iOS and Android

These commits are cherrypicked from `develop`


vlasac (approved):

JacobJaffe (approved):
Great, including in b3 now that we're doing another build

PR #1283, opened 2020-07-17
Fetch ha specific config with curl

johnschoeman:
Why:
Currently we are fetching ha specific environments with git submodules,
this causes ci builds to break for contributions that made from outside
contributors via forks. We would like to be able to quickly update the
HA config and also allow contributions from individuals outside of the
github organization.

This commit:
Removes the gitsubmodules and replaces them with a `bin/fetch_ha_env`
script that uses curl over the github api. Developers that want to use
this script will need to set a GITHUB_PERSONAL_TOKEN in their local
.env, which was also added in this pr.

So we are introducing a distinction between `.env` and `.env.bt` where
`.env` is local development environment variables and `.env.bt` are
variables meant to be used in the built binaries for the app.


mattThousand:
thanks for doing this! Looks like the `setup_env` invocations just need to be removed from `build.yml`?

johnschoeman:
> thanks for doing this! Looks like the `setup_env` invocations just need to be removed from `build.yml`?

Does that mean we dont need the env for building on the action?

aledustet (approved):
It would be great to have this so we can extend it with the display name config from #1273 not sure why the ubuntu test is failing and considering that the path there actually doesn't exist I'm surprised it ever worked 🤔

aledustet:
Oh I see, the workflow only runs when the `dev_setup.sh` file was changed. We might need to add two branches to that workflow given that we now have `bt` and `gps`

johnschoeman:
> It would be great to have this so we can extend it with the display name config from #1273

👍 

> not sure why the ubuntu test is failing and considering that the path there actually doesn't exist I'm surprised it ever worked 🤔

same, im guessing its been broken ever since the introduction of gaen into the code base

mattThousand (approved):

johnschoeman:
Turns out the android build is dependent on having a populated .env.bt file.  so added that back in. and created a bespoke script to avoid needing to use dotenv as thats not present in our current image. A future commit might want to consolidate these.

PR #1285, opened 2020-07-17
Fix Allow Exposure Notifications Button

octohub:
#### Description:
Addresses an issue where Exposure Notifications could not be enabled on the home screen, but only through onboarding.

#### How to test:
1. Install BT app.
2. Go through onboarding and do not enable exposure notifications.
3. On the home screen enable exposure notifications.
4. Observe exposure notifications can be enabled from the home screen.


mattThousand (approved):
:shipit: 💯

PR #1286, opened 2020-07-17
Add badge indicator when user has new exposure

amandabeiner:
### Testing Notes ###
- Simulate an exposure
- Expect a yellow badge to appear above the calendar icon in the main
tab navigator
- Visit the calendar tab
- Expect the badge to disappear
Include instructions for reviewers to run tests related to this PR.

 ### Fixed Issues ###
Resolves [issue](https://trello.com/c/cfe33Gbb/194-as-a-user-i-want-to-see-a-badge-on-the-calendar-icon-when-i-have-an-en-so-that-i-notice-it-as-soon-as-possible) where user was not receiving a badge on the calendar icon when they had
a new exposure.

 ### Tech Notes ###
When the exposureInfoSubscription is called after the user has a new
exposure, we were not updating the ExposureHistory context with the
information that a new exposure was found. This commit:
- Updates the userHasNewExposure when the subscription is called
- Removes a resetExposures function that is unused

#### Screenshots:

dot

Co-Authored-By: Thomas G Paresi <tparesi@gmail.com>

### DIFF
--- a/app/ExposureHistoryContext.tsx
+++ b/app/ExposureHistoryContext.tsx
@@ -23,7 +23,6 @@ interface ExposureHistoryState {
   hasBeenExposed: boolean;
   userHasNewExposure: boolean;
   observeExposures: () => void;
-  resetExposures: () => void;
   getCurrentExposures: () => void;
   lastExposureDetectionDate: Dayjs | null;
 }
@@ -33,7 +32,6 @@ const initialState = {
   hasBeenExposed: false,
   userHasNewExposure: true,
   observeExposures: (): void => {},
-  resetExposures: (): void => {},
   getCurrentExposures: (): void => {},
   lastExposureDetectionDate: null,
 };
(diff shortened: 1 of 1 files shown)


aledustet (approved):
🚢  :shipit:

mattThousand (changes requested):
If exposures are considered "viewed" only if the user has visited the calendar screen, we should persist the indicator between app launches (if a user force quits the app before "viewing" their new exposures, and then relaunches, the indicator should still be present) 🤔

aledustet:
Shouldn't this behave like the notifications badge? Once a user opens an app the badge is gone, right?

mattThousand:
ah, this might be a platform-specific convention (I am not familiar with android). On iOS, badges (https://developer.apple.com/documentation/usernotifications/unnotificationcontent/1649864-badge) on the app icon and internal to the app usually require a specific action to be taken, such as in Messages you need to visit a specific unread message thread to hide the badge. In our case I believe visiting the calendar is the UX that constitutes updating "read/unread" state. If you have unread messages, open the Messages app, do not visit the unread thread, and force quit the app, the badge state remains unchanged.

johnschoeman:
this looks like its no longer necessary in the GPS repo, should the pr be closed @amandabeiner ?

PR #1287, opened 2020-07-17
Use MS for iOS hashes

JacobJaffe:
Previously we have been using seconds for calculating our hashes. This has caused us to be out of sync with the Places MS hashes

https://stackoverflow.com/questions/28872450/conversion-from-nstimeinterval-to-hour-minutes-seconds-milliseconds-in-swift


jvisenti (changes requested):
Several tests are now failing with this change:
```
testTimeWindow
testScryptHashSpec
testValidScrypt
```

I assume we'll need to update the expected windows and hashes to the new expected values.

There may also be a mismatch in units for the interval. The docs for `timeWindows` states
> interval: location storage interval in seconds

but the function is invoked with a very large number of seconds (looks like milliseconds instead):
```
let timeWindow = timeWindows(interval: Double(60 * 5 * 1000))
```

jvisenti:
Closing in favor of #1289

PR #1289, opened 2020-07-19
SAF-826 Compute hashes using milliseconds on iOS

jvisenti:
https://pathcheck.atlassian.net/browse/SAF-826

* Standardized parameter and time to milliseconds when computing time windows for hashes
* Converted time to seconds in tests (which is what iOS expects)


JacobJaffe (approved):
💯 So much better than my hacky attempt at Swift

PR #1291, opened 2020-07-20
[SAF-821] Fix feature flag language reset

tnekret:
#### Description:
This PR addresses an issue where entering feature flags screen would reset selected language to default(en).

#### How to test:

- enable feature flags

- select (if available) language different than English

- toggle 'All Language Options' flag on

- language should stay the same

### DIFF
--- a/app/locales/languages.ts
+++ b/app/locales/languages.ts
@@ -120,14 +120,25 @@ const config = {
 };
 
 export const initProdLanguages = (): void => {
+  const selected = i18next.language;
+
+  if(PROD_RESOURCES.hasOwnProperty(selected)) {
+    i18next.use(initReactI18next).init({
+      ...config,
+        lng: selected,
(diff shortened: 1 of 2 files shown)


app/locales/languages.ts
```diff
@@ -120,14 +120,29 @@ const config = {
 };
 
 export const initProdLanguages = (): void => {
+  const selected = i18next.language;
+  const available = Object.prototype.hasOwnProperty.call(
+    PROD_RESOURCES,
+    selected,
+  );
+
+  if (available) {
```
JacobJaffe, line 129: Readability suggestion:

```js
    i18next.use(initReactI18next).init({
      ...config,
      lng: available ? selected : config.lng,
    });
```

JacobJaffe (approved):
Great, nice catch on this

PR #1298, opened 2020-07-21
HA Selection Intersect Trigger Only On Change

JacobJaffe:
The middleware for HA Selection has been triggering excessively, due to
1. a shallow comparison check on array equality, vs. a check on the sameness of the array HAs by ids
2. the persist middlelware triggering an intersection on app start always, due to the persist rehydrate action instantiating the store

This improves app launch performance & the toggle performance around toggling HAs. May fix https://pathcheck.atlassian.net/browse/SAF-817, will check that seperately

### DIFF
--- a/app/store/middleware/onChangedSelectedHealthAuthorities.ts
+++ b/app/store/middleware/onChangedSelectedHealthAuthorities.ts
@@ -1,22 +1,37 @@
 import { Dispatch, Middleware, AnyAction, MiddlewareAPI } from 'redux';
 import IntersectService from '../../services/IntersectService';
+import toggleSelectedHealthcareAuthorityAction from '../actions/healthcareAuthorities/toggleSelectedHealthcareAuthorityAction';
+import { HealthcareAuthority } from '../../api/healthcareAuthorities/getHealthcareAuthoritiesApi';
+
+const areHaListsEqual = (
+  listA: HealthcareAuthority[],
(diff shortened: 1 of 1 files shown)


app/store/middleware/onChangedSelectedHealthAuthorities.ts
```diff
@@ -1,22 +1,37 @@
...
-    healthcareAuthorities: { selectedAuthorities: stateAfter },
+    healthcareAuthorities: { selectedAuthorities: haListAfter },
   } = store.getState();
 
   try {
-    if (stateBefore !== stateAfter) {
-      IntersectService.checkIntersect(stateAfter);
+    if (
+      action.type === toggleSelectedHealthcareAuthorityAction.type &&
+      !areHaListsEqual(haListBefore, haListAfter)
```
vlasac, line 32: Is the second condition redundant here? The `toggleSelectedHealthcareAuthorityAction` removes or adds a HA, so `before` and `after` lists should also be different whenever that action is dispatched.

I decided to check the state before and after because I expected that we may add more actions that could change the HA list (auto subscribe, ...) so I wanted to make this middleware more "future proof". Looking at it now, it may be more reasonable to explicitly add those intersection-triggering actions to this condition and remove the list comparison altogether, as that has proven to be prone to unexpected behaviour.

vlasac (approved):

PR #1299, opened 2020-07-21
Update Workflows: Remove BT, Make Conditional

JacobJaffe:
Changes our workflows to only run when affected code has changed.
Always Run (very fast):
- Validate
- i18n 

Run on JS changes: 
- test-js

Run on ios changes:
- test-ios

Run on Push to develop:
- build
- e2e test (takes a long time)

**Would appreciate feedback on this! these are up in the air, I think we're spending a lot of timing waiting on tests that don't need to run, so this should be more efficient**


vlasac (approved):
💯

.github/workflows/test-ios.yml
```diff
@@ -0,0 +1,60 @@
...
+  push:
+    branches:
+      - develop
+    paths:
+      - ios/**
+  pull_request:
+    branches:
+      - develop
+    paths:
+      - ios/**
```
johnschoeman, line 13: 💯

johnschoeman (approved):
This looks great.

It is worth noting that it is possible to break the ios build by only changing js code via react native things. For example if you add or remove a library that has a native dependency that doesnt work for whatever reason. Personally, I think its a risk worth taking in making ci more useful as it would be pretty obviously broken during deploy builds and running locally.

JacobJaffe:
> This looks great.
> 
> It is worth noting that it is possible to break the ios build by only changing js code via react native things. For example if you add or remove a library that has a native dependency that doesnt work for whatever reason. Personally, I think its a risk worth taking in making ci more useful as it would be pretty obviously broken during deploy builds and running locally.

Good point -- didn't mean to imply this would give us equal confidence. All the checks on Push are kept here, so I think we'll still be able to catch those regressions fairly quickly

yonatanmk (approved):

PR #1302, opened 2020-07-21
Joi verification for manual JSON mock data

JacobJaffe:
Adds verification on the endpoint for our mock data ingest. This allows a nicer error, as we've had issues with incorrectly formatted files.

Thanks @tnekret for the suggestion on going about this with Joi


tnekret (approved):
Awesome!

PR #1309, opened 2020-07-22
Workflow Fixes

JacobJaffe:
Fixes workflow paths, also lints them via `me-dutour-mathieu.vscode-github-actions`, catching some syntax issues we've had.

Also causes workflows to always trigger on develop push, instead of conditionally on develop push

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,13 +4,13 @@ on:
   push:
     branches:
       - gps-release
-     paths-ignore:
-      - 'package.json'
-      - 'CHANGELOG.md'
+    paths-ignore:
+      - "package.json"
+      - "CHANGELOG.md"
 
 jobs:
   version_bump:
-    name: 'Bump version'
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -118,21 +118,21 @@ jobs:
...
-          bundle-install-path: 'vendor/bundle'
+          lane: "play_store_gps"
+          subdirectory: "android"
+          bundle-install-path: "vendor/bundle"
 
   deploy_ios_gps:
     # gps-release branch only
     if: github.event_name == 'push' && github.ref == 'refs/heads/gps-release'
-    runs-on: macOS-latest
+    runs-on: macos-latest
```
JacobJaffe, line 135: https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources 
this has been wrong, but github has been interpreting it correctly 👀

jvisenti (approved):

PR #1311, opened 2020-07-22
Fix error when using joi dependency

juancruzgs:
We are getting this error when running the Android app
device-2020-07-22-165300

This change fixed the issue 
Reference: https://github.com/sideway/joi/issues/2141

### DIFF
--- a/app/api/import/importLocationsApi.ts
+++ b/app/api/import/importLocationsApi.ts
@@ -1,3 +1,4 @@
+import 'text-encoding-polyfill';
 import Joi from '@hapi/joi';
 const MockData = Joi.array()
   .items(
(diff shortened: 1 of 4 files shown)


app/api/import/importLocationsApi.ts
```diff
@@ -1,3 +1,4 @@
+import 'text-encoding-polyfill';
```
JacobJaffe: Can you import this in `app.js` instead? We have a similar pollyfill there: 
```js
import 'array-flat-polyfill';
```

JacobJaffe:
@juancruzgs looks like in the tests the yarn lock thinks this was added via npm?

app/api/import/importLocationsApi.ts
juancruzgs: Done

juancruzgs:
@JacobJaffe I just run `yarn add text-encoding-polyfill`
It might be something wrong with my configuration, I'll check it

juancruzgs:
@JacobJaffe It's fixed now

JacobJaffe (approved):
Great, thanks!

PR #1313, opened 2020-07-23
[SAF-822] Fix HA code input

tnekret:
#### Description:
This PR fixes following things: 

-  input field box flicker when entering numbers in a rapid succession
-  keyboard not hiding when pressing on 'done' button


JacobJaffe (approved):
Great, that'll make this look a lot cleaner. Tagging @johnschoeman for visibility into this fix for BT usage

PR #1314, opened 2020-07-23
Cause workflows to alway run, but pass when not needed

JacobJaffe:
Improves our workflows so we can require all tests on PRs, but they will auto succeed if no changes detected. We can adjust what "changes detected" means as we see fit.

### How this works:
```sh
URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
FILES=$(curl -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -s -X GET -G $URL |  jq -r '.[] | .filename')
echo "Checking Files: $FILES"
REGEX="\.js\|\.jsx\|\.ts\|\.tsx"
if echo $FILES | grep -q $REGEX; then
  echo "Detected iOS changes. Running tests!"
  echo "Changed Files:"
  echo $FILES | grep $REGEX
  echo "::set-env name=RUN_TESTS::1"
else
  echo "No iOS changed detected. Skipping tests."
  echo "::set-env name=RUN_TESTS::0"
fi
```

1. Fetch JSON of files from current PR. Do this in an authorized request, otherwise we face rate limiting by GitHub.
2. Define a regex for our files we watch for. This is like what we previously did with `--paths`, except this happens _within_ the job.
3. if those files match the changed files, set the `RUN_TESTS` command
4. make each step conditional `RUN_TESTS`

step 4 is a little tedious, but it seems that there is no way to _succeed_ a workflow early, only fail it.

### How to test this:
change files matching the regex, make a PR with this code

### DIFF
--- a/.github/workflows/test-ios.yml
+++ b/.github/workflows/test-ios.yml
@@ -7,8 +7,6 @@ on:
   pull_request:
     branches:
       - develop
-    paths:
-      - "ios/**"
 
 jobs:
   unit-test-ios:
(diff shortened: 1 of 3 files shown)


.github/workflows/test-ios.yml
```diff
@@ -17,10 +15,28 @@ jobs:
...
       - uses: actions/checkout@master
 
+      - name: Check if run is needed
+        run: |
+          URL="https://api.github.com/repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files"
+          FILES=$(curl -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -s -X GET -G $URL |  jq -r '.[] | .filename')
+          echo "Checking Files: $FILES"
+          REGEX="\.js\|\.jsx\|\.ts\|\.tsx"
+          if echo $FILES | grep -q $REGEX; then
+            echo "Detected JavaScript changes. Running tests!"
```
johnschoeman: wouldnt we want these to run only if there are changes in the ios/ folder?

johnschoeman (approved):
Code looks good! Though the idea strikes me as adding a fair bit of complexity and prone to breakage over time. I wonder if it would be better to spend some effort to improve the speed of the test suite itself before going down the path of running tests conditionally based off file changes?

Also something about pinging the github api from a github action seems off.  Im not too familiar with the box setup here, but is the git history in the file system? could we use git to check the diff directly?

PR #1321, opened 2020-07-27
[SAF-69] - Autosubscribe to HA on app start

vlasac:
#### Description:
This PR adds a new feature for auto-subscribing users to their local health department after the location permissions have been granted and they have no HAs selected. 

#### How to test:
(Simulator)
- Mock simulator location so that it is within the bounds of at least one known HA from the YAML
- Clean install the app and perform onboarding, granting location permissions
- Upon entering the home screen, the banner should appear at the bottom, saying that you've been subscribed to a HA
- When navigating to `View your Health Departments` screen, the auto-subscribed HA should appear as selected

(Real device)
- Hope that we have a HA in the yaml that has your location within it's bounds -> you should get auto-subscribed after onboarding and get a banner on the bottom of the screen
- Rest is the same as for the simulator


app/common/types.ts
```diff
@@ -0,0 +1,28 @@
+import { Posix } from '../helpers/dateTimeUtils';
```
JacobJaffe, line 1: thanks for adding this file! nice refactor

app/gps/PermissionsContext.tsx
```diff
@@ -38,11 +38,6 @@ const initialState = {
     check: () => {},
     request: () => {},
   },
-  authSubscription: {
```
JacobJaffe, line 41: BT specific?

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -33,6 +34,7 @@ const getHealthcareAuthorities_failure = createAction(
 
 const getHealthcareAuthoritiesAction = (
   customYamlUrl?: string,
+  autoSubscriptionLocation?: Coordinates,
```
JacobJaffe: Since these are both optional, can this be

```js
= ({ 
  customYamlUrl?: string,
  autoSubscriptionLocation?: Coordinates,
})
```
(With `{ }`)

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
```diff
@@ -45,6 +47,20 @@ const getHealthcareAuthoritiesAction = (
...
     );
+
+    if (autoSubscriptionLocation) {
+      const localHealthAuthority = healthcareAuthorities.find((ha) =>
+        isLocationWithinBounds(ha, autoSubscriptionLocation),
+      );
+
+      if (localHealthAuthority) {
+        dispatch(
+          healthcareAuthorityAutoSubscribeAction({
```
JacobJaffe: Instead of using a new action, can you fire the existing toggle action? These actions duplicate in intent. 
```js
toggleSelectedHealthcareAuthorityAction({ authority: localHealthAuthority, override: true})
```

That way our actions are are named by intent, rather than context, so they are composable

app/views/Partners/PartnersOverview.tsx
```diff
@@ -113,6 +130,24 @@ const PartnersScreen = ({ navigation }: PartnersScreenProps): JSX.Element => {
         <View style={styles.divider} />
         <View style={{ height: 24 }} />
       </ScrollView>
+      <View style={styles.bottomSheet}>
```
JacobJaffe: Can we keep this hidden or removed? We still don't want this visible to the end user.

app/views/main/AllServicesOn.tsx
```diff
@@ -1,25 +1,81 @@
...
 import { Colors } from '../../styles';
+import { Stacks, NavigationProp } from '../../navigation';
+import { useSelector, useDispatch } from 'react-redux';
+import isAutoSubscriptionEnabledSelector from '../../store/selectors/isAutoSubscriptionEnabledSelector';
+import selectedHealthcareAuthoritiesSelector from '../../store/selectors/selectedHealthcareAuthoritiesSelector';
+import getHealthcareAuthorities from '../../store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction';
 
 type AllServicesOnProps = {
   noHaAvailable: boolean;
+  navigation: NavigationProp;
```
JacobJaffe, line 27: Instead of  passing navigation, can this use the `useNavigation` hook?

app/views/main/AllServicesOn.tsx
```diff
@@ -1,25 +1,81 @@
...
+      Geolocation.getCurrentPosition(({ coords }) => {
+        dispatch(getHealthcareAuthorities(undefined, coords));
+      });
+    }
+  }, [autoSubscriptionEnabled, noHaAvailable, dispatch]);
+
+  useEffect(() => {
+    autoSubscribe().then(() => {
+      if (selectedAuthorities[0]) {
+        setHAName(selectedAuthorities[0].name);
```
JacobJaffe: I don't think this callback will act as expected because this is dependent on the callback chain of thunks resolving. Also, I don't think the 
```js
   const [haName, setHAName] = useState(''); 
``` 
is needed. 
Instead, we can directly use the selector value. 
```js
const haName = useSelector(state => state.selectedAuthorities[0]?.name)
```

app/views/main/AllServicesOn.tsx
```diff
@@ -64,6 +120,33 @@ export const AllServicesOnScreen = ({
...
+              <Typography
+                style={styles.hyperlink}
+                // TODO: change this
+                onPress={() => navigation.navigate(Stacks.Partners)}>
+                {t('home.gps.auto_subscribe_link_text')}
+              </Typography>
+            </>
+          </Typography>
+          <View style={{ width: 24 }} />
+          {/* TODO: change this */}
```
JacobJaffe: Is this comment still valid?

JacobJaffe (changes requested):

app/gps/PermissionsContext.tsx
vlasac: Judging by the `auth` typo, I would say that this is some dangling dead code, not mentioned anywhere else in the repo, and it doesn't seem right to put the auto subscribe feature code alongside actual permissions.

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
vlasac: Couldn't agree more, I had the same idea at some point.

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
vlasac: The motivation for adding this specific action was to bypass the middleware that fires off the intersection algo.

I agree that this is an unnecessary bloating of redux interface and would like to change this, although it would be good to find a different way for firing off the intersection algo since it would attempt to download all of that HA data right on the app start.

app/views/Partners/PartnersOverview.tsx
vlasac: sure thing!

app/store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction.ts
vlasac: Per your advice, this was improved by adding a `meta` field to the action creator. Now we can have more control over firing off the intersection calculation

app/views/main/AllServicesOn.tsx
vlasac: fixed!

app/views/main/AllServicesOn.tsx
vlasac: nope, removed!

JacobJaffe (approved):
Looks good! Excited to get this in for 1.4.0

PR #1328, opened 2020-07-28
Calendar design GPS parity

JacobJaffe:
We have been missing some parity on the calendar to the Figma designs. 
Note that this _intentionally_ changes the locale strings besides English.

- disables buttons in future, and outside of 14 days
- fixes months to be relative to the dates shown, rather than just previous and current month. (Currently incorrect, as we're in the end of the month).

Before | After
:-------------------------:|:-------------------------:
image |image


vlasac (approved):
All lookin' good, LGTM!

PR #1333, opened 2020-07-30
Fixes/saf 791

Sameerlaul:
#### Description:

This PR is regarding SAF-791

made changes in styles to align green tick in center in "Health Departments" tab.
Also this problem was seen in while selecting HA before "Enter verification code" screen

#### Linked issues:

SAF-791

#### Screenshots:
ss


JacobJaffe:
Hi @Sameerlaul , thanks for the PR. Looks like you need to run prettier over these changes.

I'd recommend avoiding using percentage based approaches for fixing these flex issues, as that won't be very composable within other container contexts. I've included a fix for this here: https://github.com/Path-Check/covid-safe-paths/pull/1335/files#diff-9d34aa79061c949dd6d579fc47055cd4R97

Which solves this by adding flex grow (through flex) on the container which overflowed the parent on android.

Sameerlaul:
since this issue is already resolved in PR #1335, I am closing this PR

PR #1347, opened 2020-08-05
fix autosubscribing across sessions

JacobJaffe:
WIP: Fixes autosubscribing between app sessions


app/views/main/AllServicesOn.tsx
```diff
@@ -9,75 +9,47 @@ import {
...
+import { Colors, Spacing } from '../../styles';
 import { Stacks, NavigationProp } from '../../navigation';
 import { useSelector, useDispatch } from 'react-redux';
-import isAutoSubscriptionEnabledSelector from '../../store/selectors/isAutoSubscriptionEnabledSelector';
-import getHealthcareAuthorities from '../../store/actions/healthcareAuthorities/getHealthcareAuthoritiesAction';
 import { RootState } from '../../store/types';
+import toggleAutoSubscriptionBanner from '../../store/actions/healthcareAuthorities/toggleAutoSubscriptionBannerAction';
 
 type AllServicesOnProps = {
   noHaAvailable: boolean;
```
vlasac: Can we remove this prop? I think it's unused and its value can be easily pulled from redux if needed. I would also ease the return logic on the `Main` screen.

vlasac (approved):
Nice refactor of the auto-subscription logic, thanks for addressing the cross-session state keeping. Also, 💯 for overall readability and stability!

I added a potential small improvement of removing the unused prop on one of the screens, but all in all looks really nice!

PR #1349, opened 2020-08-06
[SAF-848] update cursor caching

tnekret:
#### Description:

Added the ability to store checksum as cacheKey in async storage when using new cursor file version.

#### How to test:

Since this is a new feature on safe places backend, there is currently a fallback in place for old version of HA cursor files. 

In version 1.1 we should see a checksum field when fetching cursor file. Checksum helps us determine whether the contents of fetched page has changed which in turn reduces unnecessary fetching of data and helps us keep locally stored data updated when HA publishes updates. 

This feature currently lives in test(staging) HA so to test this we need to subscribe to staging HA, wait for our app to fetch data and then update one or more pages(change hashes) and then wait for automatic update or force Cursor update by resubscribing


app/api/healthcareAuthorities/getCursorApi.ts
```diff
@@ -5,6 +5,7 @@ type Page = {
   startTimestamp: number;
   endTimestamp: number;
   filename: string; // to be appended to 'api_endpoint_url'
+  checksum?: string;
```
JacobJaffe: Maybe we can make these types based on the version?  

psudo:

```
type Page1.0 = {
  ...
}

type Page1.1 = Page1.0 & { checksum: string };

type Cursor1.0 = { ... };

type Cursor1.1 = Cursor1.0 & {version: 1.1, pages: Page1.1};
```

app/helpers/Intersect.js
```diff
@@ -528,6 +532,13 @@ async function getPageData(authority, page) {
   return pageData || {};
 }
 
+function shouldUpdatePage(page, pageData) {
+  if (pageData.version !== '1.0') {
```
JacobJaffe:
```suggestion
  if (pageData.version >= '1.1') {
```

Clarifies what changes occurred in the versioning imo

tnekret: Fair point

JacobJaffe (approved):
LGTM! thanks for updating the types

PR #1352, opened 2020-08-07
Update lokalise pull script

aledustet:
Why:
----
The pull script from lokalise is creating the translation files for ios and android in the wrong paths

This Commit:
----
- Update the bundle structure option for the ios pull script from lokalise
- Update the bundle structure option for the android pull script from lokalise


JacobJaffe (approved):
💯 thank you for updating this!

PR #1354, opened 2020-08-13
Lokalise latest august 13

alpita-masurkar:
#### Description:
Lokalise updates from aug 13


android/app/src/main/res/values-kk/strings.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<resources>
+  <!-- TODO remove and put into build.gradle -->
+  <string name="key_server_upload_uri">https://exposure-2kabcv6c4a-uc.a.run.app/</string>
```
aledustet: This looks like a reminiscence from the repo split, might be best to delete the key from lokalise

aledustet (approved):
Other than that key `key_server_upload_uri` this looks ready to 🚢

android/app/src/main/res/values-kk/strings.xml
alpita-masurkar: ok, done! deleted from lokalise

android/app/src/main/res/values-es-rPR/strings.xml
```diff
@@ -2,9 +2,6 @@
 <resources>
   <!-- Channel label for notifications shown in App > Notifications > Path Check BT -->
   <string name="notification_channel_description">Notificaciones de PathCheck BT</string>
-  <!-- Server URI for diagnosis key file download -->
-  <string name="key_server_download_base_uri">https://federationout-2kabcv6c4a-uc.a.run.app/</string>
```
alpita-masurkar, line 6: this is flagged as a non-translatable key so no other file has this "translated". but this is the only file in which there was a value in it, which is different from our standard value for this key "https://storage.googleapis.com/exposure-notification-export-svhfv/". do we want to add this key/ value back into the rPR file @aledustet ?

